PR #18, opened 2020-02-18
fix: remove erroneous timeouts for batch_create_session calls

larkee:
Currently the timeout that is used for getting a session from the pool is mistakenly being used for the batch_create_sessions calls that are used to initially fill pools. If the timeout is short, the call does not have time to complete and the pools fails to initialize. Removing this timeout means [the default timeout for BatchCreateSessions](https://github.com/googleapis/python-spanner/blob/master/google/cloud/spanner_v1/gapic/spanner_client_config.py#L44) will be used.

### DIFF
--- a/google/cloud/spanner_v1/pool.py
+++ b/google/cloud/spanner_v1/pool.py
@@ -173,7 +173,6 @@ class FixedSizePool(AbstractSessionPool):
             resp = api.batch_create_sessions(
                 database.name,
                 self.size - self._sessions.qsize(),
-                timeout=self.default_timeout,
                 metadata=metadata,
             )
             for session_pb in resp.session:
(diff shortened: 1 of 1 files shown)


google/cloud/spanner_v1/pool.py
```diff
@@ -171,10 +171,7 @@ def bind(self, database):
 
         while not self._sessions.full():
             resp = api.batch_create_sessions(
-                database.name,
-                self.size - self._sessions.qsize(),
-                timeout=self.default_timeout,
```
skuruppu: Out of curiosity, what timeout will be used now that you've removed this parameter?

larkee: It will use the default timeout that is specified by the spanner_client_config which is one minute:
https://github.com/googleapis/python-spanner/blob/master/google/cloud/spanner_v1/gapic/spanner_client_config.py#L44

skuruppu: Ah ok thanks. If you can update the PR description, that would be great.

PR #22, opened 2020-02-20
chore!: remove Python 2.7 from testing

larkee:
Python 2.7 has been deprecated which means we no longer need to run these tests to ensure compatibility. Removing these will significantly speed up the presubmit checks.


crwilcox:
Thanks @larkee! Can you also remove the classifiers from setup.py? https://github.com/googleapis/python-spanner/blob/master/setup.py#L73

Also, if we make this change, it would be best to consider it a breaking change. Any user who is pinning to the current major could be broken by us accidentally later.

PR #24, opened 2020-02-20
Revert "chore!: remove Python 2.7 from testing (#22)"

larkee:


crwilcox:
Just leaving a note. The plan is to do this for the next major release, but hold for another breaking change to not force a 2.0 for this alone.

PR #30, opened 2020-03-09
feat: implement query options versioning support

larkee:
Adds support for QueryOptions.

The QueryOptions are set in one of three ways (in ascending order of precedence):

* As an argument to the Client constructor.
* Setting the SPANNER_OPTIMIZER_VERSION env variable.
* On a per-query basis.

If the backend implements query hints, for example, to instruct the backend which query optimizer version to use, these query hints will also automatically be supported.

### DIFF
--- a/google/cloud/spanner_v1/_helpers.py
+++ b/google/cloud/spanner_v1/_helpers.py
@@ -75,13 +75,12 @@ def _merge_query_options(base, merge):
         combined = ExecuteSqlRequest.QueryOptions(
             optimizer_version=combined.get("optimizer_version", "")
         )
-    if merge:
-        if type(merge) == dict:
-            merge_ver = merge.get("optimizer_version", "")
-        else:
-            merge_ver = merge.optimizer_version
-        if merge_ver:
-            combined.optimizer_version = merge_ver
+    merge = merge or ExecuteSqlRequest.QueryOptions()
+    if type(merge) == dict:
+        merge = ExecuteSqlRequest.QueryOptions(
+            optimizer_version=merge.get("optimizer_version", "")
+        )
+    combined.MergeFrom(merge)
     if not combined.optimizer_version:
         return None
     return combined
--- a/google/cloud/spanner_v1/_helpers.py
+++ b/google/cloud/spanner_v1/_helpers.py
@@ -26,6 +26,7 @@ from google.api_core import datetime_helpers
 from google.cloud._helpers import _date_from_iso8601_date
 from google.cloud._helpers import _datetime_to_rfc3339
 from google.cloud.spanner_v1.proto import type_pb2
+from google.cloud.spanner_v1.proto.spanner_pb2 import ExecuteSqlRequest
 
 
 def _try_to_coerce_bytes(bytestring):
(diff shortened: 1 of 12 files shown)


tests/unit/test_client.py
```diff
@@ -97,15 +103,17 @@ def _constructor_test_helper(
                 client._client_options.api_endpoint,
                 expected_client_options.api_endpoint,
             )
+        if expected_query_options is not None:
+            self.assertEqual(client._query_options, expected_query_options)
 
-    @mock.patch("google.cloud.spanner_v1.client.os.getenv")
+    @mock.patch("google.cloud.spanner_v1.client._get_spanner_emulator_host")
```
skuruppu: Not sure why there's a diff for `_get_spanner_emulator_host` in this PR.

larkee: Both emulator support and query options support rely on calling os.getenv which means it's more difficult to set up the mock. The `os.getenv` call for emulator is made in `_get_spanner_emulator_host` which means mocking that function instead is much more clear.

google/cloud/spanner_v1/client.py
```diff
@@ -62,6 +63,7 @@
     "without a scheme: ex %s=localhost:8080."
 ) % ((EMULATOR_ENV_VAR,) * 3)
 SPANNER_ADMIN_SCOPE = "https://www.googleapis.com/auth/spanner.admin"
+OPTIMIZER_VERSION_VAR = "SPANNER_OPTIMIZER_VERSION"
```
hengfengli: probably use `OPTIMIZER_VERSION_ENV_VAR`

google/cloud/spanner_v1/_helpers.py
```diff
@@ -47,6 +48,45 @@ def _try_to_coerce_bytes(bytestring):
...
+        should overwrite the fields in base.
+
+    :rtype:
+        :class:`google.cloud.spanner_v1.proto.ExecuteSqlRequest.QueryOptions`
+        or None
+    :returns:
+        QueryOptions object formed by merging the two given QueryOptions.
+        If the resultant object only has empty fields, returns None.
+    """
+    combined = base or ExecuteSqlRequest.QueryOptions()
```
hengfengli: nits: probably think a better way to merge in a more generic way (less code if we add new fields). Currently, we need to keep add new fields. 

E.g., convert to `dict` first, merge, and then create `ExecuteSqlRequest.QueryOptions`.

google/cloud/spanner_v1/session.py
```diff
@@ -225,11 +226,21 @@ def execute_sql(
         :param query_mode: Mode governing return of results / query plan. See
             https://cloud.google.com/spanner/reference/rpc/google.spanner.v1#google.spanner.v1.ExecuteSqlRequest.QueryMode1
 
+        :type query_options:
+            :class:`google.cloud.spanner_v1.proto.ExecuteSqlRequest.QueryOptions`
```
hengfengli: Can it be a `dictionary`?

larkee: Done.

google/cloud/spanner_v1/_helpers.py
larkee: I discovered the MergeFrom protobuf function so I refactored to use that.

larkee:
@hengfengli `partition_query` does not support `query_options` as it does not call `executeSql` or `executeStreamingSql`

hengfengli:
> @hengfengli `partition_query` does not support `query_options` as it does not call `executeSql` or `executeStreamingSql`

Actually, I meant this method `generate_query_batches`. You have done it already.

PR #35, opened 2020-03-13
feat: add support for backups

larkee:
Adds support for the new Cloud Spanner backups feature.

See backup samples for examples of common use cases.

### DIFF
--- a/google/cloud/spanner_v1/backup.py
+++ b/google/cloud/spanner_v1/backup.py
@@ -1,4 +1,4 @@
-# Copyright 2016 Google LLC All rights reserved.
+# Copyright 2020 Google LLC All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/google/cloud/spanner_v1/backup.py
+++ b/google/cloud/spanner_v1/backup.py
@@ -1,0 +1,275 @@
+# Copyright 2016 Google LLC All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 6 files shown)


google/cloud/spanner_v1/backup.py
```diff
@@ -0,0 +1,275 @@
+# Copyright 2016 Google LLC All rights reserved.
```
skuruppu:
```suggestion
# Copyright 2020 Google LLC All rights reserved.
```

google/cloud/spanner_v1/database.py
```diff
@@ -521,6 +560,73 @@ def run_in_transaction(self, func, *args, **kw):
...
+        :rtype: :class:'~google.api_core.operation.Operation`
+        :returns: a future used to poll the status of the create request
+        :raises Conflict: if the database already exists
+        :raises NotFound:
+            if the instance owning the database does not exist, or
+            if the backup being restored from does not exist
+        :raises ValueError: if backup is not set
+        """
+        if backup is None:
+            raise ValueError("Restore source not specified")
```
skuruppu: Should the backup arg not have a default value if None throws a ValueError?

tests/unit/test_backup.py
```diff
@@ -0,0 +1,590 @@
+# Copyright 2016 Google LLC All rights reserved.
```
skuruppu:
```suggestion
# Copyright 2020 Google LLC All rights reserved.
```

google/cloud/spanner_v1/database.py
larkee: The backups argument is optional because it is optional in the gapic method. The reason it is optional is because it is a oneof. If we introduce different restore sources in the future, making this optional would be a breaking change. However, I could change the argument to source instead.

skuruppu: Ah ok, I missed that. I think it might be better to change it to source.

larkee: Done.

PR #42, opened 2020-03-18
docs: fix incorrect return types

### LINKED ISSUE #41: Incorrect return type in update_database_ddl() docs
In `update_database_ddl()` method docs it's said:
https://github.com/googleapis/python-spanner/blob/e378ba039fc5714c715386b3a1d9ba5b2c7ce306/google/cloud/spanner_admin_database_v1/gapic/database_admin_client.py#L478-L479

But looks like in fact in returns `google.api_core.operation.Operation`:

Безымянный

It's documented correctly in method `Database.update_ddl()`, which calls `update_database_ddl()` and returns it's return value:
https://github.com/googleapis/python-spanner/blob/e378ba039fc5714c715386b3a1d9ba5b2c7ce306/google/cloud/spanner_v1/database.py#L372-L373

It's the only place where `update_database_ddl()` is called, so I assume we can safely change return type in docs.

IlyaFaer:
Fixing the method doclines from the original issue, and three similar places as well
Fixes #41

### DIFF
--- a/google/cloud/spanner_admin_database_v1/gapic/database_admin_client.py
+++ b/google/cloud/spanner_admin_database_v1/gapic/database_admin_client.py
@@ -292,7 +292,7 @@ class DatabaseAdminClient(object):
                 that is provided to the method.
 
         Returns:
-            A :class:`~google.cloud.spanner_admin_database_v1.types._OperationFuture` instance.
+            A :class:`~google.api_core.operation.Operation` instance.
 
         Raises:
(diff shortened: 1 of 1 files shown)


larkee:
Hey, thanks for making this PR!

Unfortunately it makes changes to a generated file which means the changes will be overwritten next time the client is generated.

I've written a separate PR that will replace the strings during generation until the generator is fixed so I'm going to close this one.

PR #43, opened 2020-03-23
docs: Fix incorrect generated return types

### LINKED ISSUE #41: Incorrect return type in update_database_ddl() docs

larkee:
Closes #41

### DIFF
--- a/synth.py
+++ b/synth.py
@@ -133,6 +133,14 @@ s.replace(
     '"""Protocol buffer.\n\n  Attributes:',
 )
 
+# Fix LRO return types
+s.replace("google/cloud/spanner_admin_instance_v1/gapic/instance_admin_client.py",
+          "cloud.spanner_admin_instance_v1.types._OperationFuture",
+          "api_core.operation.Operation")
+s.replace("google/cloud/spanner_admin_database_v1/gapic/database_admin_client.py",
+          "cloud.spanner_admin_database_v1.types._OperationFuture",
+          "api_core.operation.Operation")
+
 # ----------------------------------------------------------------------------
 # Add templated files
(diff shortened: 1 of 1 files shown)


IlyaFaer:
> Unfortunately it makes changes to a generated file

I suppose, it would be good to add a disclaimer into the generated files. Something like `Generated file. Do not edit!`, otherwise it's not very obvious, unless you know what is gapic and expect to meet generated files.

As they are doing this in other clients, I (for example) have a habit to check if this line is in the file: if not, consider file isn't generated.

larkee:
> I suppose, it would be good to add a disclaimer into the generated files. Something like `Generated file. Do not edit!`, otherwise it's not very obvious, unless you know what is gapic and expect to meet generated files.
> 
> As they are doing this in other clients, I (for example) have a habit to check if this line is in the file: if not, consider file isn't generated.

That makes sense. I think I have seen that in some of the other generated files. I'm not sure why it has been removed from these ones. I'll mention it to the generator team.

PR #44, opened 2020-03-23
docs: Add documentation for using DML statements in a transaction

### LINKED ISSUE #17: spanner: unconsumed/uniterated StreamedResultSet doesn't send request to Cloud Spanner server
This is an issue that has plagued me for a while but I just got the time to make a repro.

Basically, if I try for example to invoke Transaction.execute_sql and do NOT consume the result e.g.
```python
txn.execute_sql('DELETE from T1 WHERE 1=1')
```
instead of
```python
res = txn.execute_sql('DELETE from T1 WHERE 1=1')
_ = list(res)
```
then the table will NOT be purged.

Seems like a bug to me with the underlying gRPC library, but it would be useful to explicitly document/call-out this bug if we don't have the bandwidth to fix it, to avoid unexpected problems for customers. It definitely sunk some hours for me in the past and also just right now.

larkee:
Closes #17

### DIFF
--- a/docs/transaction-usage.rst
+++ b/docs/transaction-usage.rst
@@ -51,6 +51,18 @@ fails if the result set is too large,
         print(row)
 
 
+Execute a SQL DML Statement
+------------------------------
+
+Modify data from a query against tables in the database.  Calls
+the ``ExecuteSql`` API, and returns the number of rows affected,
+
+.. code:: python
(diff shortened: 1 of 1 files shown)


hengfengli:
@larkee I don't have write access to this repo, :-(

PR #59, opened 2020-04-17
docs: make it clear ddl_statements is a list in docs

bpg130:
The `ddl_statements` fields expects a `list` of `strings`, not a `string`. This aims to make that more clear in the documentation.

If it is required that I open a bug for doc changes let me know and I will be happy to do so!


bpg130:
Thanks larkee, I have updated the commit message.

Issue #60, opened 2020-04-21
PingingPool and TransactionPingingPool implementation does not match documentation


larkee:
There is a major bug in the [ping()](https://github.com/googleapis/python-spanner/blob/master/google/cloud/spanner_v1/pool.py#L418) function used by both of these pools. The function breaks out of the `while True:` loop when the pool is empty or a session does not need to be pinged yet. This means it is unsuitable for use as a background thread as we suggest because the loop is likely to end the first time it is run.

Additionally, TransactionPingingPool puts used sessions into a pending sessions queue so they can have transactions started on them. However the [begin_pending_transactions()](https://github.com/googleapis/python-spanner/blob/master/google/cloud/spanner_v1/pool.py#L506) function that removes them only runs once when the pool is created and once when the pool is bound to a database. The condition for the function loops is: 
`while not self._pending_session.empty():`
which means if at any point there are no pending sessions, then any future pending sessions will not be refreshed. There is no documentation suggesting the a user needs to run this themselves which means this is another major bug.

larkee:
While creating the above PR I read the docstring for `ping()` more closely:
`This method is designed to be called from a background thread, or during the "idle" phase of an event loop.`

This means the implementation does actually make sense as it is intended to be included in a loop. However this is not how our documentation uses it so the documentation will need to be updated.

larkee:
Given the above I checked the docstring for TransactionPingingPool  which says:
> When a session is returned to the pool, if its transaction has been committed or rolled back, the pool creates a new transaction for the session and pushes the transaction onto a separate queue of "transactions to begin."  **The application is responsible for flushing this queue as appropriate via the pool's `begin_pending_transactions` method.**

The advanced-session-pool-topics documentation (which is the only reference users how on how to use this pool) does not mention this at all and implies that this will be done for the user by the background thread.

PR #61, opened 2020-04-21
fix: make ping() suitable for running as a background thread

larkee:
Background threads need to be a loop. There was a bug such that the loop was likely to break on the first iteration of being run. This refactor fixes that bug.

Towards #60


larkee:
Just realized the intended usage outlined in the comment does make sense. It's that the documentation does not match.

PR #64, opened 2020-04-21
chore: Migrate python-spanner synth.py from artman to bazel

vam-google:


vam-google:
@busunkim96 PTAL

busunkim96:
@larkee This changes how the generator is called, but it should not change the result of the generation. Let me know if something is off in the next autosynth PR.

larkee:
Running the synthtool via `python -m synthtool` without any changes gives the following error:
> TypeError: _generate_code() got an unexpected keyword argument 'include_protos'

Removing the `include_protos` kwarg allows the synthtool to complete successfully. This fix is in #72

PR #66, opened 2020-04-24
feat: add support for retrying aborted partitioned DML statements

larkee:
Partitioned DML transactions can be aborted by Cloud Spanner. These transactions were not automatically retried by the client library. This PR adds support for automatically retrying aborted partitioned DML statements.

### DIFF
--- a/google/cloud/spanner_v1/database.py
+++ b/google/cloud/spanner_v1/database.py
@@ -21,8 +21,11 @@ import re
 import threading
 
 import google.auth.credentials
+from google.api_core.retry import Retry, if_exception_type
+import google.auth.credentials
 from google.protobuf.struct_pb2 import Struct
 from google.cloud.exceptions import NotFound
+from google.api_core.exceptions import Aborted
 import six
 
 # pylint: disable=ungrouped-imports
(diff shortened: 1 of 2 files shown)


larkee:
> Usage of `api_core`'s retry looks good to me.
> 
> Just to check (I'm not at familiar with the Spanner library), this retries the whole transaction?
> 
> https://aip.dev/194#generally-non-retryable-codes
> 
> > ABORTED: This code typically means that the request failed due to a sequencer check failure or transaction abort. These should not be retried for an individual request; they should be retried at a level higher (the entire transaction, for example).

Correct. It creates a new transaction on each retry attempt.

larkee:
> This looks reasonable but it would be good to add a description to this PR to explain why this came about.

Done.

> Also, the retry logic in [`run_in_transaction`](https://github.com/googleapis/python-spanner/blob/d63dcb68409a1de7cc17bba4afdab54925d427b1/google/cloud/spanner_v1/session.py#L276) looks quite different. It takes the retry config into account and does exponential backoff. Will this method of retrying take this config into account?

The difference in logic is because PDML transactions cannot be committed or rolled back. This means much of the handling in [`run_in_transaction`](https://github.com/googleapis/python-spanner/blob/d63dcb68409a1de7cc17bba4afdab54925d427b1/google/cloud/spanner_v1/session.py#L276) does not apply here.

This method currently does not take the config into account. I'll add that in shortly.

PR #69, opened 2020-04-27
tests: add backup integration tests

larkee:


larkee:
> Looks good. But as we've been finding for other languages, you'll find that these tests will often timeout during the backup creation or restore part. So it would be good to reduce the number of create and restore calls. E.g. I think I never added a restore to different instance test in Node.js.

I haven't run into any timeouts recently so I'm okay with keeping the different instance restore.

> You can also save time by not waiting for the optimize. You can do:
> 
> 1. Wait for restore to complete.
> 2. Delete restored database.
> 3. Delete backup.
>    It's faster this way and you don't get a FAILED_PRECONDITION error when you're trying to delete the backup.

Done :+1: This also fixed a flaky test!

PR #72, opened 2020-04-29
fix: remove 'include_protos' kwarg that causes synth to fail

larkee:
#64 caused a synth failure due to the bazel generator not supporting the `include_protos` kwargs.

Removing this kwarg allows the synth to complete successfully.

### DIFF
--- a/synth.py
+++ b/synth.py
@@ -26,7 +26,6 @@ library = gapic.py_library(
     service="spanner",
     version="v1",
     bazel_target="//google/spanner/v1:spanner-v1-py",
-    include_protos=True,
 )
 
 s.move(library / "google/cloud/spanner_v1/proto")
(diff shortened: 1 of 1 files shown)


busunkim96:
@larkee Could you try upgrading the synthtool version? `include_protos` was added fairly recently to `GAPICBazel`

```sh
python3 -m pip install --user --upgrade git+https://github.com/googleapis/synthtool.git
```

larkee:
> @larkee Could you try upgrading the synthtool version? `include_protos` was added fairly recently to `GAPICBazel`
> 
> ```shell
> python3 -m pip install --user --upgrade git+https://github.com/googleapis/synthtool.git
> ```

Ahhh. I tried to do an upgrade through pip but I didn't think to use the git link. It works now :+1:

PR #75, opened 2020-05-01
refactor: PingingPool pings sessions using SELECT 1

larkee:
Currently, PingingPool pings sessions in the background by calling `session.exists()` which calls `GetSession`. 

Using `SELECT 1` is preferred and is used in other client libraries such as [Go](https://github.com/googleapis/google-cloud-go/blob/53898305c6f21b3c3eef34fcff6c61a2cb36f602/spanner/session.go#L227):

### DIFF
--- a/google/cloud/spanner_v1/pool.py
+++ b/google/cloud/spanner_v1/pool.py
@@ -387,9 +387,7 @@ class PingingPool(AbstractSessionPool):
         ping_after, session = self._sessions.get(block=True, timeout=timeout)
 
         if _NOW() > ping_after:
-            try:
-                session.ping()
-            except NotFound:
+            if not session.exists():
                 session = self._new_session()
                 session.create()
 
--- a/google/cloud/spanner_v1/pool.py
+++ b/google/cloud/spanner_v1/pool.py
@@ -387,6 +387,9 @@ class PingingPool(AbstractSessionPool):
         ping_after, session = self._sessions.get(block=True, timeout=timeout)
 
         if _NOW() > ping_after:
+            # Using session.exists() guarantees the returned session exists.
+            # session.ping() uses a cached result in the backend which could
+            # result in a recently deleted session being returned.
             if not session.exists():
                 session = self._new_session()
                 session.create()
--- a/google/cloud/spanner_v1/pool.py
+++ b/google/cloud/spanner_v1/pool.py
@@ -23,7 +23,6 @@ from google.cloud.spanner_v1._helpers import _metadata_with_prefix
 
 
 _NOW = datetime.datetime.utcnow  # unit tests may replace
-_SESSION_EXPIRY_DURATION = datetime.timedelta(hours=1)
 
 
 class AbstractSessionPool(object):
@@ -385,13 +384,14 @@ class PingingPool(AbstractSessionPool):
         if timeout is None:
             timeout = self.default_timeout
 
-        last_used, session = self._sessions.get(block=True, timeout=timeout)
+        ping_after, session = self._sessions.get(block=True, timeout=timeout)
 
-        if _NOW() > last_used + _SESSION_EXPIRY_DURATION:
-            session = self._new_session()
-            session.create()
-        elif _NOW() > last_used + self._delta:
-            session.ping()
+        if _NOW() > ping_after:
+            try:
+                session.ping()
+            except NotFound:
+                session = self._new_session()
+                session.create()
 
         return session
 
@@ -405,7 +405,7 @@ class PingingPool(AbstractSessionPool):
 
         :raises: :exc:`six.moves.queue.Full` if the queue is full.
         """
-        self._sessions.put_nowait((_NOW(), session))
+        self._sessions.put_nowait((_NOW() + self._delta, session))
 
     def clear(self):
         """Delete all sessions in the pool."""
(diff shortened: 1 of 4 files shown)


google/cloud/spanner_v1/pool.py
```diff
@@ -23,6 +23,7 @@
 
 
 _NOW = datetime.datetime.utcnow  # unit tests may replace
+_SESSION_EXPIRY_DURATION = datetime.timedelta(hours=1)
```
hengfengli, line 26: Is this the time interval between two consecutive `ping()`? Then, I think `50mins` would be better.

hengfengli: Ignore my previous comment. I see this is the expire duration for a session.

larkee: This is the time after which a session is considered expired as the backend will have GC'd it.

In the [constructor for PingingPool](https://github.com/googleapis/python-spanner/blob/a9bf3e579b5b8f64f9549a22d28f51ff8781fd83/google/cloud/spanner_v1/pool.py#L345) there is a `ping_interval` kwarg which is the time between two consecutive pings. The default for that is 3000 seconds ie 50 minutes.

google/cloud/spanner_v1/pool.py
```diff
@@ -384,12 +385,13 @@ def get(self, timeout=None):  # pylint: disable=arguments-differ
...
             timeout = self.default_timeout
 
-        ping_after, session = self._sessions.get(block=True, timeout=timeout)
+        last_used, session = self._sessions.get(block=True, timeout=timeout)
 
-        if _NOW() > ping_after:
-            if not session.exists():
-                session = self._new_session()
-                session.create()
+        if _NOW() > last_used + _SESSION_EXPIRY_DURATION:
```
hengfengli, line 390: Here, my question is that even if the expire duration (1 hr) has passed, we don't know whether or not the session has expired. We need to verify that the session does not exist before creating a new one. 

In the previous code, it calls `session.exists` to verify if a session exists. 

Can we call `ping()` first? If the session has been deleted, then we create a new session.

google/cloud/spanner_v1/pool.py
```diff
@@ -423,16 +425,18 @@ def ping(self):
...
             except queue.Empty:  # all sessions in use
                 break
-            if ping_after > _NOW():  # oldest session is fresh
+            if last_used + self._delta > _NOW():  # oldest session is fresh
                 # Re-add to queue with existing expiration
-                self._sessions.put((ping_after, session))
+                self._sessions.put((last_used, session))
                 break
-            if not session.exists():  # stale
+            if _NOW() > last_used + _SESSION_EXPIRY_DURATION:  # stale
```
hengfengli, line 435: Same here. 

What will happen to `session.ping()` if session is deleted?

google/cloud/spanner_v1/pool.py
```diff
@@ -423,16 +425,18 @@ def ping(self):
         """
         while True:
             try:
-                ping_after, session = self._sessions.get(block=False)
+                last_used, session = self._sessions.get(block=False)
             except queue.Empty:  # all sessions in use
                 break
-            if ping_after > _NOW():  # oldest session is fresh
+            if last_used + self._delta > _NOW():  # oldest session is fresh
```
hengfengli, line 431: Does the old code have a bug here? In the old code, it doesn't add `self._delta`.

google/cloud/spanner_v1/pool.py
larkee: In original testing, pinging a session that had just been deleted would lead to no error. However, adding a delay raises the expected NotFound error. So since the likelihood that it will have just been deleted is low, I'll just rely on the ping to raise the NotFound error.

I'll refactor to rely on ping raising NotFound.

google/cloud/spanner_v1/pool.py
larkee: In the original, the stored value was _NOW() + self._delta which is why it was called ping after. I changed the stored value to _NOW() so I could tell when a session had been unused for an hour.

larkee:
> LGTM. So when a session is deleted, calling `session.ping()` will cause a NotFound error, is this right?

Yes. However there seems to be a delay before the error appears. Given the following code, the error only occurs when DELAY  is over ~35s

```
from time import sleep

client = spanner.Client()
instance = client.instance(INSTANCE_ID)
database = instance.database(DATABASE_ID)
session = database.session()
session.create()
session.delete()
sleep(DELAY)
session.ping()
```

hengfengli:
It makes sense. The deletion needs time to be propagated.

So if a session is deleted, when we still execute a transaction on it, what will happen then? When it gets a SessionNotFound error, will it try to get a new session?

larkee:
> It makes sense. The deletion needs time to be propagated.
> 
> So if a session is deleted, when we still execute a transaction on it, what will happen then? When it gets a SessionNotFound error, will it try to get a new session?

Currently, it will not. However, if we keep the `session.exists()` call in the `get()` then we don't have to worry about users encountering this error. Since, that isn't called in the background I think it's still acceptable.

larkee:
> LGTM. So now, `pool.get()` uses `session.exists()` to verify the validness of a session instead of `session.ping()`. Can we leave a comment over there why we don't use `session.ping()`?
> 
> The reason why we don't use `session.ping()` is that it has a cached/outdated result, which is inaccurate, is this correct?

Good suggestion! Done :)

Yes, the backend uses a cached value which means a delete may have occurred but not propagated yet.

PR #80, opened 2020-05-11
test: remove unneeded system test skip when using emulator

larkee:
The functionality in test_partition_query is supported by the emulator. We should not be skipping it.

### DIFF
--- a/tests/system/test_system.py
+++ b/tests/system/test_system.py
@@ -2330,7 +2330,6 @@ class TestSessionAPI(unittest.TestCase, _TestData):
             # NaNs cannot be searched for by equality.
             self.assertTrue(math.isnan(float_array[2]))
 
-    @unittest.skipIf(USE_EMULATOR, "Skipping partitioned queries")
     def test_partition_query(self):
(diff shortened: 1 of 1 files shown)


larkee:
> LGTM. When did they start supporting partitioned queries?

Unclear. However, the test does not give an UNIMPLEMENTED error. It instead give the following error:
```
E       AssertionError: Lists differ: [[1, 'First000000001', 'Last000000039', 'te[2782 chars]om']] != [[0, 'First000000000', 'Last000000040', 'te[2782 chars]om']]
E       
E       First differing element 0:
E       [1, 'First000000001', 'Last000000039', 'test-000000001@example.com']
E       [0, 'First000000000', 'Last000000040', 'test-000000000@example.com']
E       
E       Diff is 4732 characters long. Set self.maxDiff to None to see it.
```
So the functionality is implemented.

larkee:
The above test failure is related to results order. There is no guarantee of order without using ORDER BY which is not supported by partitioned queries. I have a fix which I will make as a separate PR.

PR #81, opened 2020-05-13
test: fix result ordering errors for emulator

larkee:
While running the ITs against the Cloud Spanner emulator there were failures relating to result ordering. Spanner doesn't guarantee an order for its results unless specified by the ORDER BY clause. The emulator deliberately shuffles the results so users do not rely on the ordering without ORDER BY. 

This PR updates a test to use ORDER BY to fix an emulator failure. However, since ORDER BY is not supported by partitioned queries, a set of the results is created for comparison in that test instead.


hengfengli:
@larkee The kokoro test fails. Please fix it so that I can review this PR. Thanks.

PR #86, opened 2020-05-15
feat: add support for NUMERIC type

larkee:


tests/unit/test__helpers.py
```diff
@@ -431,6 +440,17 @@ def test_w_struct(self):
 
         self.assertEqual(self._callFUT(value_pb, field_type), VALUES)
 
+    def test_w_numeric(self):
+        import decimal
+        from google.protobuf.struct_pb2 import Value
+        from google.cloud.spanner_v1.proto.type_pb2 import Type, NUMERIC
+
+        VALUE = decimal.Decimal("99999999999999999999999999999.999999999")
```
hengfengli, line 448: Why name it with uppercases? Should it be `value`?

larkee: It should, however I was matching with similar tests. I'll make a separate PR to fix this for all the tests.

hengfengli: You can try to rename `from google.protobuf.struct_pb2 import Value` via: 

```
from relative_module import identifier as name
```

Or, change `VALUE` to `v` or `val`.

tests/_fixtures.py
```diff
@@ -67,4 +67,55 @@
      PRIMARY KEY(id, commit_ts DESC);
 """
 
+EMULATOR_DDL = """\
```
hengfengli, line 70: Is this added because Emulator doesn't support Numeric yet? But it means that if the emulator supports Numeric in the future, we need to remove these code, right?

larkee: > Is this added because Emulator doesn't support Numeric yet?

Yes. The alternative would be skipping a large number of tests when using the emulator which isn't ideal.

> But it means that if the emulator supports Numeric in the future, we need to remove these code, right?

Potentially. We could also set it to equal `DDL_STATEMENTS` and retain the emulator specific code rather than remove it in case there are features in the future which are not supported by the emulator initially.

hengfengli (approved):
LGTM.

skuruppu (approved):

PR #87, opened 2020-05-21
feat: add support for using the emulator programatically

larkee:
Currently, we do not support setting the emulator host outside of the env flag `SPANNER_EMULATOR_HOST`. 

This PR allows users to set the emulator host by using the `api_endpoint` in `client_options` with `google.auth.credentials.AnonymousCredentials`

e.g.
```

cred = AnonymousCredentials()
client_options = {'api_endpoint': '0.0.0.0:9010'}
client = spanner.Client(
    credentials=cred,
    client_options=client_options,
    project='emulator-project'
)
```

### DIFF
--- a/google/cloud/spanner_v1/client.py
+++ b/google/cloud/spanner_v1/client.py
@@ -28,6 +28,7 @@ import os
 import warnings
 
 from google.api_core.gapic_v1 import client_info
+from google.auth.credentials import AnonymousCredentials
 import google.api_core.client_options
 
 from google.cloud.spanner_admin_instance_v1.gapic.transports import (
@@ -173,6 +174,21 @@ class Client(ClientWithProject):
         client_options=None,
         query_options=None,
     ):
+        self._emulator_host = _get_spanner_emulator_host()
+
+        if client_options and type(client_options) == dict:
+            self._client_options = google.api_core.client_options.from_dict(
+                client_options
+            )
+        else:
+            self._client_options = client_options
+
+        if isinstance(credentials, AnonymousCredentials):
+            self._emulator_host = self._client_options.api_endpoint
+
+        if not credentials and _get_spanner_emulator_host():
+            credentials = AnonymousCredentials()
+
         # NOTE: This API has no use for the _http argument, but sending it
         #       will have no impact since the _http() @property only lazily
         #       creates a working HTTP object.
@@ -180,12 +196,6 @@ class Client(ClientWithProject):
             project=project, credentials=credentials, _http=None
         )
         self._client_info = client_info
-        if client_options and type(client_options) == dict:
-            self._client_options = google.api_core.client_options.from_dict(
-                client_options
-            )
-        else:
-            self._client_options = client_options
 
         env_query_options = ExecuteSqlRequest.QueryOptions(
             optimizer_version=_get_spanner_optimizer_version()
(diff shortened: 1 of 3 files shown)


google/cloud/spanner_v1/client.py
```diff
@@ -173,19 +174,28 @@ def __init__(
...
+        self._emulator_host = _get_spanner_emulator_host()
+
+        if client_options and type(client_options) == dict:
+            self._client_options = google.api_core.client_options.from_dict(
+                client_options
+            )
+        else:
+            self._client_options = client_options
+
+        if isinstance(credentials, AnonymousCredentials):
```
snehashah16, line 186: for other libraries, we may have done the opposite.
If the emulator_env_variable is set, we hard-code anonymous/null credentials for the emulator.

google/cloud/spanner_v1/client.py
```diff
@@ -173,19 +174,27 @@ def __init__(
...
+            self._client_options = google.api_core.client_options.from_dict(
+                client_options
+            )
+        else:
+            self._client_options = client_options
+
+        if self._emulator_host:
+            credentials = AnonymousCredentials()
+        elif isinstance(credentials, AnonymousCredentials):
+            self._emulator_host = self._client_options.api_endpoint
```
hengfengli, line 189: I am a bit unsure about the logic here: if SPANNER_EMULATOR_HOST is not set and an AnonymousCredentials is provided, then we set `_emulator_host` as `api_endpoint`. 

I think copying from `api_endpoint` to`_emulator_host` is redundant. If I specify `AnonymousCredentials` but not giving an `api_endpoint`, it will raise an error because the production backend does not support it. 

Why do we need this?

google/cloud/spanner_v1/database.py
```diff
@@ -223,9 +223,7 @@ def spanner_api(self):
                     channel=grpc.insecure_channel(self._instance.emulator_host)
                 )
                 self._spanner_api = SpannerClient(
-                    client_info=client_info,
-                    client_options=client_options,
-                    transport=transport,
+                    client_info=client_info, transport=transport
```
hengfengli, line 226: Why we don't pass `client_options` here? Will it cause any problem?

tests/unit/test_client.py
```diff
@@ -250,18 +254,44 @@ def test_instance_admin_api(self, mock_em):
         credentials.with_scopes.assert_called_once_with(expected_scopes)
 
     @mock.patch("google.cloud.spanner_v1.client._get_spanner_emulator_host")
-    def test_instance_admin_api_emulator(self, mock_em):
+    def test_instance_admin_api_emulator_env(self, mock_em):
```
hengfengli, line 257: I can see there are tests for instance and admin clients. But how about the spanner data clients?

google/cloud/spanner_v1/client.py
larkee: So the point of this is so that the correct gRPC channel is created. When using the emulator, `insecure_channel` needs to be used. We use `secure_channel` by default. So we check `self._emulator_host` to know which channel type to use.

google/cloud/spanner_v1/client.py
larkee: I have updated it to set AnonymousCredentials when the env var is set.

google/cloud/spanner_v1/database.py
larkee: Spanner only really uses api_endpoint which we don't need because we have passed it. But I'll add it back in since new options could be supported in future.

tests/unit/test_client.py
larkee: The emulator host is passed as an argument to the instance. This means for spanner there is no difference between using the flag or using AnonymousCredentials. There is an existing test for setting emulator_host: `test_spanner_api_w_emulator_host`

google/cloud/spanner_v1/client.py
hengfengli: Okay. I see your point. In Go, we directly give `api_endpoint`, `AnonymousCredentials` and `insecure channel` together: https://github.com/googleapis/google-cloud-go/blob/master/spanner/admin/database/apiv1/init.go#L28-L33.

PR #94, opened 2020-06-04
docs: update documentation for snapshot usage

larkee:
This PR fixes the links to other documentation, updates wording and fixes typos and grammar for `docs/snapshot-usage.rst`.


larkee:
> LGTM. I just wonder if there is any limit for line length in rst. I notice that on Line 89-92, the line length is quite different from previous lines.

I believe that the query has been split like that to make it easier to understand rather than due to a line length limit.

PR #97, opened 2020-06-15
test: update system tests for emulator

larkee:
This PR allows system tests to be run against the emulator without setting `GOOGLE_APPLICATION_CREDENTIALS` with no warnings.

It also skips `test_transaction_read_w_abort` which fails due to running concurrent transactions which the emulator does not support, and removes skips for `test_execute_partitioned_dml` and `test_partition_read_w_index` which are now supported.

### DIFF
--- a/tests/system/test_system.py
+++ b/tests/system/test_system.py
@@ -93,7 +93,11 @@ def _list_instances():
 
 
 def setUpModule():
-    Config.CLIENT = Client()
+    if USE_EMULATOR:
+        from google.auth.credentials import AnonymousCredentials
+        Config.CLIENT = Client(project='emulator-test-project', credentials=AnonymousCredentials())
+    else:
+        Config.CLIENT = Client()
     retry = RetryErrors(exceptions.ServiceUnavailable)
 
     configs = list(retry(Config.CLIENT.list_instance_configs)())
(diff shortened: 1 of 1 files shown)


tests/system/test_system.py
```diff
@@ -93,7 +93,11 @@ def _list_instances():
 
 
 def setUpModule():
-    Config.CLIENT = Client()
+    if USE_EMULATOR:
+        from google.auth.credentials import AnonymousCredentials
+        Config.CLIENT = Client(project='emulator-test-project', credentials=AnonymousCredentials())
```
hengfengli, line 98: I wonder can we read the project from an env var instead of using a hard-coded one?

tests/system/test_system.py
```diff
@@ -93,7 +93,12 @@ def _list_instances():
 
 
 def setUpModule():
-    Config.CLIENT = Client()
+    if USE_EMULATOR:
+        from google.auth.credentials import AnonymousCredentials
+        emulator_project = os.getenv("GCLOUD_PROJECT") or 'emulator-test-project'
```
hengfengli, line 98: nits: I think you can do `os.getenv("GCLOUD_PROJECT", "emulator-test-project")`.

PR #98, opened 2020-06-19
ci: set up kokoro system tests to run on separate instances

### LINKED ISSUE #83: Backup tests are flaky
The Kokoro tests are often failing on backup tests for unrelated changes. There are two problems:
- `UpdateBackup` sometimes times out with DEADLINE_EXCEEDED as seen in https://github.com/GoogleCloudPlatform/python-docs-samples/issues/3241
- `test_list_backups` is failing due to a backup from a different test being included in the returned list for `size_bytes`

The first problem can be resolved by increasing the timeout for `UpdateBackup`.

The second problem is difficult to replicate and the exact cause is unclear given that the tests are not run in parallel and the backups are being deleted at the end of each test. The simplest solution will be to modify the test to ensure that no backups from previous tests meet the condition.

larkee:
Closes #83

### DIFF
--- a/.kokoro/build.sh
+++ b/.kokoro/build.sh
@@ -29,7 +29,7 @@ export GOOGLE_APPLICATION_CREDENTIALS=${KOKORO_GFILE_DIR}/service-account.json
 # Setup project id.
 export PROJECT_ID=$(cat "${KOKORO_GFILE_DIR}/project-id.json")
 
-# Set up creating a new instance for each system tests run
+# Set up creating a new instance for each system test run
 export GOOGLE_CLOUD_TESTS_CREATE_SPANNER_INSTANCE=true
 
 # Remove old nox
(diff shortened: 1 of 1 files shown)


.kokoro/build.sh
```diff
@@ -29,6 +29,9 @@ export GOOGLE_APPLICATION_CREDENTIALS=${KOKORO_GFILE_DIR}/service-account.json
 # Setup project id.
 export PROJECT_ID=$(cat "${KOKORO_GFILE_DIR}/project-id.json")
 
+# Set up system tests to run on different instances.
+export GOOGLE_CLOUD_TESTS_CREATE_SPANNER_INSTANCE=true
```
hengfengli, line 33: Does this mean that we will create new spanner instances for system tests instead of using the existing one? The comment is not quite consistent with the variable name. The name is "create spanner instance" and the comment is "run on different instances". Could you please elaborate the comment a bit?

larkee: I have updated the comment to match more closely to the variable name. Does the new comment clarify things?

PR #105, opened 2020-06-25
feat: add support for CMEK

larkee:

### DIFF
--- a/google/cloud/spanner_v1/database.py
+++ b/google/cloud/spanner_v1/database.py
@@ -17,6 +17,7 @@
 import copy
 import functools
 import grpc
+import logging
 import re
 import threading
 
@@ -123,13 +124,13 @@ class Database(object):
     _spanner_api = None
 
     def __init__(
-            self,
-            database_id,
-            instance,
-            ddl_statements=(),
-            pool=None,
-            logger=None,
-            encryption_config=None,
+        self,
+        database_id,
+        instance,
+        ddl_statements=(),
+        pool=None,
+        logger=None,
+        encryption_config=None,
     ):
         self.database_id = database_id
         self._instance = instance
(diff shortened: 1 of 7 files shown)


google/cloud/spanner_v1/database.py
```diff
@@ -104,6 +112,13 @@ def __init__(self, database_id, instance, ddl_statements=(), pool=None):
         self._create_time = None
         self._restore_info = None
 
+        if type(encryption_config) == dict:
+            self._encryption_config = EncryptionConfig(
+                kms_key_name=encryption_config["kms_key_name"]
+            )
+        else:
```
hengfengli, line 119: Would it happen that a user passes a `str` ("kms_key") into this function? From what I can see, `encryption_config` only allows two types: `dict` and `EncryptionConfig`. Should we say if type is not a `EncryptionConfig`, then we should raise an InvalidArgument error?

google/cloud/spanner_admin_database_v1/gapic/database_admin_client.py
```diff
@@ -100,6 +101,17 @@ def backup_path(cls, project, instance, backup):
             backup=backup,
         )
 
+    @classmethod
```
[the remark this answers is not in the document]
larkee: Yes, this is a generated file. There isn't a surefire way to know if a file is generated unfortunately which is why I've been caught out a couple times but generally all files contained in a `gapic/` folder are generator.

google/cloud/spanner_v1/database.py
larkee: I've added some documentation explaining the permitted formats of the `encryption_config`. I have not added a TypeError as it is inconsistent with other similar conversions in the library which do not raise errors. However, I may look into adding these errors in a future PR.

google/cloud/spanner_v1/database.py
```diff
@@ -95,7 +96,14 @@ class Database(object):
```
skuruppu, line 96: I think there should be a docs parameter here for `encryption_config`. This will also clarify the types question below.

larkee: Done.

PR #107, opened 2020-07-02
feat: add OpenTelemetry tracing to spanner calls

cnnradams:
1. Creates spans around batch/snapshot/transaction calls with relevant information that could help debug latency issues + provide insight into spanner calls
2. Removed python2.7 from unit/system tests (since OpenTelemetry does not support 2.7). However since OT is an optional dependency with some extra effort the tests could keep using 2.7, is this desired?

Pool/gRPC metrics are also on my TODO list, they will come in a seperate PR once some issues with the OpenTelemetry Metrics SDK are worked out :)

### DIFF
--- a/docs/opentelemetry-tracing.rst
+++ b/docs/opentelemetry-tracing.rst
@@ -1,0 +1,36 @@
+Tracing with OpenTelemetry
+==================================
+Python-spanner uses `OpenTelemetry <https://opentelemetry.io/>`_ to automatically generates traces providing insight on calls to Cloud Spanner. 
+For information on the benefits and utility of tracing, see the `Cloud Trace docs <https://cloud.google.com/trace/docs/overview>`_.
+
+To take advantage of these traces, we first need to install opentelemetry:
+
+.. code-block:: sh
+
+    pip install opentelemetry-api opentelemetry-sdk opentelemetry-instrumentation
+
+We also need to tell OpenTelemetry which exporter to use. For example, to export python-spanner traces to `Cloud Tracing <https://cloud.google.com/trace>`_, add the following lines to your application:
+
+.. code:: python
+
+    from opentelemetry import trace
+    from opentelemetry.sdk.trace import TracerProvider
+    from opentelemetry.trace.sampling import ProbabilitySampler
+    from opentelemetry.exporter.cloud_trace import CloudTraceSpanExporter
+    # BatchExportSpanProcessor exports spans to Cloud Trace 
+    # in a seperate thread to not block on the main thread
+    from opentelemetry.sdk.trace.export import BatchExportSpanProcessor
+
+    # create and export one trace every 1000 requests
+    sampler = ProbabilitySampler(1/1000)
+    # Uses the default tracer provider
+    trace.set_tracer_provider(TracerProvider(sampler=sampler))
+    trace.get_tracer_provider().add_span_processor(
+        # initialize the cloud tracing exporter
+        BatchExportSpanProcessor(CloudTraceSpanExporter())
+    )
+
+Generated spanner traces should now be available on `Cloud Trace <https://console.cloud.google.com/traces>`_.
+
+Tracing is most effective when many libraries are instrumented to provide insight over the entire lifespan of a request.
+For a list of libraries that can be instrumented, see the `OpenTelemetry Integrations` section of the `OpenTelemetry Python docs <https://opentelemetry-python.readthedocs.io/en/stable/>`_
--- a/docs/opentelemetry-tracing.rst
+++ b/docs/opentelemetry-tracing.rst
@@ -1,6 +1,6 @@
 Tracing with OpenTelemetry
 ==================================
-Python-spanner uses `OpenTelemetry <https://opentelemetry.io/>`_ to automatically generates traces providing insight on calls to Cloud Spanner. 
+This library uses `OpenTelemetry <https://opentelemetry.io/>`_ to automatically generate traces providing insight on calls to Cloud Spanner. 
 For information on the benefits and utility of tracing, see the `Cloud Trace docs <https://cloud.google.com/trace/docs/overview>`_.
 
 To take advantage of these traces, we first need to install opentelemetry:
--- a/docs/opentelemetry-tracing.rst
+++ b/docs/opentelemetry-tracing.rst
@@ -21,12 +21,12 @@ We also need to tell OpenTelemetry which exporter to use. For example, to export
     # in a seperate thread to not block on the main thread
     from opentelemetry.sdk.trace.export import BatchExportSpanProcessor
 
-    # create and export one trace every 1000 requests
+    # Create and export one trace every 1000 requests
     sampler = ProbabilitySampler(1/1000)
-    # Uses the default tracer provider
+    # Use the default tracer provider
     trace.set_tracer_provider(TracerProvider(sampler=sampler))
     trace.get_tracer_provider().add_span_processor(
-        # initialize the cloud tracing exporter
+        # Initialize the cloud tracing exporter
         BatchExportSpanProcessor(CloudTraceSpanExporter())
     )
 
--- a/google/cloud/spanner_v1/_opentelemetry_tracing.py
+++ b/google/cloud/spanner_v1/_opentelemetry_tracing.py
@@ -1,0 +1,65 @@
+# Copyright 2016 Google LLC All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Manages OpenTelemetry trace creation and handling"""
+
+from contextlib import contextmanager
+
+from google.api_core.exceptions import GoogleAPICallError
+from google.cloud.spanner_v1.gapic import spanner_client
+
+try:
+    from opentelemetry import trace
+    from opentelemetry.trace.status import Status, StatusCanonicalCode
+    from opentelemetry.instrumentation.utils import http_status_to_canonical_code
+
+    HAS_OPENTELEMETRY_INSTALLED = True
+except ImportError:
+    HAS_OPENTELEMETRY_INSTALLED = False
+
+
+@contextmanager
+def trace_call(name, session, extra_attributes=None):
+    if not HAS_OPENTELEMETRY_INSTALLED:
+        # empty context manager. users will have to check if the generated value is None or a span
+        yield None
+        return
+
+    tracer = trace.get_tracer(__name__)
+
+    # base attributes that we know for every trace created
+    attributes = {
+        "db.type": "spanner",
+        "db.url": spanner_client.SpannerClient.SERVICE_ADDRESS,
+        "db.instance": session._database.name,
+        "net.host.name": spanner_client.SpannerClient.SERVICE_ADDRESS,
+    }
+
+    if extra_attributes:
+        attributes.update(extra_attributes)
+
+    with tracer.start_as_current_span(
+        name, kind=trace.SpanKind.CLIENT, attributes=attributes
+    ) as span:
+        try:
+            yield span
+        except GoogleAPICallError as error:
+            if error.code is not None:
+                span.set_status(Status(http_status_to_canonical_code(error.code)))
+            elif error.grpc_status_code is not None:
+                span.set_status(
+                    # OpenTelemetry's StatusCanonicalCode maps 1-1 with grpc status codes
+                    Status(StatusCanonicalCode(error.grpc_status_code.value[0]))
+                )
+            raise
--- a/google/cloud/spanner_v1/_opentelemetry_tracing.py
+++ b/google/cloud/spanner_v1/_opentelemetry_tracing.py
@@ -31,7 +31,7 @@ except ImportError:
 
 @contextmanager
 def trace_call(name, session, extra_attributes=None):
-    if not HAS_OPENTELEMETRY_INSTALLED:
+    if not HAS_OPENTELEMETRY_INSTALLED or not session:
         # empty context manager. users will have to check if the generated value is None or a span
         yield None
         return
--- a/google/cloud/spanner_v1/_opentelemetry_tracing.py
+++ b/google/cloud/spanner_v1/_opentelemetry_tracing.py
@@ -1,4 +1,4 @@
-# Copyright 2016 Google LLC All rights reserved.
+# Copyright 2020 Google LLC All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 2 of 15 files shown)


google/cloud/spanner_v1/_opentelemetry_tracing.py
```diff
@@ -0,0 +1,65 @@
...
+        # empty context manager. users will have to check if the generated value is None or a span
+        yield None
+        return
+
+    tracer = trace.get_tracer(__name__)
+
+    # base attributes that we know for every trace created
+    attributes = {
+        "db.type": "spanner",
+        "db.url": spanner_client.SpannerClient.SERVICE_ADDRESS,
```
larkee, line 44: Users can specify different API endpoints. Currently, I'm not clear on whether we want the tracing to support using different endpoints. @hengfengli Do the other libraries use user given endpoints when tracing?

google/cloud/spanner_v1/_opentelemetry_tracing.py
```diff
@@ -0,0 +1,65 @@
+# Copyright 2016 Google LLC All rights reserved.
```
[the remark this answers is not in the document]
cnnradams: fixed

google/cloud/spanner_v1/transaction.py
```diff
@@ -268,13 +277,18 @@ def batch_update(self, statements):
...
-        response = api.execute_batch_dml(
-            session=self._session.name,
-            transaction=transaction,
-            statements=parsed,
-            seqno=seqno,
-            metadata=metadata,
-        )
+        trace_attributes = {
+            # Get just the queries from the DML statement batch
+            "db.statement": [statement[0] for statement in statements]
```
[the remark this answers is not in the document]
cnnradams: fixed

tests/system/test_system.py
```diff
@@ -1226,6 +1331,64 @@ def test_transaction_batch_update_wo_statements(self):
             with self.assertRaises(InvalidArgument):
                 transaction.batch_update([])
 
+    def test_transaction_batch_update_w_parent_span(self):
+        import sys
```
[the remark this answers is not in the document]
cnnradams: fixed

tests/_helpers.py
```diff
@@ -0,0 +1,35 @@
...
+        self.assertEqual(len(span_list), 0)
+
+    def assertSpanAttributes(
+        self, name, status=StatusCanonicalCode.OK, attributes=None, span=None
+    ):
+        if not span:
+            span_list = self.memory_exporter.get_finished_spans()
+            self.assertEqual(len(span_list), 1)
+            span = span_list[0]
+        print(status, attributes, span.status, span.attributes)
```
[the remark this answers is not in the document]
cnnradams: missed that, fixed

google/cloud/spanner_v1/snapshot.py
```diff
@@ -30,17 +30,22 @@
...
+def _restart_on_unavailable(restart, trace_name=None, session=None, attributes=None):
     """Restart iteration after :exc:`.ServiceUnavailable`.
 
     :type restart: callable
     :param restart: curried function returning iterator
     """
     resume_token = b""
     item_buffer = []
-    iterator = restart()
+    if trace_name and session:
```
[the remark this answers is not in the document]
cnnradams: good idea, changed it

tests/unit/test_session.py
```diff
@@ -35,6 +41,12 @@ class TestSession(unittest.TestCase):
     DATABASE_NAME = INSTANCE_NAME + "/databases/" + DATABASE_ID
     SESSION_ID = "session-id"
     SESSION_NAME = DATABASE_NAME + "/sessions/" + SESSION_ID
+    BASE_ATTRIBUTES = {
+        "db.type": "spanner",
+        "db.url": "spanner.googleapis.com:443",
+        "db.instance": "projects/project-id/instances/instance-id/databases/database-id",
```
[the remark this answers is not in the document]
cnnradams: fixed

tests/unit/test__opentelemetry_tracing.py
```diff
@@ -0,0 +1,139 @@
...
+from opentelemetry import trace as trace_api
+from opentelemetry.trace.status import StatusCanonicalCode
+from opentelemetry.sdk.trace import TracerProvider, export
+from opentelemetry.sdk.trace.export.in_memory_span_exporter import InMemorySpanExporter
+
+from google.api_core.exceptions import GoogleAPICallError
+from google.cloud.spanner_v1 import _opentelemetry_tracing
+
+
+class OpenTelemetryBase(unittest.TestCase):
```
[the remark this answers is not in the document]
cnnradams: yep, I just forgot to remove this one when I moved it into _helpers

noxfile.py
```diff
@@ -83,13 +89,13 @@ def default(session):
...
 
 
-@nox.session(python=["2.7", "3.5", "3.6", "3.7", "3.8"])
+@nox.session(python=["3.5", "3.6", "3.7", "3.8"])
 def unit(session):
     """Run the unit test suite."""
     default(session)
 
 
-@nox.session(python=["2.7", "3.7"])
```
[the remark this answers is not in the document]
cnnradams: added back 2.7 support - just had to wrap all the OT tests in a dependency check and not install OT if using 2.7

tests/_helpers.py
```diff
@@ -0,0 +1,50 @@
+import unittest
+from unittest import mock
```
larkee, line 2: The tests fail for Python 2.7 due to this import.

```suggestion
import mock
```

The system tests for Python 2.7 still fail with this fix. You can run the system tests locally [using nox](https://github.com/googleapis/python-spanner/blob/master/CONTRIBUTING.rst#using-nox) to help with debugging.

cnnradams: whoops, should be good now (all tests pass locally) - thought I could get away with not installing 2.7 :P

google/cloud/spanner_v1/_opentelemetry_tracing.py
hengfengli: In Go, it uses OpenCensus/trace and I don't see a way to use user given endpoints: https://github.com/googleapis/google-cloud-go/blob/master/internal/trace/trace.go.

hengfengli: It seems that in Go, spans are not annotated. In the design doc, I can see that both `db.url` and `net.host.name` are endpoint urls, spanner.googleapis.com:443 by default.

cnnradams: > Currently, I'm not clear on whether we want the tracing to support using different endpoints.

Not sure what you mean by this? Yes, the endpoint would be recorded as an attribute on the trace. But that doesn't affect anything about where tracing data goes, which is handled by a user-specified exporter not set up here? What is the downside of recording the endpoint here?

google/cloud/spanner_v1/session.py
```diff
@@ -114,7 +115,11 @@ def create(self):
         kw = {}
         if self._labels:
             kw = {"session": {"labels": self._labels}}
-        session_pb = api.create_session(self._database.name, metadata=metadata, **kw)
+
+        with trace_call("CloudSpanner.CreateSession", self, self._labels):
```
hengfengli, line 119: I wonder do we have a uniform format for name? 

- In Go, we have a format like `cloud.google.com/go/spanner.ReadWriteTransaction` (https://github.com/googleapis/google-cloud-go/blob/e143982c0485e1ff2726f7e2b607c26997dd0d90/spanner/client.go#L423). 
- In Java, we have a format "CloudSpannerOperation.Commit" (https://github.com/googleapis/java-spanner/blob/462839b625e58e235581b8ba10b398e1d222eaaf/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerImpl.java#L59-L67). 

I guess it is okay to use this format because every language has a different format.

google/cloud/spanner_v1/snapshot.py
```diff
@@ -243,7 +249,13 @@ def execute_sql(
             timeout=timeout,
         )
 
-        iterator = _restart_on_unavailable(restart)
+        trace_attributes = {"db.statement": sql}
```
hengfengli, line 252: So we just directly expose `sql` to traces without any processing, right? We probably should document this somewhere to let users know if they have PII data or data privacy concerns. 

I wonder why do we need this `db.statement`? How do we come up with the decision to add this field?

google/cloud/spanner_v1/transaction.py
```diff
@@ -268,13 +277,18 @@ def batch_update(self, statements):
...
-        response = api.execute_batch_dml(
-            session=self._session.name,
-            transaction=transaction,
-            statements=parsed,
-            seqno=seqno,
-            metadata=metadata,
-        )
+        trace_attributes = {
+            # Get just the queries from the DML statement batch
+            "db.statement": [statement["sql"] for statement in parsed]
```
hengfengli, line 282: Previously, `sql` in `trace_attributes = {"db.statement": sql}` is a string. Here, we pass a list of strings as a parameter. Will this cause an issue?

tests/_helpers.py
```diff
@@ -0,0 +1,50 @@
...
+    from opentelemetry.trace.status import StatusCanonicalCode
+
+    from opentelemetry.sdk.trace import TracerProvider, export
+    from opentelemetry.sdk.trace.export.in_memory_span_exporter import (
+        InMemorySpanExporter,
+    )
+
+    HAS_OPENTELEMETRY_INSTALLED = True
+except ImportError:
+    HAS_OPENTELEMETRY_INSTALLED = False
```
hengfengli, line 15: Why cannot OpenTelemetry be installed? This adds a lot of checks `if HAS_OPENTELEMETRY_INSTALLED` in the following code. Is it possible we can get rid of this? 

If `OpenTelemetry` is not installed, the test will fail and I think this should be fine.

tests/system/test_system.py
```diff
@@ -2333,7 +2513,7 @@ def test_execute_sql_returning_transfinite_floats(self):
                 )
             )
             self.assertEqual(len(rows), 1)
-            float_array, = rows[0]
+            (float_array,) = rows[0]
```
hengfengli, line 2516: Just curious: why `()` is required?

tests/unit/test__opentelemetry_tracing.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+def _make_session():
+    from google.cloud.spanner_v1.session import Session
+
+    return mock.Mock(autospec=Session, instance=True)
+
+
+# Skip all of these tests if we don't have OpenTelemetry
+if HAS_OPENTELEMETRY_INSTALLED:
```
hengfengli, line 33: I think we should get rid of this. Instead, we should make sure that OpenTelemetry is installed. If not, the test should fail.

tests/_helpers.py
cnnradams: OpenTelemetry can't be installed for python2.7, so as long as we're running those tests this needs to be here

google/cloud/spanner_v1/snapshot.py
cnnradams: `db.statement` is mentioned as a "required" attribute in the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md#call-level-attributes) (although we can do whatever we want). We could sanitize the data for PII ourselves, although that's probably a bad idea. Personally I'm fine with removing it but I think it would be a very useful attribute for debugging. And correct me if I'm wrong, but if done right the SQL should be split into the query (statement) and actual values (params) right? And since we only upload the query it could be fine?

tests/system/test_system.py
cnnradams: It's not, I think my auto linter added it for some reason. removed

google/cloud/spanner_v1/transaction.py
cnnradams: Good point, not sure what the right answer is - could remove it, only publish the first statement, or merge them together?

google/cloud/spanner_v1/_opentelemetry_tracing.py
```diff
@@ -0,0 +1,65 @@
...
+@contextmanager
+def trace_call(name, session, extra_attributes=None):
+    if not HAS_OPENTELEMETRY_INSTALLED or not session:
+        # empty context manager. users will have to check if the generated value is None or a span
+        yield None
+        return
+
+    tracer = trace.get_tracer(__name__)
+
+    # base attributes that we know for every trace created
```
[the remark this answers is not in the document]
cnnradams: fixed all of the above

tests/unit/test__opentelemetry_tracing.py
cnnradams: see above comment on py2.7, same thing applies here

noxfile.py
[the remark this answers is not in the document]
hengfengli: @larkee Do you have an estimate when will python 2.7 be deprecated? 

If we need to merge this first before deprecating 2.7, then we need to remember to remove the dependency checks in OT tests when 2.7 is deprecated.

google/cloud/spanner_v1/session.py
hengfengli: Go is a bit special. Can you match it with `Java`? Thanks for the update.

google/cloud/spanner_v1/snapshot.py
hengfengli: I think I am fine with it. 

You're right. It is supposed to split query and params. However, it might happen that users just compose a query string and pass an empty params. Not all users will follow the best practice.

google/cloud/spanner_v1/transaction.py
hengfengli: Probably merge them together and use `;` for a delimiter?

tests/system/test_system.py
hengfengli: So why don't we write: 

```py3
float_array = rows[0][0]
```

The above way is still confusing, which is usually used for tuples.

noxfile.py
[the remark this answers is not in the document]
larkee: Python 2.7 will be deprecated with the microgenerator migration which is currently scheduled for the end of August. I am happy to remove the dependency checks during that migration myself.

google/cloud/spanner_v1/_opentelemetry_tracing.py
larkee: > What is the downside of recording the endpoint here?

There's no downside. There are just certain cases, such as using the emulator, where the endpoint on the span wouldn't match the endpoint being used. It's a pretty minor concern though so I'm okay with leaving this as is.

google/cloud/spanner_v1/transaction.py
larkee: > Probably merge them together and use ; for a delimiter?

This option makes the most sense :+1:

tests/system/test_system.py
larkee: > ... nevermind. the system tests fail if this isn't there.

Interesting. I haven't had any issues with this line before. However, I do agree it is a confusing way to write this.

google/cloud/spanner_v1/transaction.py
cnnradams: done

tests/system/test_system.py
cnnradams: fixed, went with rows[0][0]

cnnradams:
what would it take to get this PR moving again?

hengfengli:
@larkee It looks good to me. Can we merge this in?

docs/opentelemetry-tracing.rst
```diff
@@ -0,0 +1,36 @@
+Tracing with OpenTelemetry
+==================================
+Python-spanner uses `OpenTelemetry <https://opentelemetry.io/>`_ to automatically generates traces providing insight on calls to Cloud Spanner.
```
larkee, line 3:
```suggestion
This library uses `OpenTelemetry <https://opentelemetry.io/>`_ to automatically generates traces providing insight on calls to Cloud Spanner. 
```

The documentation is already associated with the library so it shouldn't need to be stated.

PR #114, opened 2020-07-14
ci: add GitHub action for emulator system tests

larkee:


.github/workflows/integration-tests-against-emulator.yaml
```diff
@@ -0,0 +1,32 @@
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+name: integration-tests-against-emulator
```
hengfengli, line 6: Do we need to rename this `name` field?

larkee: I'll rename it to match PHP

PR #122, opened 2020-08-04
fix: resume iterator on EOS internal error

### LINKED ISSUE #120: spanner-client: Retry PDML on "Received unexpected EOS on DATA frame from server"
This bug is related to the Spanner client library.

For long lived transactions (>= 30 minutes), in the case of large PDML changes, it is possible that the gRPC connection is terminated with an error "Received unexpected EOS on DATA frame from server".

In this case, we need to retry the transaction either with the received resume token obtained on reading the stream or from scratch. This will ensure that the PDML transaction continues to execute until it is successful or a hard timeout is reached.

We have already implemented such change in the Java client library, for more information see this PR: https://github.com/googleapis/java-spanner/pull/360.

In order to test the fix, we can use a large spanner database. Please speak to @thiagotnunes for more details.

larkee:
Fixes #120

This fixes the same issue as googleapis/java-spanner#360.

A similar issue was encountered by `bigquery-storage` and was fixed in googleapis/google-cloud-python#9994.

### DIFF
--- a/google/cloud/spanner_v1/snapshot.py
+++ b/google/cloud/spanner_v1/snapshot.py
@@ -20,7 +20,7 @@ from google.protobuf.struct_pb2 import Struct
 from google.cloud.spanner_v1.proto.transaction_pb2 import TransactionOptions
 from google.cloud.spanner_v1.proto.transaction_pb2 import TransactionSelector
 
-from google.api_core.exceptions import ServiceUnavailable
+from google.api_core.exceptions import ServiceUnavailable, InternalServerError
 import google.api_core.gapic_v1.method
 from google.cloud._helpers import _datetime_to_pb_timestamp
 from google.cloud.spanner_v1._helpers import _merge_query_options
@@ -32,6 +32,9 @@ from google.cloud.spanner_v1.streamed import StreamedResultSet
 from google.cloud.spanner_v1.types import PartitionOptions
 from google.cloud.spanner_v1._opentelemetry_tracing import trace_call
 
+_STREAM_RESUMPTION_INTERNAL_ERROR_MESSAGES = (
+    "RST_STREAM",
+)
 
 def _restart_on_unavailable(restart, trace_name=None, session=None, attributes=None):
     """Restart iteration after :exc:`.ServiceUnavailable`.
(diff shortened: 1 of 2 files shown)


google/cloud/spanner_v1/snapshot.py
```diff
@@ -55,6 +60,17 @@ def _restart_on_unavailable(restart, trace_name=None, session=None, attributes=N
             with trace_call(trace_name, session, attributes):
                 iterator = restart(resume_token=resume_token)
             continue
+        except InternalServerError as exc:
+            resumable_error = any(
+                resumable_message in exc.message
+                for resumable_message in _STREAM_RESUMPTION_INTERNAL_ERROR_MESSAGES
```
hengfengli, line 66: `resumable_error = exc.message in _STREAM_RESUMPTION_INTERNAL_ERROR_MESSAGES`? 

Even better: 

```python
if exc.message not in _STREAM_RESUMPTION_INTERNAL_ERROR_MESSAGES:
    raise
...
```

thiagotnunes:
@larkee thanks for the fix, I will test it out with a long running PDML to make sure everything we need is in place!

google/cloud/spanner_v1/snapshot.py
```diff
@@ -20,7 +20,7 @@
 from google.cloud.spanner_v1.proto.transaction_pb2 import TransactionOptions
 from google.cloud.spanner_v1.proto.transaction_pb2 import TransactionSelector
 
-from google.api_core.exceptions import ServiceUnavailable
+from google.api_core.exceptions import ServiceUnavailable, InternalServerError
```
tseaver, line 23: We normally do one import per line, e.g.:
```suggestion
from google.api_core.exceptions import InternalServerError
from google.api_core.exceptions import ServiceUnavailable
```

thiagotnunes:
@larkee Quick update, the PDML long running task finished successfully in production testing. Thanks for fixing this!

google/cloud/spanner_v1/snapshot.py
larkee: I believe the intention here is that `_STREAM_RESUMPTION_INTERNAL_ERROR_MESSAGES` contains error substrings. This mean we don't need to include multiple entries for `RST_STREAM` errors. I know we haven't encountered those errors in Spanner yet but I think it's worth including since the rationale appears to be the same.

hengfengli: Okay. I see your point.

hengfengli (approved):
LGTM. 👍

thiagotnunes (approved):

PR #145, opened 2020-09-22
DO NOT MERGE: test with pre-release google-cloud-core

tswast:
This pull request is to check that the changes made in https://github.com/googleapis/python-cloud-core/pull/28 do not break this library.

Unit tests pass locally.


tswast:
Closing as all tests passed without modification.

I was curious why this library uses google-cloud-core, even though it's a gRPC client. It appears to mostly be for the type conversion helpers in https://github.com/googleapis/python-cloud-core/blob/master/google/cloud/_helpers.py

PR #147, opened 2020-10-02
feat!: migrate to v2.0.0

larkee:
This PR is a preview of the 2.0 release.

**TODO list:**
- [x] Fix docs.

- [x] Describe changes in UPGRADING guide.

BREAKING CHANGES
- `list_instances`, `list_databases`, `list_instance_configs`, and `list_backups` will now return protos rather than the handwritten wrapper


busunkim96:
Thank you for working on this!

* I noticed there doesn't seem to be a samples presubmit. I opened 335755839 to add those kokoro configs internally.


    > Handwritten files have been moved from google/cloud/spanner_v1 to google/cloud/spanner

    I believe the other handwritten/generated mixes keep the generated surface in the versioned directory. What is the 
    motivation for the change?
    https://github.com/googleapis/python-firestore/tree/master/google/cloud/firestore_v1
    https://github.com/googleapis/python-pubsub/tree/v2.1.0/google/cloud/pubsub_v1

* I'm not familiar with the Spanner client, are there particular files we should direct our attention to?

larkee:
> * I noticed there doesn't seem to be a samples presubmit. I opened 335755839 to add those kokoro configs internally.

Thanks! I wasn't sure what was missing.

>   > Handwritten files have been moved from google/cloud/spanner_v1 to google/cloud/spanner
>   
>   I believe the other handwritten/generated mixes keep the generated surface in the versioned directory. What is the
>   motivation for the change?
>   https://github.com/googleapis/python-firestore/tree/master/google/cloud/firestore_v1
>   https://github.com/googleapis/python-pubsub/tree/v2.1.0/google/cloud/pubsub_v1

The motivation was two fold. It separates the handwritten surfaces which makes it clearer to contributes what is written and what is generated. It also prevents having to customize the generated `google/cloud/spanner_v1/__init__.py`. However given that this file shouldn't change too much, it's probably reasonable to exclude it from generation and manually update it when required.

I'm happy to revert these files back to their original folder if you think that would be best.

> * I'm not familiar with the Spanner client, are there particular files we should direct our attention to?

Most of the handwritten surfaces are used directly by users. The major ones are:

- `client.py`
- `instance.py`
- `database.py`
- `backup.py`
- `batch.py
- `snapshot.py`
- `transaction.py`

However, the files that changed the most during migration are:

- `_helpers.py`
- `snapshot.py`
- `streamed.py`
- `transaction.py`

google/cloud/spanner/snapshot.py
```diff
@@ -150,17 +152,18 @@ def read(self, table, columns, keyset, index="", limit=0, partition=None):
...
-            api.streaming_read,
-            self._session.name,
-            table,
-            columns,
-            keyset._to_pb(),
+        request = ReadRequest(
+            session=self._session.name,
+            table=table,
+            columns=columns,
+            key_set=keyset._to_pb(),
```
busunkim96, line 159: Is this conversion still necessary?

busunkim96 (commented):
I would prefer the handwritten code live in the versioned directory for consistency with the other libraries. It looks like firestore excludes `__init__.py` in `synth.py` https://github.com/googleapis/python-firestore/blob/c3acd4a04745c93edb2f61bf9be6fa33f439f4b0/synth.py#L41

Fully generated libraries use the unversioned path `google/cloud/{api}/` for  the default alias (equivalent of `google/cloud/spanner.py`) https://github.com/googleapis/python-speech/blob/master/google/cloud/speech/__init__.py so there's also a small chance for confusion there.

google/cloud/spanner/snapshot.py
larkee: Yes. `keyset` is a handwritten wrapper for the `KeySet` proto. Arguably, with the `proto-plus` generated types, this handwritten wrapper could be removed and the `KeySet` proto could be used directly. My only concern about doing this is whether the error messages from the proto for invalid inputs would be as clear and helpful.

larkee: I did a trial refactor where the handwritten `KeySet` and `KeyRange` were removed and ran into some issues with type conversion due to Spanner expecting a `string_value` for `INT64` values. So at this stage, I think keeping these handwritten wrappers makes sense.

google/cloud/spanner_v1/streamed.py
```diff
@@ -121,7 +119,7 @@ def _consume_next(self):
             if source is not None and source._transaction_id is None:
                 source._transaction_id = metadata.transaction.id
 
-        if response.HasField("stats"):  # last response
+        if type(response).pb(response).HasField("stats"):  # last response
```
busunkim96, line 122: Does `"stats" in response` fail? `proto.Message` has `__contains__` so I would expect it to work.

https://github.com/googleapis/proto-plus-python/blob/b12971a9265e5f35010677fb0d9e1c0eab43320b/proto/message.py#L472

busunkim96 (approved):

google/cloud/spanner.py
```diff
@@ -12,37 +12,36 @@
...
+from google.cloud.spanner_v1 import COMMIT_TIMESTAMP
 
 
 __all__ = (
+    # google.cloud.spanner
     "__version__",
-    "AbstractSessionPool",
-    "BurstyPool",
+    "param_types",
+    # google.cloud.spanner_v1.client
```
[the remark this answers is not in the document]
larkee: These are to indicate where the imports are from. This style came from `google/cloud/spanner_v1/__init__.py`. I can remove these comments if they're against the style guide.

google/cloud/spanner_v1/streamed.py
larkee: `"stats" in response` works. Updated :+1:

UPGRADING.md
```diff
@@ -0,0 +1,274 @@
...
+ )
+ ```
+ 
+ ```py
+ response = client.execute_sql(
+     parent=project_name,
+ )
+ ```
+ 
+ This call is invalid because it mixes `request` with a keyword argument `params`. Executing this code
```
skuruppu, line 210:
```suggestion
 This call is invalid because it mixes `request` with a keyword argument `parent`. Executing this code
```

skuruppu (approved):
I only reviewed UPGRADING.md so my LGTM is for that file.

wyardley:
@larkee: In the end, the "breaking change" that shows up in the changelog was
`migrate to v2.0.0 (#147)` vs

`list_instances, list_databases, list_instance_configs, and list_backups will now return protos rather than the handwritten wrapper`

which would have been much more useful

larkee:
@wyardley My apologies! I'll do my best to label changes better in future :+1: I'll update the changelog directly to show the more useful message

PR #148, opened 2020-10-07
feat: add support for Point In Time Recovery (PITR)

larkee:
Implements support for PITR.

With this functionality users will be able to:
- set the retention period for their databases.
- set version time of the database when creating a backup

Creating a backup with a specific version time is shown below:
```
backup = instance.backup(backup_id, expire_time=expire_time, version_time=version_time)
backup.create()
```

### DIFF
--- a/tests/system/test_system.py
+++ b/tests/system/test_system.py
@@ -355,6 +355,64 @@ class TestDatabaseAPI(unittest.TestCase, _TestData):
         database_ids = [database.name for database in Config.INSTANCE.list_databases()]
         self.assertIn(temp_db.name, database_ids)
 
+    @unittest.skipIf(
+        USE_EMULATOR, "PITR-lite features are not supported by the emulator"
+    )
+    def test_create_database_pitr_invalid_retention_period(self):
+        pool = BurstyPool(labels={"testcase": "create_database_pitr"})
+        temp_db_id = "temp_db" + unique_resource_id("_")
+        retention_period = "0d"
+        ddl_statements = [
+            "ALTER DATABASE {}"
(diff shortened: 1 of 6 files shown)


thiagotnunes (approved):
LGTM

tests/system/test_system.py
```diff
@@ -558,9 +670,16 @@ def test_backup_workflow(self):
         backup_id = "backup_id" + unique_resource_id("_")
         expire_time = datetime.utcnow() + timedelta(days=3)
         expire_time = expire_time.replace(tzinfo=UTC)
+        version_time = datetime.utcnow() - timedelta(seconds=5)
```
thiagotnunes, line 673: There is an assumption here that the instance is 5 seconds old. I think this is a fair assumption, but just want to make sure that we know the database that we are backing up is always at least 5 seconds old when we reach this, otherwise we can get flaky tests.

tests/system/test_system.py
```diff
@@ -691,9 +889,14 @@ def test_list_backups(self):
         instance = Config.INSTANCE
         expire_time_1 = datetime.utcnow() + timedelta(days=21)
         expire_time_1 = expire_time_1.replace(tzinfo=UTC)
+        version_time_1 = datetime.utcnow() - timedelta(minutes=5)
```
thiagotnunes, line 892: There is an assumption here that the instance is 5 minutes old. I think this is a fair assumption, but just want to make sure that we know the database that we are backing up is always at least 5 minutes old when we reach this, otherwise we can get flaky tests.

thiagotnunes:
LGTM

PR #152, opened 2020-10-13
fix: handle Unmergable errors when merging struct responses

larkee:
This PR fixes a bug in the `StreamedResultSet` logic for merging chunked values where the chunked value is an array of structs and the chunk boundary splits a struct element in two and the halves aren't correctly merged.

This bug results in the struct element incorrectly being split into two partial struct elements.

**Expected:**
```
('foo', 0.5, DatetimeWithNanoseconds(2020, 1, 1, 0, 0, tzinfo=<UTC>))
```

**Actual:**
```
('foo', 0.5)
('2020-01-01T00:00:00Z',)
```

This bug is caused by an unhandled `Unmergable` error in `_merge_struct`. This PR adds the same handling used for `_merge_array` and a unit test to prove the fix works.

### DIFF
--- a/google/cloud/spanner_v1/streamed.py
+++ b/google/cloud/spanner_v1/streamed.py
@@ -286,7 +286,13 @@ def _merge_struct(lhs, rhs, type_):
         lhs.append(first)
     else:
         last = lhs.pop()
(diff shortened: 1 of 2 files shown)


thiagotnunes (approved):
LGTM

PR #160, opened 2020-10-30
feat: DB-API driver + unit tests

mf2199:
The DB-API driver refactored from [python-spanner-django](https://github.com/q-logic/python-spanner-django) repository, prior to its beta-release.

_Some additional changes have been made to skip unit tests found incompatible with Python 2._


c24t (approved):
LGTM, we'll a PR in https://github.com/googleapis/python-spanner-django to remove `spanner_dbapi` and then coordinate releasees. We can follow up with tests for 2.x.

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -0,0 +1,333 @@
...
+                # is returned or many, yet mixing results that
+                # are for .fetchone() with those that would result in
+                # many items returns a RuntimeError if .fetchone() is
+                # invoked and vice versa.
+                self._result_set = res
+                # Read the first element so that the StreamedResultSet can
+                # return the metadata after a DQL statement. See issue #155.
+                self._itr = PeekIterator(self._result_set)
+                # Unfortunately, Spanner doesn't seem to send back
+                # information about the number of rows available.
```
larkee, line 290: Yes, unfortunately the row count is sent as part of `ResultSetStats` which is not available until the final `ResultSet` has been received.

See: https://github.com/googleapis/python-spanner/blob/master/google/cloud/spanner_v1/streamed.py#L46

google/cloud/spanner_dbapi/parse_utils.py
```diff
@@ -0,0 +1,546 @@
...
+    """Convert the param to its Cloud Spanner equivalent type.
+
+    :type value: Any
+    :param value: Value to convert to a Cloud Spanner type.
+
+    :rtype: Any
+    :returns: Value converted to a Cloud Spanner type.
+    """
+    if isinstance(value, decimal.Decimal):
+        return float(value)
```
larkee, line 497: Converting `Decimal` to `float` can lose precision. Please use `str` instead. 

If you have a good reason for not doing so, you will instead need to clearly document that NUMERICs may lose precision when using this driver.

larkee (changes requested):
LGTM :+1: Just one concern about NUMERIC loss of precision.

google/cloud/spanner_dbapi/_helpers.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+def _execute_insert_heterogenous(transaction, sql_params_list):
+    for sql, params in sql_params_list:
+        sql, params = sql_pyformat_args_to_spanner(sql, params)
+        param_types = get_param_types(params)
+        res = transaction.execute_sql(sql, params=params, param_types=param_types)
+        # TODO: File a bug with Cloud Spanner and the Python client maintainers
+        # about a lost commit when res isn't read from.
+        _ = list(res)
```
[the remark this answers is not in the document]
mf2199: Totally, thanks for the clarification. Done.

google/cloud/spanner_dbapi/parse_utils.py
mf2199: Changed as suggested, based on this [recommendation](https://cloud.google.com/spanner/docs/storing-numeric-data#recommendation_store_arbitrary_precision_numbers_as_strings).

larkee (approved):

PR #171, opened 2020-11-19
fix: Add sqlparse dependency

c24t:
#160 added `sqlparse` as a test dependency only, but included [non-test code that imports it](https://github.com/googleapis/python-spanner/blob/3db4e5f8d729fb35d416691489d13c2fd306361a/google/cloud/spanner_dbapi/parse_utils.py#L22). Running the tests outside of nox gives this error: `E   ModuleNotFoundError: No module named 'sqlparse'`.

This PR makes `sqlparse` a regular non-test dependency.

The version number comes from https://github.com/googleapis/python-spanner-django/commit/2b096c50949db674e2c017cb4e05e9f71bc0e405.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,8 +32,9 @@ dependencies = [
     "google-api-core[grpc] >= 1.22.0, < 2.0.0dev",
     "google-cloud-core >= 1.4.1, < 2.0dev",
     "grpc-google-iam-v1 >= 0.12.3, < 0.13dev",
-    "proto-plus==1.11.0",
     "libcst >= 0.2.5",
+    "proto-plus == 1.11.0",
+    "sqlparse >= 0.3.0",
 ]
 extras = {
     "tracing": [
--- a/noxfile.py
+++ b/noxfile.py
@@ -72,7 +72,7 @@ def default(session):
     # Install all test dependencies, then install this package in-place.
     session.install("asyncmock", "pytest-asyncio")
 
-    session.install("mock", "pytest", "pytest-cov", "sqlparse")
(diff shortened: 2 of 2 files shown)


mf2199 (approved):
👍 Good catch! This is something I wasn't sure about adding on a global scale.

setup.py
```diff
@@ -32,8 +32,9 @@
     "google-api-core[grpc] >= 1.22.0, < 2.0.0dev",
     "google-cloud-core >= 1.4.1, < 2.0dev",
     "grpc-google-iam-v1 >= 0.12.3, < 0.13dev",
-    "proto-plus==1.11.0",
     "libcst >= 0.2.5",
+    "proto-plus == 1.11.0",
+    "sqlparse >= 0.3.0",
```
larkee, line 37: `sqlparse` is only used for `dbapi` and is not required for using the Cloud Spanner client. I think it makes more sense to have this as a separate extra dependency similar to the `OpenTelemetry` dependencies under "tracing".

c24t: The problem with making this an optional dependency is that we've got code in the library that will fail to import without it. The library still works without tracing, it just doesn't emit traces.

If we want to avoid sqlparse as a required dependency I see two options: (1) make it an optional import and catch the import errors in the DBAPI package and log an error, or (2) make `spanner_dbapi` a separate installable package in this repo, with its own set of requirements.

larkee: As discussed I think (2) is the best option

larkee (approved):
LGTM. I would prefer to have `spanner_dbapi` as a separate installable before doing a release but I won't block you on this.

PR #177, opened 2020-11-30
fix(dbapi): autocommit enabling fails if no transactions begun

IlyaFaer:
When enabling an `autocommit` mode, all the begun transactions must be commited, but while trying to commit a transaction we don't check if it is already committed/rollbacked.


c24t (approved):
Note to other reviewers, this addresses a bug in the django tests: https://github.com/googleapis/python-spanner-django/pull/559/checks?check_run_id=147277725, but the fix isn't specific to django.

@IlyaFaer LGTM, but please confirm that the django tests pass with this change, and that this is the right autocommit behavior according to the DBAPI spec and sqlite implementation before merging this one.

IlyaFaer:
@c24t, this behavior is mimicked from the Java implementation, because in Python (and its DB APIs) we didn't find any strict directions of what should be done on enabling an autocommit mode. Right now, we're trying to commit the current transaction without checking its state, so errors are raised in several cases: when the transaction was normally commited, and we enabling an autocommit mode; when it was rolled back, and we're enabling autocommit mode; when the transaction wasn't even started. With this PR we'll first be checking if the transaction is in progress - in this case it should be commited.

c24t:
FYI @larkee, @skuruppu this brings back the "dummy where clause" in https://github.com/googleapis/python-spanner/pull/177/files#diff-e942c1913bf8770f329585a42ab8ce8abe3d2109a0a77431ed7a8f337f467425R534. There's some more conversation on this in https://github.com/googleapis/python-spanner-django/pull/516#discussion_r497215446. I wanted to get your OK on this before merging.

@IlyaFaer is working on some tests that touch this now, and found that inserting the dummy clause is a more difficult change in spanner-django than in the DB API layer here. Since spanner-django is the only consumer at the moment this wouldn't change anything in practice, but it is the wrong place for it in principle. If you're willing to approve it for now we'll merge and open a ticket to revert the change.

larkee (approved):
> If you're willing to approve it for now we'll merge and open a ticket to revert the change.

Approving under the mentioned condition :+1:

PR #193, opened 2020-12-15
feat: add support for instance labels

### LINKED ISSUE #190: Add support for instance labels
Cloud Spanner supports adding labels to resources such as instances which can be used for filtering. Currently, the Python library does not allow users to set or get the labels through the provided surface. Adding this support would allow instances created for running system tests to be labelled. This would allow instances from previous system test runs which were not deleted to be cleaned up as part of the testing setup.

larkee:
Closes #190


google/cloud/spanner_v1/client.py
```diff
@@ -313,6 +314,9 @@ def instance(
         :param node_count: (Optional) The number of nodes in the instance's
                             cluster; used to set up the instance's cluster.
 
+        :type labels: str
```
skuruppu, line 317: In Node.js, the type is `labels?: {[k: string]: string} | null`. It seems that below you expect it to be a dict. Should it be made explicit here? Otherwise, if it's just a string, then it should be `label`.

google/cloud/spanner_v1/instance.py
```diff
@@ -99,6 +99,15 @@ class Instance(object):
                          Cloud Console UI. (Must be between 4 and 30
                          characters.) If this value is not set in the
                          constructor, will fall back to the instance ID.
+
+    :type emulator_host: str
+    :param emulator_host: (Optional) The address for the Cloud Spanner emulator
+                          that the database of this instance should connect to.
+                          If this value is not set, the instance will connect
+                          to the Cloud Spanner endpoint.
```
skuruppu, line 107: I'm curious why this is added here now.

skuruppu (changes requested):

google/cloud/spanner_v1/instance.py
larkee: While adding the `labels` parameter docstring I noticed it was missing and figured the right thing would be to add it. Should this be done in a separate PR?

google/cloud/spanner_v1/client.py
larkee: Whoops! Good catch! I'll update it

google/cloud/spanner_v1/instance.py
skuruppu: Yeah probably, just to make sure that it's easy to revert the labels change if necessary.

skuruppu (approved):

PR #195, opened 2020-12-16
test: improve cleanup to prevent RESOURCE_EXHAUSTED error

larkee:
Currently, there is no guarantee that the resources created during the system tests are cleaned up as the jobs running the tests can be cancelled before finishing. This means that over time the testing project runs out of resources due to undeleted instances which causes all system tests runs to fail.

This PR adds labels to the instance created for the test so that any instance created for a test that is older than an hour can be identified and deleted. This will keep the testing project from running out of resources over time.


IlyaFaer (approved):
Looks okay to me. Thanks!

c24t (approved):

PR #199, opened 2020-12-21
fix: Convert PBs in system test cleanup

### LINKED ISSUE #198: System test failures cleaning up backups
`instance.list_backups` fails at this line: https://github.com/googleapis/python-spanner/blob/2b74f9c9dc45bcaefed44697833d65fa44f95ade/tests/system/test_system.py#L125

because `instance` is the protobuf type `google.cloud.spanner_admin_instance_v1.types.spanner_instance_admin.Instance` instead of `google.cloud.spanner_v1.instance.Instance`. This is causing the following error in CI:

```
During handling of the above exception, another exception occurred:

    def setUpModule():
        if USE_EMULATOR:

            emulator_project = os.getenv("GCLOUD_PROJECT", "emulator-test-project")
            Config.CLIENT = Client(
                project=emulator_project, credentials=AnonymousCredentials()
            )
        else:
            Config.CLIENT = Client()
        retry = RetryErrors(exceptions.ServiceUnavailable)

        configs = list(retry(Config.CLIENT.list_instance_configs)())

        instances = retry(_list_instances)()
        EXISTING_INSTANCES[:] = instances

        # Delete test instances that are older than an hour.
        cutoff = int(time.time()) - 1 * 60 * 60
        for instance in Config.CLIENT.list_instances("labels.python-spanner-systests:true"):
            if "created" not in instance.labels:
                continue
            create_time = int(instance.labels["created"])
            if create_time > cutoff:
                continue
            # Instance cannot be deleted while backups exist.
>           for backup in instance.list_backups():

tests/system/test_system.py:125:
```

See e.g. [this failing test run](https://source.cloud.google.com/results/invocations/cbfa924e-6586-47e4-b36d-f3328c482ff3/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-spanner%2Fpresubmit%2Fpresubmit/log).

It looks like it was introduced in #195, but the tests passed for that PR. The API changes to `list_instances` and `list_backups` that would have caused this happened back in #147 (see https://github.com/googleapis/python-spanner/pull/147/files#diff-f9e7537fc73135ee5d350541c1147e8ce8c71c505b01c9ea1187f9ee80540b19R328-R356), so I'm not sure why we're only seeing this issue now.

c24t:
This PR converts some Spanner API resource protos following https://github.com/googleapis/python-spanner/blob/master/UPGRADING.md#handwritten-surface-changes, and fixes an apparent bug in the system test.

Fixes #198 🦕

### DIFF
--- a/tests/system/test_system.py
+++ b/tests/system/test_system.py
@@ -12,16 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from pprint import pformat
 import collections
 import datetime
 import decimal
-import logging
 import math
 import operator
 import os
 import struct
-import sys
 import threading
 import time
 import unittest
@@ -53,11 +50,6 @@ from test_utils.system import unique_resource_id
 from tests._fixtures import DDL_STATEMENTS
 from tests._fixtures import EMULATOR_DDL_STATEMENTS
 from tests._helpers import OpenTelemetryBase, HAS_OPENTELEMETRY_INSTALLED
-
-
-logger = logging.getLogger(__name__)
-logger.addHandler(logging.StreamHandler(sys.stderr))
-logger.setLevel(logging.INFO)
 
 
 CREATE_INSTANCE = os.getenv("GOOGLE_CLOUD_TESTS_CREATE_SPANNER_INSTANCE") is not None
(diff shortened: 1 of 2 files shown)


larkee (approved):
Whoops! I can't believe I missed this! Thanks 👍

c24t:
I suspect the CI errors around tracing spans are because we're creating spans for retries, but now that spanner is healthy we're seeing green builds without any retries (at least in the affected tests, assuming I'm right about the retry spans). If this comes up again we can cherry pick c0923e1 to investigate. I merged #201 in here too, hopefully this gets us a fully green build.

PR #200, opened 2020-12-22
fix: Introspect to get column types in cursor.execute

c24t:
Addresses https://github.com/googleapis/python-spanner-django/issues/566.

This PR adds an extra call to get the table's schema before making inserts from `cursor.execute`. Spanner requires us to 
pass [parameter types to `Transaction.execute_update`](https://github.com/c24t/python-spanner/blob/094c0cd1780cc144d381d7229e5a23d8e15f8c4c/google/cloud/spanner_v1/transaction.py#L190-L193), before this change we inferred the Spanner column types from the native python types in the prepared statement. Besides being a hack, this meant we couldn't handle inserts with null values. See https://github.com/googleapis/python-spanner-django/issues/566 for more detail.

This change fixes many django test failures, notably the `modeladmin` and (most of) `expressions_case` tests, which all failed with `InvalidArgument('Unable to infer type for parameter ....')` before.

Note that `Connection.run_statement` still calls `_execute_insert_heterogenous` without `param_types`, and this still results in a call to `get_param_types`. I didn't replace this because it didn't affect the django tests I was running. If we pursue this change we should remove `get_param_types` altogether next.

There may be other problems with this approach that don't show up in the tests. For example, this change assumes the schema doesn't change between reading `INFORMATION_SCHEMA.COLUMNS` and doing the insert. Doing an extra read op before each insert may also slow the client down.

### DIFF
--- a/google/cloud/spanner_dbapi/cursor.py
+++ b/google/cloud/spanner_dbapi/cursor.py
@@ -175,18 +175,40 @@ class Cursor(object):
                 sql, params = sql_pyformat_args_to_spanner(sql, args)
 
                 statement = Statement(
-                    sql, params, get_param_types(params), ResultsChecksum(),
+                    sql,
+                    params,
+                    get_param_types(params),
+                    ResultsChecksum(),
                 )
-                (self._result_set, self._checksum,) = self.connection.run_statement(
-                    statement
-                )
+                (
+                    self._result_set,
+                    self._checksum,
+                ) = self.connection.run_statement(statement)
                 self._itr = PeekIterator(self._result_set)
                 return
 
             if classification == parse_utils.STMT_NON_UPDATING:
                 self._handle_DQL(sql, args or None)
             elif classification == parse_utils.STMT_INSERT:
-                _helpers.handle_insert(self.connection, sql, args or None)
+
+                # Read INFORMATION_SCHEMA.COLUMNS to get the Spanner types for
+                # the columns included in this insert
+                table_name, columns = parse_utils.get_table_cols_for_insert(
+                    sql
+                )
+                schema = self.get_table_column_schema(table_name)
+                column_type_names = [
+                    schema[col].spanner_type.split("(")[0] for col in columns
+                ]
+                param_types = [
+                    parse_utils.COL_TYPE_NAME_TO_TYPE[ctn]
+                    for ctn in column_type_names
+                ]
+
+                _helpers.handle_insert(
+                    self.connection, sql, args or None, param_types=param_types
+                )
+
             else:
                 self.connection.database.run_in_transaction(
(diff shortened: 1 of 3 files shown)


c24t:
> Note that `Connection.run_statement` still calls `_execute_insert_heterogenous` without `param_types`, and this still results in a call to `get_param_types`.

Running more of the django tests, it looks like this is responsible for another error:

```
TypeError: Parameter to MergeFrom() must be instance of same class: expected google.spanner.v1.Type got NoneType.
```

This shows up in `model_inheritance` among other tests.

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -175,18 +175,40 @@ def execute(self, sql, args=None):
...
+                ) = self.connection.run_statement(statement)
                 self._itr = PeekIterator(self._result_set)
                 return
 
             if classification == parse_utils.STMT_NON_UPDATING:
                 self._handle_DQL(sql, args or None)
             elif classification == parse_utils.STMT_INSERT:
-                _helpers.handle_insert(self.connection, sql, args or None)
+
+                # Read INFORMATION_SCHEMA.COLUMNS to get the Spanner types for
```
olavloite, line 194: I would strongly recommend that you only do this as a workaround when running against the emulator, as real Spanner seems to accept the parameter type to be missing. Otherwise you will be slowing down statements unnecessarily in production environments.
Furthermore, you probably need to do this for all types of statements on the emulator, not only insert statements.

IlyaFaer:
I suppose this PR can be closed. It adds some redundant code, which will be in production, but will work only for emulator cases. On SQLAlchemy we've just skipped the tests, which are not working on emulator, like this:
https://github.com/googleapis/python-spanner-sqlalchemy/blob/e97325de221d201a306ba92fecf8afc9856fa2c5/test/test_suite_13.py#L343-L345

I think let's do the same for Django. @olavloite, @larkee, what do you say?

olavloite:
> I suppose this PR can be closed. It adds some redundant code, which will be in production, but will work only for emulator cases. On SQLAlchemy we've just skipped the tests, which are not working on emulator, like this: https://github.com/googleapis/python-spanner-sqlalchemy/blob/e97325de221d201a306ba92fecf8afc9856fa2c5/test/test_suite_13.py#L343-L345
> 
> I think let's do the same for Django. @olavloite, @larkee, what do you say?

Yes, this PR is more than a year old, and I would have serious doubts about the performance impacts of this. In addition, real Spanner supports non-typed parameters, and the missing support for non-typed parameters for some type (DATE and TIMESTAMP) in the emulator is considered a bug: https://github.com/GoogleCloudPlatform/cloud-spanner-emulator/issues/31

PR #201, opened 2020-12-22
fix: Rename to fix "Mismatched region tag" check

c24t:
Quick fix for failing "Mismatched region tag" checks like [this one](https://github.com/googleapis/python-spanner/pull/199/checks?check_run_id=1593297712).

I didn't change the names of any class-level constants, just the ones defined in individual tests. Even though we only need to rename "START" and "END" to fix the check, I thought it would be better to make them all consistent.


larkee (approved):
LGTM 👍 Good call on the consistency!

PR #203, opened 2020-12-30
chore: add constraints file

busunkim96:
Add constraints file to test lower bounds. These files will not be used until the noxfile is changed in googleapis/synthtool#869.

### DIFF
--- a/testing/constraints-3.6.txt
+++ b/testing/constraints-3.6.txt
@@ -1,0 +1,16 @@
+# This constraints file is used to check that lower bounds
+# are correct in setup.py
+# List *all* library dependencies and extras in this file.
+# Pin the version to the lower bound.
+#
+# e.g., if setup.py has "foo >= 1.14.0, < 2.0.0dev",
+# Then this file should have foo==1.14.0
+google-api-core==1.22.0
+google-cloud-core==1.4.1
+grpc-google-iam-v1==0.12.3
+libcst==0.2.5
+proto-plus==1.13.0
+sqlparse==0.3.0
+opentelemetry-api==0.11b0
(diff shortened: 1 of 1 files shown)


c24t (approved):
Looks fine to me, but how come there's only a constraints file for 3.6? This project uses 3.8 as the default for tests, and doesn't run any tests for python versions >3.8.

Maybe a problem for another PR, but since the constraints files are just mechanically built from the requirements in setup.py, could synthtool keep the constraints files up to date instead of doing this manually?

busunkim96:
> Looks fine to me, but how come there's only a constraints file for 3.6? This project uses 3.8 as the default for tests, and doesn't run any tests for python versions >3.8.

Hey Chris! Right now the unit and system tests in the libraries always fetch the latest versions of dependencies.  It's led to a few instances where we forgot to bump the minimum required version of some dependency (`google-api-core` for example).  The idea is to test the lower bound in one of the test sessions (3.6 was chosen arbitrarily). We also want to continue to make sure the library works with the *latest* versions of the dependencies. When the constraints file is empty, the latest versions will be installed. If I'm reading the noxfile correctly it looks like unit tests run for [3.6, 3.7, and 3.8](https://github.com/googleapis/python-spanner/blob/87789c939990794bfd91f5300bedc449fd74bd7e/noxfile.py#L31). I added future python versions pre-emptively to reduce noise when the new versions come out. 

> Maybe a problem for another PR, but since the constraints files are just mechanically built from the requirements in setup.py, could synthtool keep the constraints files up to date instead of doing this manually?

That's a fair point, it would be nice if this were more automated. I've opened https://github.com/googleapis/google-cloud-python/issues/10521 to track that.

PR #205, opened 2021-01-07
feat: add support to log commit stats

larkee:
This PR adds support for logging commit statistics when requested by the user via `log_commit_stats`.

Commit statistics will be logged at the INFO log level.

The default logger will log commit stats using `sys.stderr`, however the user can pass their own logger when constructing the database if needed.

A sample of how to use the feature is given below:
```
database = instance.database(database_id, logger=custom_logger)
database.log_commit_stats = True
```

Once `log_commit_stats` is set, commit stats will be logged via the database logger any time commit is called by the library.

### DIFF
--- a/google/cloud/spanner_v1/database.py
+++ b/google/cloud/spanner_v1/database.py
@@ -656,7 +656,7 @@ class BatchCheckout(object):
             if exc_type is None:
                 self._batch.commit(return_commit_stats=self._database.log_commit_stats)
         finally:
-            if self._database.log_commit_stats:
+            if self._database.log_commit_stats and self._batch.commit_stats:
                 self._database.logger.info(
                     "CommitStats: {}".format(self._batch.commit_stats),
                     extra={"commit_stats": self._batch.commit_stats},
--- a/google/cloud/spanner_v1/database.py
+++ b/google/cloud/spanner_v1/database.py
@@ -237,12 +237,6 @@ class Database(object):
 
             ch = logging.StreamHandler()
             ch.setLevel(logging.INFO)
-
-            formatter = logging.Formatter(
-                "%(asctime)s - %(name)s - %(levelname)s - %(message)s"
-            )
-            ch.setFormatter(formatter)
-
             self._logger.addHandler(ch)
         return self._logger
 
@@ -657,11 +651,7 @@ class BatchCheckout(object):
                 self._batch.commit(return_commit_stats=self._database.log_commit_stats)
         finally:
             if self._database.log_commit_stats:
-                self._database.logger.info(
-                    "Transaction mutation count: {}".format(
-                        self._batch.commit_stats.mutation_count
-                    )
-                )
+                self._database.logger.info(self._batch.commit_stats)
             self._database._pool.put(self._session)
 
 
--- a/google/cloud/spanner_v1/database.py
+++ b/google/cloud/spanner_v1/database.py
@@ -657,7 +657,10 @@ class BatchCheckout(object):
                 self._batch.commit(return_commit_stats=self._database.log_commit_stats)
         finally:
             if self._database.log_commit_stats:
-                self._database.logger.info(self._batch.commit_stats)
+                self._database.logger.info(
+                    "CommitStats: {}".format(self._batch.commit_stats),
+                    extra={"commit_stats": self._batch.commit_stats},
+                )
             self._database._pool.put(self._session)
 
 
--- a/google/cloud/spanner_v1/database.py
+++ b/google/cloud/spanner_v1/database.py
@@ -96,6 +96,12 @@ class Database(object):
     :param pool: (Optional) session pool to be used by database.  If not
                  passed, the database will construct an instance of
                  :class:`~google.cloud.spanner_v1.pool.BurstyPool`.
+
+    :type logger: `logging.Logger`
+    :param logger: (Optional) a custom logger that is used if `log_commit_stats`
+                   is `True` to log commit statistics. If not passed, a logger
+                   will be created when needed that will log the commit statistics
+                   to stdout.
     """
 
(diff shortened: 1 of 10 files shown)


google/cloud/spanner_v1/transaction.py
```diff
@@ -132,14 +138,17 @@ def commit(self):
...
-            response = api.commit(
-                session=self._session.name,
-                mutations=self._mutations,
-                transaction_id=self._transaction_id,
-                metadata=metadata,
-            )
+            response = api.commit(request=request, metadata=metadata,)
         self.committed = response.commit_timestamp
+        if return_commit_stats:
+            self.commit_stats = response.commit_stats
```
skuruppu, line 151: Can a transaction object be used across different transactions? If so, is there a way to clear `commit_stats` when a new transaction starts?

google/cloud/spanner_v1/types/spanner.py
```diff
@@ -922,12 +930,36 @@ class CommitResponse(proto.Message):
...
+    class CommitStats(proto.Message):
+        r"""Additional statistics about a commit.
+
+        Attributes:
+            mutation_count (int):
+                The total number of the mutations for the
+                transaction.
+            overload_delay (~.duration.Duration):
+                Length of time the commit was delayed due to
+                overloaded servers.
```
skuruppu, line 948: I assume this is a generated file. If so it's fine, but if not, then remove the `overload_delay` references as we're not releasing that right now.

google/cloud/spanner_v1/transaction.py
larkee: No. Once a transaction is committed, it is no longer referenced by the library and will be GC'd once it is out of scope.

skuruppu (approved):

busunkim96 (approved):
Stamping for the samples noxfile update.

busunkim96 (approved):
The logging bits look good to me, but I haven't worked much with loggers. 🙈 

@Daniel-Sanche would you be able to review this? Are there any best practices to ensure client library logs work nicely with the Observability products?

google/cloud/spanner_v1/database.py
```diff
@@ -216,6 +221,31 @@ def ddl_statements(self):
...
+        """
+        if self._logger is None:
+            self._logger = logging.getLogger(self.name)
+            self._logger.setLevel(logging.INFO)
+
+            ch = logging.StreamHandler()
+            ch.setLevel(logging.INFO)
+
+            formatter = logging.Formatter(
+                "%(asctime)s - %(name)s - %(levelname)s - %(message)s"
```
Daniel-Sanche, line 242: I don't know all the context, but are you sure all these fields would be useful to all users? (Is it necessary to put the levelname? It seems like all logs would have level `INFO`)

I'd suggest using a less opionated default, and letting the user add formatting if they need it

Daniel-Sanche:
> Are there any best practices to ensure client library logs work nicely with the Observability products?

I think that really depends on where this code is run, since different GCP environments have different logging policies. Many GCP environments work best if the log is printed as [structured JSON](https://cloud.google.com/logging/docs/structured-logging), but that isn't supported everywhere. If we want to make sure the logs look good in Cloud Logging no matter where it's run, right now the best option is probably to include the logging client library here. 

But if we want to avoid the new dependency, I'd say the best option would be to print a simple message to stdout, without extra formatting.

larkee:
@Daniel-Sanche Thank you for the feedback! 

I've removed the formatter and passed the `CommitStats` object directly to the `info` calls. This means the default behavior will print the `CommitStats` object to std_out, e.g.
```
mutation_count: 2
```

Users can use a custom logger with a formatter to suit their logging needs, or a custom logger which overloads `info` method and has access to the `CommitStats` object, e.g.
```
class CustomLogger(logging.Logger):
    def __init__(self):
        self._last_commit_stats = None

    def info(self, commit_stats):
        self._last_commit_stats = commit_stats
        super().info(commit_stats)
        
    def get_last_commit_stats():
        return self._last_commit_stats
```

I think this is a much better design. WDYT?

Daniel-Sanche:
For the most part that seems good to me, I think it's a good idea to just write the stats to stdout, and let the user override formatting/handlers if they decide to.

I'm not sure if I understand the `CustomLogger` part though. It's not a good idea to re-interpret the input to `info` like that in a logger subclass.  It seems like you might want to pass in `commit_stats` in the [`extra` argument](https://docs.python.org/3/library/logging.html#logging.Logger.debug). That would allow users to reference it in format strings, or access it in custom Handler classes

google/cloud/spanner_v1/session.py
```diff
@@ -358,11 +358,7 @@ def run_in_transaction(self, func, *args, **kw):
                 raise
             else:
                 if self._database.log_commit_stats:
-                    self._database.logger.info(
-                        "Transaction mutation count: {}".format(
-                            txn.commit_stats.mutation_count
-                        )
-                    )
+                    self._database.logger.info(txn.commit_stats)
```
Daniel-Sanche, line 361: I wnet over this in my other comment, but I'll repeat it here too: The docs for Python Logging say they expect the main `msg` argument to be a string. If you want to add a custom object, see the `extra` argument

larkee: Ah, that's exactly what I want. I've updated the call to use a formatted message _and_ pass the `CommitStats` via the extra argument. Thank you for your patience while I learn how to use loggers correctly :+1:

Daniel-Sanche (approved):

google/cloud/spanner_v1/session.py
Daniel-Sanche: No problem! Yeah this now LGTM from a logging perspective

PR #217, opened 2021-01-29
test: make system test timeout configurable and default to 60 seconds

arithmetic1728:
b/173067462

It seems 30 seconds are too short for mtls test (which uses the system test, and runs on an internal platform). This makes the mtls test very flaky. This PR introduces a `SPANNER_OPERATION_TIMEOUT_IN_SECONDS` env var to make the timeout configurable. 

The default value is now 60 seconds.

### DIFF
--- a/tests/system/test_system.py
+++ b/tests/system/test_system.py
@@ -55,6 +55,7 @@ from tests._helpers import OpenTelemetryBase, HAS_OPENTELEMETRY_INSTALLED
 CREATE_INSTANCE = os.getenv("GOOGLE_CLOUD_TESTS_CREATE_SPANNER_INSTANCE") is not None
 USE_EMULATOR = os.getenv("SPANNER_EMULATOR_HOST") is not None
 SKIP_BACKUP_TESTS = os.getenv("SKIP_BACKUP_TESTS") is not None
+SPANNER_RPC_TIMEOUT_IN_SECONDS = int(os.getenv("SPANNER_RPC_TIMEOUT_IN_SECONDS", 30))
 
 if CREATE_INSTANCE:
     INSTANCE_ID = "google-cloud" + unique_resource_id("-")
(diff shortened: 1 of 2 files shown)


tests/system/test_system.py
```diff
@@ -55,6 +55,7 @@
 CREATE_INSTANCE = os.getenv("GOOGLE_CLOUD_TESTS_CREATE_SPANNER_INSTANCE") is not None
 USE_EMULATOR = os.getenv("SPANNER_EMULATOR_HOST") is not None
 SKIP_BACKUP_TESTS = os.getenv("SKIP_BACKUP_TESTS") is not None
+SPANNER_RPC_TIMEOUT_IN_SECONDS = int(os.getenv("SPANNER_RPC_TIMEOUT_IN_SECONDS", 30))
```
larkee, line 58: This timeout is not an RPC timeout. It defines how long to poll for a result on the operation object. I would rename this `SPANNER_OPERATION_TIMEOUT_IN_SECONDS` to indicate this.

larkee (commented):
LGTM with a minor change to the flag name 👍 

We have actually also encountered this flake on the repo but I haven't had time to investigate. In light of this, I would maybe suggest doubling the default value to `60`.

Thank you very much for making this PR!

tests/system/test_system.py
arithmetic1728: Now renamed the env var to `SPANNER_OPERATION_TIMEOUT_IN_SECONDS`, and default it to 60 seconds.

arithmetic1728:
> LGTM with a minor change to the flag name 👍
> 
> We have actually also encountered this flake on the repo but I haven't had time to investigate. In light of this, I would maybe suggest doubling the default value to `60`.
> 
> Thank you very much for making this PR!

I changed the default value to 60 seconds, and renamed the env var. Please take another look. Thank you!

tests/system/test_system_dbapi.py
```diff
@@ -39,6 +39,9 @@
 )
 
 
+SPANNER_RPC_TIMEOUT_IN_SECONDS = int(os.getenv("SPANNER_RPC_TIMEOUT_IN_SECONDS", 30))
```
larkee, line 42: Please update this to be 60 seconds as well 👍

larkee (approved):
LGTM. Please update the PR description to mention the default is now 60 seconds.

tests/system/test_system_dbapi.py
arithmetic1728: Updated the name and value. Thanks!

PR #219, opened 2021-01-29
feat: add `Database.list_tables` method

### LINKED ISSUE #156: FR: Adding additional functions in the API library
**Is your feature request related to a problem? Please describe.**

The current Google.cloud.spanner library has very limited functionalities compared to other db libraries (e.g., bigquery) and some of the key APIs are missing. For example,

get_table
table
list_tables
schema
query
execute

I would like to have the above functions added to the library.

 **Describe the solution you'd like**

These functions could be implementable via the INFORMATION_SCHEMA query syntax.

tswast:
Adds Table-management methods required by PSO data validation tool.

Fixes #156 🦕

### DIFF
--- a/docs/table-usage.rst
+++ b/docs/table-usage.rst
@@ -19,8 +19,8 @@ To iterate over all existing tables for an database, use its
 This method yields :class:`~google.cloud.spanner_v1.table.Table` objects.
 
 
-Constructing a Table
---------------------
+Table Factory
+-------------
 
 A table object can be created by using the
 :class:`~google.cloud.spanner_v1.table.Table` constructor. Since table
@@ -36,11 +36,11 @@ operations are executed via SQL, a
 Getting the Table Schema
 ------------------------
 
-Use the :meth:`~google.cloud.spanner_v1.table.Table.get_schema` method to
-inspect the columns of a table as a list of
+Use the :attr:`~google.cloud.spanner_v1.table.Table.schema` property to inspect
+the columns of a table as a list of
 :class:`~google.cloud.spanner_v1.types.StructType.Field` objects.
 
 .. code:: python
 
-    for field in table.get_schema():
+    for field in table.schema
         # `field` is a `Field` object.
--- a/docs/table-usage.rst
+++ b/docs/table-usage.rst
@@ -22,16 +22,16 @@ This method yields :class:`~google.cloud.spanner_v1.table.Table` objects.
 Table Factory
 -------------
 
-A table object can be created by using the
-:class:`~google.cloud.spanner_v1.table.Table` constructor. Since table
-operations are executed via SQL, a
-:class:`~google.cloud.spanner_v1.database.Database` instance is required.
+A :class:`~google.cloud.spanner_v1.table.Table` object can be created with the
+:meth:`~google.cloud.spanner_v1.database.Database.table` factory method:
 
 .. code:: python
 
-    table = google.cloud.spanner_v1.table.Table(
-       "my_table_id", database
-    )
+    table = database.table("my_table_id")
+    if my_table.exists():
+        print("Table with ID 'my_table' exists.")
+    else:
+        print("Table with ID 'my_table' does not exist."
 
 Getting the Table Schema
 ------------------------
--- a/docs/table-usage.rst
+++ b/docs/table-usage.rst
@@ -38,7 +38,7 @@ Getting the Table Schema
 
 Use the :meth:`~google.cloud.spanner_v1.table.Table.get_schema` method to
 inspect the columns of a table as a list of
-:class:`~google.cloud.spanner_v1.types.Field` objects.
+:class:`~google.cloud.spanner_v1.types.StructType.Field` objects.
 
 .. code:: python
 
--- a/docs/table-usage.rst
+++ b/docs/table-usage.rst
@@ -28,7 +28,7 @@ A :class:`~google.cloud.spanner_v1.table.Table` object can be created with the
 .. code:: python
 
     table = database.table("my_table_id")
-    if my_table.exists():
+    if table.exists():
         print("Table with ID 'my_table' exists.")
     else:
         print("Table with ID 'my_table' does not exist."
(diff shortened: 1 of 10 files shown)


google/cloud/spanner_v1/table.py
```diff
@@ -0,0 +1,36 @@
...
+
+class Table(object):
+    """Representation of a Cloud Spanner Table.
+    """
+
+    def __init__(self, table_id, database):
+        self.table_id = table_id
+        self._database = database
+
+    def get_schema(self):
```
larkee, line 28: I think instead of having `get_schema`, it would make more sense to match the rest of the library and the BigQuery implementation by having this as a property. What do you think?

larkee (commented):
LG, just a couple design details to work out regarding the Table methods :+1:

google/cloud/spanner_v1/table.py
tswast: That would be more consistent. My main worry with a property is it might be less clear that `table.schema` makes an API call compared to `table.get_schema()`.

tswast:
I'm marking this as ready for review, though I suspect I'll need to add some unit tests for coverage.

google/cloud/spanner_v1/table.py
larkee: That is a good point. For the other resource objects (i.e. `Instance`, `Backup`, `Database`) we store local copies of properties and have a `reload()` method to make an RPC call to update them. I think it may make sense to stick to this design, especially if we decide to support create, update, and delete on the `Table` object.

I would initialize a private schema attribute as `None` and then call `reload` in `schema` if it is `None`. This means the property would make one API call at most. Then, if the user is expecting the schema to change frequently, they can all `reload` themselves to ensure the schema remains in sync with production. WDYT?

larkee:
The new Table type will also need documentation similar to [`database-api`](https://github.com/googleapis/python-spanner/blob/master/docs/database-api.rst) and [`database-usage`](https://github.com/googleapis/python-spanner/blob/master/docs/database-usage.rst] so that documentation for the type and how to use it can be generated.

tswast:
Docs added:

image

image

google/cloud/spanner_v1/table.py
tswast: > I would initialize a private schema attribute as None and then call reload in schema if it is None. This means the property would make one API call at most. Then, if the user is expecting the schema to change frequently, they can all reload themselves to ensure the schema remains in sync with production. WDYT?

I think that's a good approach. In BigQuery we didn't have this logic, so customers would surprisingly get `None` values, which caused all sorts of issues. I'll make this a property and add a reload method.

tswast: Updated. Also added a `exists()` method to match [Database.reload()](https://googleapis.dev/python/spanner/latest/database-api.html#google.cloud.spanner_v1.database.Database.reload)'s `NotFound` exception behavior.

docs/table-usage.rst
```diff
@@ -0,0 +1,46 @@
...
+-------------
+
+A table object can be created by using the
+:class:`~google.cloud.spanner_v1.table.Table` constructor. Since table
+operations are executed via SQL, a
+:class:`~google.cloud.spanner_v1.database.Database` instance is required.
+
+.. code:: python
+
+    table = google.cloud.spanner_v1.table.Table(
```
larkee, line 32: Please add a `table()` method on the `Database` object which creates the `Table` object to be consistent with the other classes e.g.
```
client = spanner.Client()
instance = client.instance(instance_id)
table = database.table(table_id)
```

tswast: Done! Thanks for your patience while I was OOO.

tswast:
@larkee I've made the requested changes and synced with latest `master` branch. Should be ready for another round of review.

tswast:
Test failure

```
______________________ TestBackupAPI.test_backup_workflow ______________________

self = <tests.system.test_system.TestBackupAPI testMethod=test_backup_workflow>

    def test_backup_workflow(self):
        from datetime import datetime
        from datetime import timedelta
        from pytz import UTC

        instance = Config.INSTANCE
        backup_id = "backup_id" + unique_resource_id("_")
        expire_time = datetime.utcnow() + timedelta(days=3)
        expire_time = expire_time.replace(tzinfo=UTC)

        # Create backup.
        backup = instance.backup(
            backup_id,
            database=self._db,
            expire_time=expire_time,
            version_time=self.database_version_time,
        )
        operation = backup.create()
        self.to_delete.append(backup)

        # Check metadata.
        metadata = operation.metadata
        self.assertEqual(backup.name, metadata.name)
        self.assertEqual(self._db.name, metadata.database)
        operation.result()

        # Check backup object.
        backup.reload()
        self.assertEqual(self._db.name, backup._database)
        self.assertEqual(expire_time, backup.expire_time)
        self.assertIsNotNone(backup.create_time)
        self.assertEqual(self.database_version_time, backup.version_time)
        self.assertIsNotNone(backup.size_bytes)
        self.assertIsNotNone(backup.state)

        # Update with valid argument.
        valid_expire_time = datetime.utcnow() + timedelta(days=7)
        valid_expire_time = valid_expire_time.replace(tzinfo=UTC)
        backup.update_expire_time(valid_expire_time)
        self.assertEqual(valid_expire_time, backup.expire_time)

        # Restore database to same instance.
        restored_id = "restored_db" + unique_resource_id("_")
        database = instance.database(restored_id)
        self.to_drop.append(database)
        operation = database.restore(source=backup)
        restored_db = operation.result()
>       self.assertEqual(
            self.database_version_time, restored_db.restore_info.backup_info.create_time
        )
E       AssertionError: datetime.datetime(2021, 2, 22, 21, 31, 33[16 chars]UTC>) != DatetimeWithNanoseconds(2021, 2, 22, 21, [40 chars].utc)
```

appears to be a problem with `master` branch. I'll try merging from upstream again.

larkee (approved):
LGTM other than a minor docs fix :+1:

PR #233, opened 2021-02-11
fix(db_api): update parse_util output for homogeneous statements

### LINKED ISSUE #232: db_api:  parse_insert method doesn't work as expected
As mentioned in the docstring `Case c` return the dict with the info like `homogenous`, `homogenous` etc:
https://github.com/googleapis/python-spanner/blob/1f80a395ff4ac51a785ea6bdfbe3e18a8cf4c951/google/cloud/spanner_dbapi/parse_utils.py#L235-L244

Where code of `Case c` returns sql_params_list:
https://github.com/googleapis/python-spanner/blob/1f80a395ff4ac51a785ea6bdfbe3e18a8cf4c951/google/cloud/spanner_dbapi/parse_utils.py#L305-L321

So it never execute for the homogeneous insert
https://github.com/googleapis/python-spanner/blob/1f80a395ff4ac51a785ea6bdfbe3e18a8cf4c951/google/cloud/spanner_dbapi/connection.py#L306

cc:// @larkee @c24t

HemangChothani:
Fixes #232

### DIFF
--- a/google/cloud/spanner_dbapi/parse_utils.py
+++ b/google/cloud/spanner_dbapi/parse_utils.py
@@ -306,19 +306,15 @@ def parse_insert(insert_sql, params):
         # Case c)
 
         columns = [mi.strip(" `") for mi in match.group("columns").split(",")]
-        sql_params_list = []
-        insert_sql_preamble = "INSERT INTO %s (%s) VALUES %s" % (
-            match.group("table_name"),
-            match.group("columns"),
-            values.argv[0],
-        )
         values_pyformat = [str(arg) for arg in values.argv]
         rows_list = rows_for_insert_or_update(columns, params, values_pyformat)
-        insert_sql_preamble = sanitize_literals_for_upload(insert_sql_preamble)
-        for row in rows_list:
-            sql_params_list.append((insert_sql_preamble, row))
 
-        return {"sql_params_list": sql_params_list}
+        return {
+            "homogenous": True,
+            "table": match.group("table_name"),
+            "columns": columns,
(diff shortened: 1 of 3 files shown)


google/cloud/spanner_dbapi/parse_utils.py
```diff
@@ -306,19 +306,15 @@ def parse_insert(insert_sql, params):
...
-        for row in rows_list:
-            sql_params_list.append((insert_sql_preamble, row))
 
-        return {"sql_params_list": sql_params_list}
+        return {
+            "homogenous": True,
+            "table": match.group("table_name"),
+            "columns": columns,
+            "values": rows_list,
+        }
```
larkee, line 317: Is this a breaking change?

HemangChothani: Nope i think.

larkee:
Please update the title to be more descriptive i.e. `fix: update parse_util output for homogeneous statements`

The PR title will be used in the release notes.

google/cloud/spanner_dbapi/parse_utils.py
larkee: Oh, I see. The docstring correctly documents this behavior. In that case, I don't think this is a breaking change since it is fixing it to match the documented behavior.

larkee (commented):
LGTM once the title is updated

HemangChothani:
> LGTM once the title is updated

I have changed the title.

larkee (approved):

PR #240, opened 2021-02-22
perf: improve streaming performance

### LINKED ISSUE #207: Slow _merge_values ( parsing values from protobuf to string)
https://github.com/googleapis/python-spanner/blob/4a58bcb4016977a8b2ae7fe886c4db40a13e85cd/google/cloud/spanner_v1/streamed.py#L103

We have a query returning 80.000 rows with 71 fields on the select list using Python3, Google Cloud Spanner API 1.17 (we tried 1.18, 1.19, and 2.x). I choose version 1.17 because the performance decreases with newer versions of API.

The query returns the data up to 0.9ms when I start to copy the rows from StreamedResultset iterator to a list.

I started to isolate the code, and I'm using an empty "for loop" to simulate the problem and discard any other application performance problem.

```
Code Snippet:

import threading
import time
import os
from google.cloud.spanner import Client, PingingPool
from google.cloud.spanner_v1 import instance, database

"""Queries sample data from the database using SQL."""

projectId = 'cerc2-datalake-int-01'
instanceId = 'datalake-int-spanner-01'
database_id = 'cerc_datalake_bk_int14'

spanner_client = Client(projectId)

instance = spanner_client.instance(instanceId)

# Read query from query.sql file.
file = open('query.sql',mode='r')
content  = file.read()
file.close()

# Executing the query
start = time.time()
print('Starting query')
    results = snapshot.execute_sql(
        content 
    )
end = time.time()
print(f'Query execution time: {end - start}')

# Count returned rows ( USING VERSION 2.x:  Error in this piece of code, some timestamp conversion error into google-core-api - removed all datetime fields to test)
start = time.time()
i = 0    
for row in results:
    i+=1
end = time.time()
print(f'Iterate itens {end - start}')


```
Using cProfile libraries to profile the application, I realized that the call of method _parse_value_pb inside _merge_values into streamed.py is the slowest method in my application. 

For testing purposes, I removed the streamed.py file line 106:
>> self._current_row.append(_parse_value_pb(value, field.type))

The line has been replaced by: 
>> self._current_row.append(value, field.type)

The first scenario (with _parse_value_pb):
Query 0.9ms, resultset iteration: 25 seconds;

The second scenario, removing _parse_value_pb:
Query 0.9ms, resultset iteration: 6 seconds;

All these tests is running on my 2.4Ghz laptop, but when we use Appengine, this routine gets more than 120 seconds;

I've tested with google spanner library 2.1 version, I got 103 seconds instead of 25 seconds ( version 1.17), with the same behavior;

The performance is fair when the resultset has many rows and a few columns in each row. I tested up to 30 resultset columns, and this behavior is not a problem. In my case, I need to work with 71 columns in each row.

### LINKED ISSUE #234: Bug: StreamedResultSet double-encodes merged byte chunks after querying, throwing error

larkee:
This PR fixes performance issues that were introduced during the mirgration to v2 and optimizes several streaming methods. 

From v2 onwards, the library uses `proto-plus` to produce user friendly Python object to represent proto messages. Unfortunately, it significantly impacted the performance of streaming for cases with many columns as the library must unnecessarily parse the types many times. By reverting the streaming changes to use the raw protobuf messages, the streaming speed is greatly improved.

Fixes #207, fixes #234

### DIFF
--- a/google/cloud/spanner_v1/streamed.py
+++ b/google/cloud/spanner_v1/streamed.py
@@ -14,12 +14,15 @@
 
 """Wrapper for streaming results."""
 
+from google.protobuf.struct_pb2 import ListValue
+from google.protobuf.struct_pb2 import Value
 from google.cloud import exceptions
+from google.cloud.spanner_v1 import PartialResultSet
 from google.cloud.spanner_v1 import TypeCode
 import six
 
 # pylint: disable=ungrouped-imports
-from google.cloud.spanner_v1._helpers import _parse_value
+from google.cloud.spanner_v1._helpers import _parse_value_pb
 
 # pylint: enable=ungrouped-imports
 
@@ -88,7 +91,7 @@ class StreamedResultSet(object):
         field = self.fields[current_column]
         merged = _merge_by_type(self._pending_chunk, value, field.type_)
         self._pending_chunk = None
-        return _parse_value(merged, field.type_)
+        return merged
 
     def _merge_values(self, values):
         """Merge values into rows.
@@ -96,14 +99,17 @@ class StreamedResultSet(object):
         :type values: list of :class:`~google.protobuf.struct_pb2.Value`
         :param values: non-chunked values from partial result set.
         """
-        width = len(self.fields)
+        print(self.fields)
+        field_types = [field.type_ for field in self.fields]
+        width = len(field_types)
+        index = len(self._current_row)
         for value in values:
-            index = len(self._current_row)
-            field = self.fields[index]
-            self._current_row.append(_parse_value(value, field.type_))
-            if len(self._current_row) == width:
+            self._current_row.append(_parse_value_pb(value, field_types[index]))
+            index += 1
+            if index == width:
                 self._rows.append(self._current_row)
                 self._current_row = []
+                index = 0
 
     def _consume_next(self):
         """Consume the next partial result set from the stream.
@@ -111,6 +117,7 @@ class StreamedResultSet(object):
         Parse the result set into new/existing rows in :attr:`_rows`
         """
         response = six.next(self._response_iterator)
+        response_pb = PartialResultSet.pb(response)
 
         if self._metadata is None:  # first response
             metadata = self._metadata = response.metadata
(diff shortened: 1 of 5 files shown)


skuruppu (approved):

google/cloud/spanner_v1/streamed.py
```diff
@@ -88,29 +91,33 @@ def _merge_chunk(self, value):
...
+        return merged
 
     def _merge_values(self, values):
         """Merge values into rows.
 
         :type values: list of :class:`~google.protobuf.struct_pb2.Value`
         :param values: non-chunked values from partial result set.
         """
-        width = len(self.fields)
+        print(self.fields)
```
harmon, line 102: Did this make it into the final code?

PR #247, opened 2021-03-01
samples: PITR samples backup fix

thiagotnunes:
Instead of using the earliest version time of the database, uses the current time (from spanner). If we used the earliest version time of the database instead we would be creating an empty backup, since the database we are backing up is not 1 hour old (default version retention period).

### DIFF
--- a/samples/samples/backup_sample.py
+++ b/samples/samples/backup_sample.py
@@ -32,26 +32,29 @@ def create_backup(instance_id, database_id, backup_id):
     instance = spanner_client.instance(instance_id)
     database = instance.database(database_id)
 
-    # Create a backup
-    expire_time = datetime.utcnow() + timedelta(days=14)
-    version_time = database.earliest_version_time
-    backup = instance.backup(backup_id, database=database, expire_time=expire_time, version_time=version_time)
-    operation = backup.create()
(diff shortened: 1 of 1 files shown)


samples/samples/backup_sample.py
```diff
@@ -32,26 +32,29 @@ def create_backup(instance_id, database_id, backup_id):
...
-    backup = instance.backup(backup_id, database=database, expire_time=expire_time, version_time=version_time)
-    operation = backup.create()
-
-    # Wait for backup operation to complete.
-    operation.result(1200)
-
-    # Verify that the backup is ready.
-    backup.reload()
-    assert backup.is_ready() is True
+    with database.snapshot() as snapshot:
```
larkee, line 35: Rather than moving the whole test under the with statement, I think it would be better to only use the snapshot to get the version time. e.g.
```
version_time = None
    results = snapshot.execute_sql("SELECT CURRENT_TIMESTAMP()")
    version_time = list(results)[0][0]
```
This way it is clear that a snapshot is needed to get the current timestamp and it isn't needed when creating a backup.

thiagotnunes: Done.

larkee (approved):

PR #251, opened 2021-03-02
feat: add support for custom timeout and retry parameters in execute_update method in transactions

### LINKED ISSUE #71: Add support for custom timeout and retry parameters through object surfaces
As noted in #67, for most methods there is no way to set timeout and retry parameters without using the GAPIC client directly. 

Adding support for custom timeout and retries will give users more flexibility if they require it.

vi3k6i5:
feat: Added support for custom timeout and retry parameters in transactions.

Fixes #71 🦕

### DIFF
--- a/google/cloud/spanner_v1/transaction.py
+++ b/google/cloud/spanner_v1/transaction.py
@@ -11,7 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 """Spanner read-write transaction support."""
 
 from google.protobuf.struct_pb2 import Struct
@@ -192,6 +191,7 @@ class Transaction(_SnapshotBase, _BatchBase):
         param_types=None,
         query_mode=None,
         query_options=None,
+        *,
         retry=gapic_v1.method.DEFAULT,
         timeout=None,
     ):
@@ -221,7 +221,7 @@ class Transaction(_SnapshotBase, _BatchBase):
         :param query_options: (Optional) Options that are provided for query plan stability.
 
         :type retry: :class:`~google.api_core.retry.Retry`
-        :param retry: (Optional) Designation of what errors, if any, should be retried.
+        :param retry: (Optional) The retry settings for this request.
 
         :type timeout: float
         :param timeout: (Optional) The timeout for this request.
--- a/google/cloud/spanner_v1/transaction.py
+++ b/google/cloud/spanner_v1/transaction.py
@@ -194,7 +194,7 @@ class Transaction(_SnapshotBase, _BatchBase):
         query_options=None,
         *,
         retry=gapic_v1.method.DEFAULT,
-        timeout=None,
+        timeout=gapic_v1.method.DEFAULT,
     ):
         """Perform an ``ExecuteSql`` API request with DML.
 
--- a/google/cloud/spanner_v1/transaction.py
+++ b/google/cloud/spanner_v1/transaction.py
@@ -29,6 +29,8 @@ from google.cloud.spanner_v1 import TransactionOptions
 from google.cloud.spanner_v1.snapshot import _SnapshotBase
 from google.cloud.spanner_v1.batch import _BatchBase
 from google.cloud.spanner_v1._opentelemetry_tracing import trace_call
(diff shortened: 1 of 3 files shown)


olavloite (commented):
This change in itself seems reasonable to me, but I don't know the exact background for this change. From the description of the PR it seems that the goal is to add custom timeout and retry parameters for transactions. This PR only introduces timeout and retry parameters for single update statements in a transaction. It does not:
1. Add the possibility to specify a timeout for `Begin`, `Commit` or `ExecuteBatchDML` statements in a transaction.
2. Add the possibility to set a timeout for the entire transaction (which I would also expect not to be something we want, but I'm a little confused by the PR title)

Is this intentional?

vi3k6i5:
Thanks @olavloite . I Updated the PR title, that should clarify it. We were planning to add this parameter 1 by 1 in other places as well. @larkee : Shall I update the same PR with other implementations as well?

skuruppu:
I think this is fine since we may want to see how this looks for one method first. Then in the next PR, we can do multiples of these.

olavloite (approved):

thiagotnunes (commented):
This looks good to me, but I have a few questions:

1. Are we going to add custom timeouts for other methods as well? 
2. Are we also planning on making it possible to override the default retry strategy of the methods?
3. Do we want to provide a way so that a timeout is set during initialization and the caller does not have to specify the timeout on every call (or do we already have this functionality)?

thiagotnunes:
Sorry, I missed the above comments, so I think it looks good, since this is the first step of the feature.

thiagotnunes (approved):

thiagotnunes:
Please wait for @larkee 's review before merging, as they have more context on the library implementation.

google/cloud/spanner_v1/transaction.py
```diff
@@ -185,7 +186,14 @@ def _make_params_pb(params, param_types):
...
 
     def execute_update(
-        self, dml, params=None, param_types=None, query_mode=None, query_options=None
+        self,
+        dml,
+        params=None,
+        param_types=None,
+        query_mode=None,
+        query_options=None,
+        retry=gapic_v1.method.DEFAULT,
```
larkee, line 195: I think having `retry` and `timeout` as [keyword-only arguments](https://www.python.org/dev/peps/pep-3102/) is tidier. If more arguments relating to the contents of the request are added later, it will be possible to group them with the other content arguments without a breaking change, e.g.:
```
    def execute_update(
        self,
        dml,
        params=None,
        param_types=None,
        query_mode=None,
        query_options=None,
        additional_feature=None
        *,
        retry=gapic_v1.method.DEFAULT,
        timeout=None,
    ):
```
WDYT?

google/cloud/spanner_v1/transaction.py
```diff
@@ -212,6 +220,12 @@ def execute_update(
             or :class:`dict`
         :param query_options: (Optional) Options that are provided for query plan stability.
 
+        :type retry: :class:`~google.api_core.retry.Retry`
+        :param retry: (Optional) Designation of what errors, if any, should be retried.
```
larkee, line 224: The `Retry` object also specifies the back-off retry behavior. Perhaps: `The retry settings for this request.` or something similar.

larkee (changes requested):

google/cloud/spanner_v1/transaction.py
vi3k6i5: Makes sense, updated the comment.

larkee (approved):

vi3k6i5 (commented):
Build check failed, retrying.

google/cloud/spanner_v1/transaction.py
```diff
@@ -185,7 +186,15 @@ def _make_params_pb(params, param_types):
...
-        self, dml, params=None, param_types=None, query_mode=None, query_options=None
+        self,
+        dml,
+        params=None,
+        param_types=None,
+        query_mode=None,
+        query_options=None,
+        *,
+        retry=gapic_v1.method.DEFAULT,
+        timeout=None,
```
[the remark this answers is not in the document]
vi3k6i5: changed to gapic_v1.method.DEFAULT in test as well. Thanks for pointing out out, I thought since timeout value has to be a float so we are not setting default as object.

PR #276, opened 2021-03-16
feat: adding support for spanner request options tags

### LINKED ISSUE #253: Add tagging support on database operations such as reads, queries and transactions.
Description:
Java client library is adding support for tagging https://github.com/googleapis/java-spanner/pull/576
Similarly python client can also add support to enable users to set tags on database operations such as reads, queries and transactions.

Background
The statistics of queries, reads and transactions in Cloud Spanner are stored. Providing tags for database operations will allow these statistics to be grouped by a tag and makes queries/transactions easily searchable by tag. This will help make the information provided by the statistics more useful.

vi3k6i5:
Adding request options to relevant methods and adding unit tests for the same.
Fixes #253 🦕


vi3k6i5:
Added sleep time of 200 seconds before looking up the tags in db. Just for validation it's okay for now, this will have to be fixed before we merge with mainline.

tests/system/test_system.py
```diff
@@ -51,10 +51,12 @@
 from tests._fixtures import DDL_STATEMENTS
 from tests._fixtures import EMULATOR_DDL_STATEMENTS
 from tests._helpers import OpenTelemetryBase, HAS_OPENTELEMETRY_INSTALLED
+from google.cloud.spanner_v1 import RequestOptions
 
 
 CREATE_INSTANCE = os.getenv("GOOGLE_CLOUD_TESTS_CREATE_SPANNER_INSTANCE") is not None
 USE_EMULATOR = os.getenv("SPANNER_EMULATOR_HOST") is not None
+STAGING_API_ENDPOINT = os.getenv("SPANNER_STAGING_HOST_API_ENDPOINT") is not None
```
thiagotnunes, line 59: nit: this could be any endpoint potentially right? Something like `SPANNER_API_ENDPOINT` might be more general.

tests/system/test_system.py
```diff
@@ -552,6 +559,44 @@ def _unit_of_work(transaction, test):
...
+
+            transaction.insert_or_update(test.TABLE, test.COLUMNS, test.ROW_DATA)
+
+        self._db.run_in_transaction(_unit_of_work, test=self)
+
+        request_tag = "req-tag" + unique_resource_id("-")
+        with self._db.snapshot() as before:
+            rows = list(
+                before.execute_sql(
+                    self.SQL, request_options={"request_tag": request_tag}
```
thiagotnunes, line 584: Out of curiosity, do we do any validation of the `request_options` given? If I pass in `request_options={"invalid": 10}` would that result in an error?

tests/system/test_system.py
```diff
@@ -552,6 +559,44 @@ def _unit_of_work(transaction, test):
...
+                )
+            )
+        self._check_rows_data(rows)
+        time.sleep(200)
```
thiagotnunes, line 588: nit: could you add a fix-me / todo to rewrite this, since it will potentially delay the test?

One thing we could do is test if the request tag is there every `10` seconds and timeout after `200` seconds. This way the test can finish sooner.

thiagotnunes (approved):

larkee (commented):
Added some comments with more details on transactions.

Additionally, we discussed adding validation but had difficulties due to class inheritance. I think if we add a `_read_only` attribute to `_SnapshotBase` and `Transaction` which is set to `True` and `False` respectively, similar to `_multi_use`, then we should be able to add validation for the `request_options`.

tests/system/test_system.py
vi3k6i5: Thanks will use that instead.

tests/system/test_system.py
vi3k6i5: for now added a TODO.

google/cloud/spanner_v1/snapshot.py
```diff
@@ -145,6 +146,11 @@ def read(
                           from :meth:`partition_read`.  Incompatible with
                           ``limit``.
 
+        :type request_options:
+            :class:`~google.cloud.spanner_v1.ExecuteSqlRequest.RequestOptions`
+            or :class:`dict`
+        :param request_options: (Optional) Tags that are provided for request.
```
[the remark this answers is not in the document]
vi3k6i5: This comment and param was redundant so removed it.

google/cloud/spanner_v1/batch.py
```diff
@@ -138,13 +137,18 @@ def _check_state(self):
         if self.committed is not None:
             raise ValueError("Batch already committed")
 
-    def commit(self, return_commit_stats=False):
+    def commit(self, return_commit_stats=False, request_options=None):
```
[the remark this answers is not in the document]
vi3k6i5: okay, removed the parameter from here.

google/cloud/spanner_v1/transaction.py
```diff
@@ -122,13 +122,18 @@ def rollback(self):
         self.rolled_back = True
         del self._session._transaction
 
-    def commit(self, return_commit_stats=False):
+    def commit(self, return_commit_stats=False, request_options=None):
```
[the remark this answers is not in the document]
vi3k6i5: `request_options=RequestOptions(transaction_tag=self._transaction_tag),` won't that make it complicated to set both transaction_tag and request_tag also `Priority` is there as well.

tests/unit/test_transaction.py
```diff
@@ -314,7 +314,9 @@ def test_commit_w_other_error(self):
             attributes=dict(TestTransaction.BASE_ATTRIBUTES, num_mutations=1),
         )
 
-    def _commit_helper(self, mutate=True, return_commit_stats=False):
+    def _commit_helper(
+        self, mutate=True, return_commit_stats=False, request_options=None
```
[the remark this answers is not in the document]
vi3k6i5: Made this change in tests, but I feel there are multiple edge cases that needs to be tested with the changes and I feel like I don't have enough context to add test for all of them.

skuruppu:
@thiagotnunes I ran out of time to review this before I go on vacation. Would you please be able to review and approve on Monday?

larkee (approved):

PR #277, opened 2021-03-16
feat(db_api): support executing several DDLs separated by semicolon

IlyaFaer:
With this change users will be able to put several DDLs into one `execute()` call, like:

```python
cursor.execute("CREATE TABLE ...;CREATE INDEX...")
```

### DIFF
--- a/tests/unit/spanner_dbapi/test_cursor.py
+++ b/tests/unit/spanner_dbapi/test_cursor.py
@@ -862,3 +862,31 @@ class TestCursor(unittest.TestCase):
                     cursor.fetchmany(len(row))
 
                 run_mock.assert_called_with(statement, retried=True)
+
+    def test_ddls_with_semicolon(self):
+        """
+        Check that one script with several DDL statements separated
+        with semicolons is splitted into several DDLs.
+        """
+        from google.cloud.spanner_dbapi.connection import connect
+
+        EXP_DDLS = [
+            "CREATE TABLE table_name (row_id INT64) PRIMARY KEY ()",
+            "DROP TABLE table_name",
+        ]
+
+        with mock.patch(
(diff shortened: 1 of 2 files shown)


IlyaFaer:
@larkee, these are changes needed for SQLAlchemy dialect, first of all to make it possible to drop indexes and constraints before dropping the table.

larkee:
Please add or update a unit test with an example of a DDL statement with semicolons.

tests/unit/spanner_dbapi/test_cursor.py
```diff
@@ -862,3 +862,31 @@ def test_fetchmany_retry_aborted_statements_checksums_mismatch(self):
...
+        ):
+            with mock.patch(
+                "google.cloud.spanner_v1.database.Database.exists", return_value=True,
+            ):
+                connection = connect("test-instance", "test-database")
+
+        cursor = connection.cursor()
+        cursor.execute(
+            "CREATE TABLE table_name (row_id INT64) PRIMARY KEY ();"
+            "DROP TABLE table_name;"
```
IlyaFaer, line 889: Set semicolons in both lines to check that no empty DDLs are executed (as `split()` can produce empty lines, if the separator is at the end of the line)

larkee:
Would it be reasonable to also allow for each statement to start on a new line? See [here](https://github.com/googleapis/python-spanner/blob/master/tests/_fixtures.py) for an example of what I mean.

IlyaFaer:
@larkee, it's already working with new lines. I've added a line with `\n` symbol into the test to be sure.

`str.strip()` erases `\n` symbols as well (though it's something not mentioned in docs directly). Here is a short example:

image

larkee (approved):

PR #285, opened 2021-03-22
fix(db_api): revert Mutations API usage

IlyaFaer:
This PR reverts changes made in #233.

I've executed SQLAlchemy tests several times - looks like with #233 changes reverted everything is working fine now. I've mentioned that DLL `test_create` is failing when 223 is reverted, but it's no more actual, so we can safely revert it I assume - I only have those 3 tests mentioned in [the comment](https://github.com/cloudspannerecosystem/python-spanner-sqlalchemy/pull/33#issuecomment-802853769) failing, which is expected for now.

### DIFF
--- a/google/cloud/spanner_dbapi/parse_utils.py
+++ b/google/cloud/spanner_dbapi/parse_utils.py
@@ -237,10 +237,10 @@ def parse_insert(insert_sql, params):
             Params: ['a', 'b', 'c', 'd']
         it produces:
             {
-                'homogenous': True,
-                'table': 'T',
-                'columns': ['f1', 'f2'],
-                'values': [('a', 'b',), ('c', 'd',)],
+                'sql_params_list': [
+                    ('INSERT INTO T (f1, f2) VALUES (%s, %s)', ('a', 'b')),
+                    ('INSERT INTO T (f1, f2) VALUES (%s, %s)', ('c', 'd'))
+                ],
             }
 
     Case d)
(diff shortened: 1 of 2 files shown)


larkee:
Could you run the Django system tests with and without this change to make sure that any there aren't any unexpected breakages?

IlyaFaer:
@larkee, I've used [the PR](https://github.com/googleapis/python-spanner-django/pull/603) to execute all the checks. Penultimate run wasn't using any special branches, while the last one does. Both have the same several failures, which seems to be unrelated.

larkee:
Confirmed the failures are the same and unrelated.

LGTM, could you update the docstring to match the changed behaviour? The original change was made to match the docstring.

google/cloud/spanner_dbapi/parse_utils.py
```diff
@@ -249,7 +249,7 @@ def parse_insert(insert_sql, params):
         it produces:
             {
                 'sql_params_list': [
-                    ('INSERT INTO T (f1, f2) VALUES (%s, LOWER(%s))', ('a', 'b',))
+                    ('INSERT INTO T (f1, f2) VALUES (%s, LOWER(%s))', ('a', 'b',)),
```
IlyaFaer, line 252: A comma is missing here - it's a `list`, so inner tuples must be separated with a comma.

larkee (approved):

PR #287, opened 2021-03-23
build: loosen the proto-plus dependency

### LINKED ISSUE #279: overly strict dependency on proto-plus
We noticed in building an internal tool that the Spanner dependency is much more strict than in other packages, such as BigQuery.

https://github.com/googleapis/python-spanner/blob/1b0ce1d2986085ce4033cf773eb6c5d3b904473c/setup.py#L36

Since `proto-plus` is a core dependency of all client libraries, it is often bundled with an environment (e.g. in AI Platform Notebooks). The strict dependency makes it much harder to upgrade the `google-cloud-spanner` library in these environments.

larkee:
Fixes #279


tswast (approved):
Thanks!

PR #291, opened 2021-03-24
docs: fix snapshot usage

kit494way:
Fix a wrong keyword argument `key_set`.
Correctly, it is `keyset`, and the value must be a instance of `spanner.KeySet()`.

`snapshot.read()` and `snapshot.execute_sql()` returns an iterable.
It does not have `rows` attribute.


larkee (approved):
Thank you for taking the time to correct the documentation! I really appreciate it :)

PR #313, opened 2021-04-22
feat: add support for low-cost instances

zoercai:
Allow users to create an instance using `processing_units`.
1 node = 1000 processing units. There is currently a default node count used if node count is not specified.
`create` and `update` now only update processing units rather than node count, as only one can be specified.

### DIFF
--- a/google/cloud/spanner_v1/client.py
+++ b/google/cloud/spanner_v1/client.py
@@ -49,7 +49,6 @@ from google.cloud.spanner_admin_instance_v1 import InstanceAdminClient
 from google.cloud.client import ClientWithProject
 from google.cloud.spanner_v1 import __version__
 from google.cloud.spanner_v1._helpers import _merge_query_options, _metadata_with_prefix
-from google.cloud.spanner_v1.instance import DEFAULT_NODE_COUNT
 from google.cloud.spanner_v1.instance import Instance
 from google.cloud.spanner_v1 import ExecuteSqlRequest
 from google.cloud.spanner_admin_instance_v1 import ListInstanceConfigsRequest
@@ -294,8 +293,9 @@ class Client(ClientWithProject):
         instance_id,
         configuration_name=None,
         display_name=None,
-        node_count=DEFAULT_NODE_COUNT,
+        node_count=None,
         labels=None,
+        processing_units=None,
     ):
         """Factory to create a instance associated with this client.
 
@@ -320,6 +320,10 @@ class Client(ClientWithProject):
         :param node_count: (Optional) The number of nodes in the instance's
                             cluster; used to set up the instance's cluster.
 
+        :type processing_units: int
+        :param processing_units: (Optional) The number of processing units
+                                allocated to this instance.
+
         :type labels: dict (str -> str) or None
         :param labels: (Optional) User-assigned labels for this instance.
 
@@ -334,6 +338,7 @@ class Client(ClientWithProject):
             display_name,
             self._emulator_host,
             labels,
+            processing_units,
         )
 
     def list_instances(self, filter_="", page_size=None):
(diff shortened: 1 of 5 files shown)


google/cloud/spanner_v1/client.py
```diff
@@ -289,6 +289,7 @@ def instance(
         configuration_name=None,
         display_name=None,
         node_count=DEFAULT_NODE_COUNT,
```
larkee, line 291: My understanding is that `node_count` and `processing_units` are exclusive (i.e. only one can be specified) so I would change this to `node_count=None` and add logic to set the default if neither are set and raise an error if both are set.

google/cloud/spanner_v1/client.py
```diff
@@ -314,6 +315,10 @@ def instance(
         :param node_count: (Optional) The number of nodes in the instance's
                             cluster; used to set up the instance's cluster.
 
+        :type processing_units: int
+        :param processing_units: (Optional) The number of processing units
+                                allocated to this instance.
```
larkee, line 320: Might be worth mentioning that you can only set either `node_count` or `processing_units`

google/cloud/spanner_v1/instance.py
```diff
@@ -306,13 +315,14 @@ def update(self):
...
-            Updates the ``display_name``, ``node_count`` and ``labels``. To change those
-            values before updating, set them via
+            Updates the ``display_name``, ``node_count``, ``processing_units``
+            and ``labels``. To change those values before updating, set them via
 
             .. code:: python
 
                 instance.display_name = 'New display name'
                 instance.node_count = 5
+                instance.processing_unit = 2000
```
larkee, line 325: This is invalid. You can't set both `node_count` and `processing_units`. Either revert this or remove the `node_count` value.

tests/unit/test_client.py
```diff
@@ -37,6 +37,7 @@ class TestClient(unittest.TestCase):
     INSTANCE_NAME = "%s/instances/%s" % (PATH, INSTANCE_ID)
     DISPLAY_NAME = "display-name"
     NODE_COUNT = 5
+    PROCESSING_UNITS = 2000
```
larkee, line 40: nit: set this to 5000 so it is equivalent to `NODE_COUNT`.

larkee (changes requested):

leahecole:
I think it's adding me as a reviewer because of the synthtool changes in `.github/header-checker-lint.yml` and those LGTM :)

google/cloud/spanner_v1/instance.py
```diff
@@ -125,6 +128,11 @@ def __init__(
         self.configuration_name = configuration_name
         self.node_count = node_count
         self.processing_units = processing_units
+        if node_count is None and processing_units is None:
+            self.node_count = DEFAULT_NODE_COUNT
+        if node_count is not None and processing_units is not None:
+            raise InvalidArgument(
+                'Only one of node count and processing units can be set.')
```
larkee, line 135: nit: put the validation check before setting the default

If the validation fails then the default isn't needed.

google/cloud/spanner_v1/instance.py
```diff
@@ -94,10 +95,12 @@ class Instance(object):
 
     :type node_count: int
     :param node_count: (Optional) Number of nodes allocated to the instance.
+                        Cannot be set if processing_units is set.
```
larkee, line 98: nit: use `processing_units` for the docs render. Also applies to processing_units docstring below

larkee (commented):
Just a couple comments to go with the proposed changes to resolve the update issue we discussed :+1:

larkee (approved):

google/cloud/spanner_v1/instance.py
```diff
@@ -190,6 +211,44 @@ def name(self):
...
+        """
+        self._processing_units = value
+        self._node_count = value // PROCESSING_UNITS_PER_NODE
+
+    @property
+    def node_count(self):
+        """Node count used in requests.
+
+        :rtype: int
+        :returns node_count:
```
larkee, line 237: nit: returns don't need a name
```suggestion
        :returns:
```

PR #325, opened 2021-04-28
perf: use protobuf for metadata to reduce type conversions

### LINKED ISSUE #310: Performance issues for queries returning larger arrays
We recently upgraded the google-cloud-spanner library to the latest version (1.17.1 -> 3.3.0) as it includes a fix for a bug we were experiencing. Unfortunately, we are now experiencing some large performance issues when executing queries that return some rows with large arrays: one specific query went from 4 to 17 seconds. This results in longer calculation times in our pipelines. 

We investigated the issue and found that the performance drop was a result of the [_merged_values](https://github.com/googleapis/python-spanner/blob/master/google/cloud/spanner_v1/streamed.py#L96) taking longer using the latest version of the library. This function parses the protobuf using `_parse_value_pb` and timing the parsing of the query mentioned above on both versions of the library gives the following result: 

| proto type | # calls | avg execution time v1.17.1 (s) | avg execution time v3.3.0 (s) |
|:----------:|:-------:|:------------------------------:|:-----------------------------:|
|    BOOL    |   672   |          4.00 x 10^-6          |          3.87 x 10^-6         |
|    INT64   |   384   |          7.28 x 10^-6          |          7.14 x 10^-6         |
|   FLOAT64  |   960   |          5.11 x 10^-6          |          4.40 x 10^-6         |
|  TIMESTAMP |   480   |          8.12 x 10^-5          |          8.70 x 10^-5         |
|    DATE    |   672   |          4.57 x 10^-5          |          4.21 x 10^-5         |
|   STRING   |   2112  |          4.46 x 10^-6          |          3.93 x 10^-6         |
|    ARRAY   |   1920  |          4.43 x 10^-3          |          3.96 x 10^-2         |

As you can see, parsing an array has become a lot slower which is an issue if it is called quite often. This seems similar to [another issue](https://github.com/googleapis/python-spanner/issues/207) in this repo but the performance fixes mentioned over there are already merged.

#### Environment details

- OS type and version: macOs Big Sur (version 11.2.3)
- Python version: Python 3.7.5
- pip version: pip 21.0.1
- google-cloud-spanner version: 3.3.0

#### Steps to reproduce
Execute a query that returns one or more array columns with the arrays containing a large number of elements (for example: 999)

larkee:
To fix performance issues in #207, the streaming logic switched to using the underlying protobuf objects instead of the `proto-plus` objects which have an type conversion overhead when accessing values. However, the metadata proto was kept in the `proto-plus` format. This metadata object contains the type information of the different fields which means it is frequently accessed when iterating over arrays. This causes major performance issues for large scale arrays due to the type conversion overhead. Using the underlying protobuf for the metadata removes this overhead and brings the performance in line with `v1.17.1`.

Fixes #310

### DIFF
--- a/google/cloud/spanner_v1/_helpers.py
+++ b/google/cloud/spanner_v1/_helpers.py
@@ -205,7 +205,7 @@ def _parse_value_pb(value_pb, field_type):
             _parse_value_pb(item_pb, field_type.struct_type.fields[i].type_)
             for (i, item_pb) in enumerate(value_pb.list_value.values)
         ]
-    elif field_type.code == TypeCode.NUMERIC:
+    elif type_code == TypeCode.NUMERIC:
         return decimal.Decimal(value_pb.string_value)
     else:
         raise ValueError("Unknown type: %s" % (field_type,))
--- a/google/cloud/spanner_v1/streamed.py
+++ b/google/cloud/spanner_v1/streamed.py
@@ -66,7 +66,9 @@ class StreamedResultSet(object):
         :rtype: :class:`~google.cloud.spanner_v1.types.ResultSetMetadata`
         :returns: structure describing the results
         """
-        return ResultSetMetadata.wrap(self._metadata)
+        if self._metadata:
+            return ResultSetMetadata.wrap(self._metadata)
+        return None
 
     @property
     def stats(self):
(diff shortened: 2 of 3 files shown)


zoercai (approved):

vi3k6i5 (commented):
LGTM

vi3k6i5 (approved):

PR #340, opened 2021-05-12
feat: add decimal validation for numeric precision and scale supported by Spanner

### LINKED ISSUE #339: Add validation in decimal to numeric conversion
Add validation in decimal to numeric conversion to ensure precision and recall is supported by spanner.

read more here: https://cloud.google.com/spanner/docs/data-types#numeric_type.

Similar validations are present in other clients eg: https://github.com/googleapis/java-spanner/blob/master/google-cloud-spanner/src/main/java/com/google/cloud/spanner/Value.java#L141-L155

vi3k6i5:
feat: add decimal validation for numeric precision and scale supported by spanner
Fixes #339 🦕


google/cloud/spanner_dbapi/parse_utils.py
```diff
@@ -509,11 +518,37 @@ def sql_pyformat_args_to_spanner(sql, params):
             resolved_value = pyfmt % params
             named_args[key] = resolved_value
         else:
+            assert_numeric_precision_and_scale(params[i])
             named_args[key] = params[i]
 
     return sanitize_literals_for_upload(sql), named_args
 
 
+def assert_numeric_precision_and_scale(value):
```
olavloite, line 527: I don't know whether this is normal in the Python client, but it seems a little strange that this function is called for __all__ parameter types, and not only numeric values. I would have expected either:
1. That this function is only called for numeric values (i.e. the `if isinstance(value, decimal.Decimal)` would be outside of the function).
2. Or: That there is a generic 'assert_valid_value(value)` function that first checks the type of the input parameter, and then calls the appropriate validation function for that type.

olavloite (approved):
This seems good to my untrained Python eyes, with a small question on whether it is logical to call the validate function for all values, or whether it would be better to move the if-check for the type out of the validation function.
(You should in any case await approval from other reviewers before merging)

google/cloud/spanner_dbapi/parse_utils.py
larkee, line 527: I think this logic would make more sense to be added to the base client [here](https://github.com/googleapis/python-spanner/blob/master/google/cloud/spanner_v1/_helpers.py#L131-L132).

vi3k6i5: makes sense added an instance check before the method is called.

vi3k6i5: moved it to base client.

vi3k6i5:
> This seems good to my untrained Python eyes, with a small question on whether it is logical to call the validate function for all values, or whether it would be better to move the if-check for the type out of the validation function.
> (You should in any case await approval from other reviewers before merging)

moved the method after a check. and also moved it to base client as @larkee suggested.

google/cloud/spanner_v1/_helpers.py
```diff
@@ -87,6 +97,27 @@ def _merge_query_options(base, merge):
...
 
 
+def _assert_numeric_precision_and_scale(value):
+    """
+    Spanner supports fixed 38 digits of precision and 9 digits of scale.
+    This number can be optionally prefixed with a plus or minus sign.
+    Read more here: https://cloud.google.com/spanner/docs/data-types#numeric_type
+
+    Asserts that input numeric field is within spanner supported range.
+    :type value: Any
```
larkee, line 107:
```suggestion
    :type value: decimal.Decimal
```

google/cloud/spanner_v1/_helpers.py
```diff
@@ -87,6 +97,27 @@ def _merge_query_options(base, merge):
     return combined
 
 
+def _assert_numeric_precision_and_scale(value):
+    """
+    Spanner supports fixed 38 digits of precision and 9 digits of scale.
+    This number can be optionally prefixed with a plus or minus sign.
+    Read more here: https://cloud.google.com/spanner/docs/data-types#numeric_type
+
+    Asserts that input numeric field is within spanner supported range.
```
larkee, line 106:
```suggestion
    Asserts that input numeric field is within Spanner supported range.
```
Also this should be moved to the top of the docstring. That way you have an explanation of what function does, followed by why it's needed.

google/cloud/spanner_v1/_helpers.py
```diff
@@ -87,6 +97,27 @@ def _merge_query_options(base, merge):
...
+    :type value: Any
+    :param value: The value to check for Cloud Spanner compatibility.
+
+    :raises NotSupportedError: if value is not within supporteed precision or scale of spanner.
```
larkee, line 110:
```suggestion
    :raises NotSupportedError: if value is not within supported precision or scale of Spanner.
```

tests/unit/test__helpers.py
```diff
@@ -233,6 +233,66 @@ def test_w_unknown_type(self):
...
+            decimal.Decimal("42"),
+            decimal.Decimal("9.9999999999999999999999999999999999999E+28"),
+            decimal.Decimal("-9.9999999999999999999999999999999999999E+28"),
+            decimal.Decimal("99999999999999999999999999999.999999999"),
+            decimal.Decimal("1E+28"),
+            decimal.Decimal("1E-9"),
+        ]
+        for value in cases:
+            with self.subTest(value=value):
+                value = decimal.Decimal("9999999999999999999999999999.999999999")
```
larkee, line 250: I don't think this line is needed. The value should be coming from the cases.

google/cloud/spanner_v1/_helpers.py
vi3k6i5: fixed.

google/cloud/spanner_v1/_helpers.py
vi3k6i5: Done.

tests/unit/test__helpers.py
vi3k6i5: changed to read from cases.

google/cloud/spanner_v1/_helpers.py
vi3k6i5: fixed comment.

larkee (approved):
LGTM :+1:  You could remove `test_w_numeric` if you wanted since `test_w_numeric_precision_and_scale_valid` should have more coverage.

PR #358, opened 2021-06-02
fix: update to support the open-telemetry status code spec change

### LINKED ISSUE #357: Update open telemetry tracing code according to new spec.
The method used in [open telemetry tracing](https://github.com/googleapis/python-spanner/blob/master/google/cloud/spanner_v1/_opentelemetry_tracing.py#L24) has been updated in this [commit]( https://github.com/lzchen/opentelemetry-python/commit/cf13537eaa8f6159598b5c17e1159b9571a3996c#diff-af63e517bdd2a33c2a8833eb083e0b0f640e0c62a591f0d289deb42d5cb3ce60R25)

#### Environment details
opentelemetry-api                    1.2.0
opentelemetry-sdk                   1.2.0
google-cloud-spanner             3.4.0

#### Steps to reproduce

```python
from opentelemetry.trace.status import Status, StatusCanonicalCode    
```

```python   
from opentelemetry.instrumentation.utils import http_status_to_canonical_code
```

#### Stack trace
```
ImportError: cannot import name 'StatusCanonicalCode' from 'opentelemetry.trace.status'
```

SO question link: https://stackoverflow.com/questions/67775140/getting-import-error-while-trying-to-import-statuscanonicalcode-using-openteleme

vi3k6i5:
fix: for open-telemetry status code spec change

Fixes #357 🦕

Open telemetry has changed their specs for status code. Since the open telemetry library for python is still in beta it doesn't make much sense to use/support older version of that library.

Couple of reference:
1. ['StatusCanonicalCode' is not part of 'opentelemetry.trace.status'](https://stackoverflow.com/questions/67775140/getting-import-error-while-trying-to-import-statuscanonicalcode-using-openteleme)
2. [In memory exporter not resetting after test case ends](https://stackoverflow.com/questions/67826617/open-telemetry-inmemoryspanexporter-not-reseting-after-test-case-is-completed)


larkee (commented):
Is this compatible with older versions of OpenTelemetry?

Also, the py3.6 unit tests are failing.

vi3k6i5:
> Is this compatible with older versions of OpenTelemetry?

Don't think we should support any older versions of the library. As per the [library doc](https://opentelemetry-python.readthedocs.io/en/latest/) the software is in Beta, and since they have changed their api, so not sure if we should support older versions?

> Also, the py3.6 unit tests are failing.

Fixed the unit tests and system tests.

larkee (commented):
The emulator action is timing out while trying to resolve the dependencies. It seems to have an issue with `opentelemetry-instrumentation == 0.22b0` in the constraints file. Using `opentelemetry-instrumentation == 0.20b0` seems to resolve it but I don't know if this is a valid version..

vi3k6i5:
> The emulator action is timing out while trying to resolve the dependencies. It seems to have an issue with `opentelemetry-instrumentation == 0.22b0` in the constraints file. Using `opentelemetry-instrumentation == 0.20b0` seems to resolve it but I don't know if this is a valid version..

Thanks for the help, that fixed the issue. Thanks 👍

larkee (approved):

PR #360, opened 2021-06-03
fix: classify batched DDL statements

larkee:
Currently an SQL statement such as `"CREATE TABLE t (...); INSERT INTO T (...)"` would be classified as batched DDL despite not all the statements being DDL.

This PR adds a classification check for each of the statements and raises an error if any of the batched statements are non-DDL.

### DIFF
--- a/google/cloud/spanner_dbapi/cursor.py
+++ b/google/cloud/spanner_dbapi/cursor.py
@@ -182,9 +182,7 @@ class Cursor(object):
                         if ddl[-1] == ";":
                             ddl = ddl[:-1]
                         if parse_utils.classify_stmt(ddl) != parse_utils.STMT_DDL:
-                            raise ValueError(
-                                "Only DDL statements may be batched."
-                            )
+                            raise ValueError("Only DDL statements may be batched.")
                         ddl_statements.append(ddl)
                 # Only queue DDL statements if they are all correctly classified.
                 self.connection._ddl_statements.extend(ddl_statements)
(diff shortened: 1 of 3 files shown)


google/cloud/spanner_dbapi/cursor.py
```diff
@@ -174,10 +174,14 @@ def execute(self, sql, args=None):
...
             if classification == parse_utils.STMT_DDL:
+                ddl_statements = []
                 for ddl in sql.split(";"):
                     ddl = ddl.strip()
                     if ddl:
-                        self.connection._ddl_statements.append(ddl)
+                        if parse_utils.classify_stmt(ddl) != parse_utils.STMT_DDL:
+                            raise ValueError("Only DDL statements may be batched.")
+                        ddl_statements.append(ddl)
+                self.connection._ddl_statements.extend(ddl_statements)
```
IlyaFaer, line 184: Maybe I'm missing something, but I don't see strong reasons to initiate `ddl_statements` and then extend `self.connection._ddl_statements`. I assume ddls can be appended into `self.connection._ddl_statements` directly.

IlyaFaer (approved):
LGTM

google/cloud/spanner_dbapi/cursor.py
larkee: If one statement is invalid, none of the statements is executed. In the case that there is a non-DDL statement, I think it's less confusing to queue none of the statements than to queue the statements up to the non-DDL one. WDYT?

IlyaFaer: Ah, if to add first `n` ddls and fail than the next `execute()` call will execute these first `n` ddls as well. Which would be quiet unexpected. Yeah, that's a good catch!

PR #372, opened 2021-06-17
fix(db_api): use sqlparse to split DDL statements

IlyaFaer:
Instead of simple `str.split(";")` method use more smart `sqlparse` package to split DDL statements executed in a form:
```python
cursor.execute("""
    ddl_statement1;
    ddl_statement2;
    ddl_statement3;
""")
```

### DIFF
--- a/tests/unit/spanner_dbapi/test_cursor.py
+++ b/tests/unit/spanner_dbapi/test_cursor.py
@@ -939,16 +939,16 @@ class TestCursor(unittest.TestCase):
         from google.cloud.spanner_dbapi.connection import connect
 
         EXP_DDLS = [
-            "CREATE TABLE table_name (row_id INT64) PRIMARY KEY ();",
-            "DROP INDEX index_name;",
+            "CREATE TABLE table_name (row_id INT64) PRIMARY KEY ()",
+            "DROP INDEX index_name",
             (
                 "CREATE TABLE papers ("
                 "\n    id INT64,"
                 "\n    authors ARRAY<STRING(100)>,"
                 '\n    author_list STRING(MAX) AS (ARRAY_TO_STRING(authors, ";")) stored'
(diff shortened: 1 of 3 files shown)


tests/unit/spanner_dbapi/test_cursor.py
```diff
@@ -956,7 +963,12 @@ def test_ddls_with_semicolon(self):
         cursor.execute(
             "CREATE TABLE table_name (row_id INT64) PRIMARY KEY ();"
             "DROP INDEX index_name;\n"
-            "DROP TABLE table_name;"
+            "CREATE TABLE papers ("
+            "\n    id INT64,"
+            "\n    authors ARRAY<STRING(100)>,"
+            '\n    author_list STRING(MAX) AS (ARRAY_TO_STRING(authors, ";")) stored'
+            ") PRIMARY KEY (id);"
+            "DROP TABLE table_name;",
```
IlyaFaer, line 971: The only difference `sqlparse.split()` brings is that it keeps `;` symbols at the end of statements.

IlyaFaer:
According to what I see in the `sqlparse` source code and documentation, it should work okay with GoogleSQL. Plus to the test case I've tried several types of DDL statements (interleaved tables, alters, indexes - cases, which seems complex to me). All was splitted correсtly.

tests/unit/spanner_dbapi/test_cursor.py
larkee: That is correct. Spanner expects the DDL statements to not have a `;` character. A simple check can be added after the null check like so:
```
if ddl:
    if ddl[-1] == ";":
        ddl = ddl[:-1]
    self.connection._ddl_statements.append(ddl)
```

larkee (approved):

PR #399, opened 2021-07-06
feat: add configurable leader placement support

zoercai:


hengfengli (approved):
LGTM.

hengfengli:
Hi @zoercai , the error is: 

```
E   google.api_core.exceptions.FailedPrecondition: 400 The instance configuration projects/***/instanceConfigs/regional-us-central1 does not support default_leader settings.
```

We need to create a multi-region instance in order to test `default_leader`. `default_leader` does not work for a regional instance. Two options we have here: 

1) Create a multi-region instance to test default_leader
2) Do the integration test in sample tests (this is what we do in java)

tests/system/test_system.py
```diff
@@ -358,11 +358,14 @@ def tearDownClass(cls):
...
 
     def setUp(self):
+        self.instances_to_delete = []
         self.to_delete = []
 
     def tearDown(self):
         for doomed in self.to_delete:
             doomed.drop()
+        for instance in self.instances_to_delete:
+            instance.delete()
```
larkee, line 368: By having this logic here, we have to create and delete an instance for every test that needs a multi-region instance. This is likely going to increase the number of `RESOURCE_EXHAUSTED` errors we see.

I think moving the instance creation into `setUpClass` similar to how `cls._db` is done would be better. That way the tests can all reuse the instance :+1:

tests/system/test_system.py
```diff
@@ -443,6 +446,58 @@ def test_create_database_pitr_success(self):
...
+        operation.result(30)  # raises on failure / timeout.
+
+        database_ids = [database.name for database in instance.list_databases()]
+        self.assertIn(temp_db.name, database_ids)
+
+        temp_db.reload()
+        self.assertEqual(temp_db.default_leader, default_leader)
+
+        with self.assertRaises(exceptions.InvalidArgument):
+            temp_db.create()
```
larkee, line 490: This doesn't make sense. The database already exists. With the above change to instance create, I would move this to a separate test like `test_create_database_with_default_leader_invalid_args`.

tests/system/test_system.py
```diff
@@ -443,6 +446,58 @@ def test_create_database_pitr_success(self):
...
+        pool = BurstyPool(labels={"testcase": "create_database_default_leader"})
+
+        # Create a multi-region instance
+        multi_region_config = "nam3"
+        ALT_INSTANCE_ID = "new" + unique_resource_id("-")
+        config_name = "{}/instanceConfigs/{}".format(
+            Config.CLIENT.project_name, multi_region_config
+        )
+        instance = Config.CLIENT.instance(
+            instance_id=ALT_INSTANCE_ID, configuration_name=config_name
```
larkee, line 462: Please add the labels used by the other instance, see [here](https://github.com/googleapis/python-spanner/blob/e939fa7fbf5472f117f7150c5bd54fb0ccdc9810/tests/system/test_system.py#L151-L156). They are used to clean up the instance if there is a resource leak, e.g. if the testing is interrupted before clean up can happen.

larkee (changes requested):
Implementation LGTM. 

Some changes are needed for the system tests

tests/system/test_system.py
zoercai: Thanks for spotting this, I must've accidentally pasted this in. I've deleted it now.

tests/system/test_system.py
```diff
@@ -353,16 +353,34 @@ def setUpClass(cls):
             SPANNER_OPERATION_TIMEOUT_IN_SECONDS
         )  # raises on failure / timeout.
 
+        # Create a multi-region instance
+        multi_region_config = "nam3"
+        ALT_INSTANCE_ID = "new" + unique_resource_id("-")
```
larkee, line 358: nit: All caps are usually used for global/static constants. Move this to the top with `INSTANCE_ID` so it's clear what format the instance IDs are generated. Additionally, rather than `new` perhaps `multi-region` would be better. That way it's clear what the instance is for.

larkee (approved):
Minor nit for Python style but the rest LGTM :+1:

PR #412, opened 2021-07-14
feat: use DML batches in `executemany()` method

IlyaFaer:


google/cloud/spanner_dbapi/cursor.py
```diff
@@ -258,9 +259,21 @@ def executemany(self, operation, seq_of_params):
 
         many_result_set = StreamedManyResultSets()
 
-        for params in seq_of_params:
-            self.execute(operation, params)
-            many_result_set.add_iter(self._itr)
+        if classification == parse_utils.STMT_INSERT:
```
olavloite, line 262: If I understand this correctly, this method will now use mutations for all insert statements, and DML for all other update statements (`UPDATE`, `DELETE`). I think we should try to make this in a slightly different way, so users can have some more control over it, and it can be used for update and delete as well. I would suggest:
1. Use mutations by default when the connection is in auto-commit mode. Reason: Mutations do not support read-your-writes, but that is not possible in auto-commit mode anyways.
2. Use DML by default when the connection is in transactional mode (for the same reason as above).
3. Let the user configure whether a specific transaction should use DML or mutations. This could be through a custom method or a custom SQL statement. As far as I know, dbapi does not have a standard method for setting the isolation level for transactions. That is the method that we [used for ActiveRecord](https://github.com/googleapis/ruby-spanner-activerecord/blob/b083a8750babc2653fd4dc3b2ffa7ced47bf41c6/examples/snippets/mutations/application.rb#L19).
4. Also let the user turn off the use of mutations in auto-commit mode. Dbapi only accepts SQL strings as input, and this implementation tries to 'parse' the SQL string and convert it to a mutation in a relatively simple way. If the SQL string contains special characters, comments, etc. this parsing would currently fail.

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -258,9 +259,21 @@ def executemany(self, operation, seq_of_params):
...
+            match = RE_INSERT.search(operation)
+
+            table_name = match["table_name"].strip("`")
+
+            transaction = self.connection.transaction_checkout()
+            transaction.insert(
+                table=table_name,
+                columns=[col.strip('" ') for col in match["columns"].split(",")],
```
olavloite, line 270: This could fail if the SQL string contains comments that include a " or other special characters. Also, if the SQL string contains string literals with quotes or other special characters, this will also fail. So I think this should be considered a 'best-effort' that works with simple SQL strings that are generated by frameworks like SQLAlchemy, but it should not be something that all users **must** use for everything.

IlyaFaer: Yeah, there was a couple of edge cases, I've added some changes today and a test to check that column are designated correctly.

google/cloud/spanner_dbapi/connection.py
```diff
@@ -63,6 +63,7 @@ def __init__(self, instance, database):
 
         self.is_closed = False
         self._autocommit = False
+        self._use_mutations = False
```
olavloite, line 66: I think it would be better to allow three different values (so using something like an enum instead of boolean):
- Never
- OnlyForAutocommit (Default)
- Always

The user should have the option of turning the feature completely off, also when in autocommit, as we are trying to parse the SQL statements in `execute_many`. As that parsing is a best-effort feature, there will be cases when the parsing does not work as intended, and the user should have the option of turning the usage of mutations off.

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -258,9 +259,30 @@ def executemany(self, operation, seq_of_params):
 
         many_result_set = StreamedManyResultSets()
 
-        for params in seq_of_params:
-            self.execute(operation, params)
-            many_result_set.add_iter(self._itr)
+        if self.connection.autocommit or self.connection.use_mutations:
+            if classification == parse_utils.STMT_INSERT:
```
olavloite, line 263: I'm not sure whether this is still work in progress, but to be sure: This `if` statement has no `else`, which means that if a user executes an update or delete statement, nothing will be executed (in autocommit or use_mutations mode).

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -258,9 +258,26 @@ def executemany(self, operation, seq_of_params):
...
+                sql, params = parse_utils.sql_pyformat_args_to_spanner(
+                    operation, params
+                )
+                statements.append((sql, params, get_param_types(params)))
+
+            transaction = self.connection.transaction_checkout()
+            _, res = transaction.batch_update(statements)
+
+            if self.connection.autocommit:
+                transaction.commit()
```
skuruppu, line 274: Is there a test to verify that `commit()` is called in this case?

tests/unit/spanner_dbapi/test_cursor.py
```diff
@@ -337,6 +337,42 @@ def test_executemany(self):
...
+                (
+                    """INSERT INTO table (col1, "col2", `col3`, `"col4"`) VALUES (@a0, @a1, @a2, @a3)""",
+                    {"a0": 1, "a1": 2, "a2": 3, "a3": 4},
+                    {"a0": INT64, "a1": INT64, "a2": INT64, "a3": INT64},
+                ),
+                (
+                    """INSERT INTO table (col1, "col2", `col3`, `"col4"`) VALUES (@a0, @a1, @a2, @a3)""",
+                    {"a0": 5, "a1": 6, "a2": 7, "a3": 8},
+                    {"a0": INT64, "a1": INT64, "a2": INT64, "a3": INT64},
+                ),
```
skuruppu, line 372: I would add tests for delete and update as well to verify that they're batched as expected all part of the batch. Also that non-DML statements are not batched.

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -258,9 +258,26 @@ def executemany(self, operation, seq_of_params):
...
+            statements = []
+
+            for params in seq_of_params:
+                sql, params = parse_utils.sql_pyformat_args_to_spanner(
+                    operation, params
+                )
+                statements.append((sql, params, get_param_types(params)))
+
+            transaction = self.connection.transaction_checkout()
+            _, res = transaction.batch_update(statements)
```
olavloite, line 271: The usage of `batch_update` is a little bit tricky, and you need to check the status code here. The semantics are as follows:
1. It could be that the `batch_update` function fails with an error. That will be handled like any other error, so there is no special action required for that.
2. It could also be that one of the statements in the list of statements returns an error. If for example your list of statements has 4 statements. If statement number 3 fails, the return value of `batch_update` will be the status code of the error + the update count of the first 2 statements that did succeed (so the length of the second return value should be 2). The 4th statement will not be executed.
3. The error code that is returned as the first return value of this function can also be `Aborted`. Using the same example as above: This means that the first 2 statements executed successfully, but that during the execution of the third statement the transaction was aborted by Cloud Spanner. This should be handled as if the entire statement aborted the transaction (so it should cause a retry of the transaction).

The DbApi specification states for the `executemany` function that the `Return values are not defined.`, so we are free to choose what to return in case an error code is returned by Spanner. The best might be to return the same as Cloud Spanner, which means both the error code and the update counts, except when the error code is `Aborted`. If the error code is `Aborted`, then it's (in my opinion) better to throw an `AbortedError`, as there is nothing more that a user can do with the transaction anyways, and the entire transaction has already been rolled back.

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -258,9 +258,26 @@ def executemany(self, operation, seq_of_params):
 
         many_result_set = StreamedManyResultSets()
 
-        for params in seq_of_params:
-            self.execute(operation, params)
-            many_result_set.add_iter(self._itr)
+        if classification in (parse_utils.STMT_INSERT, parse_utils.STMT_UPDATING):
```
olavloite, line 261: I think we should skip the parsing here and assume that the method is only called with DML statements. The DbApi specification states that implementations are free to accept or refuse statements that would return result sets. Cloud Spanner will refuse queries in the `ExecuteBatchDml` RPC, so we can leave that check to the backend. That means that this implementation will only support DML in `executemany`, but that is not an unreasonable limitation.

google/cloud/spanner_dbapi/cursor.py
IlyaFaer: The thing is that we don't ever return anything from `execute()` and `executemany()` methods. If one wants to get rows, they should use `fetch*()` methods after `executemany()` method finished. In our case it also will be possible as we're adding DML batch result into cursor results object.

google/cloud/spanner_dbapi/connection.py
```diff
@@ -201,17 +201,14 @@ def transaction_checkout(self):
...
-        if not self.autocommit:
-            if not self.inside_transaction:
-                self._transaction = self._session_checkout().transaction()
-                self._transaction.begin()
+        if not self.inside_transaction:
+            self._transaction = self._session_checkout().transaction()
+            self._transaction.begin()
 
-            return self._transaction
+        return self._transaction
```
IlyaFaer, line 211: As we're going to use transactions for batches in `autocommit` mode, this method now should also work in `autocommit` mode.

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -258,9 +260,56 @@ def executemany(self, operation, seq_of_params):
...
+                )
+                statements.append((sql, params, get_param_types(params)))
+
+            if self.connection.autocommit:
+                transaction = self.connection.transaction_checkout()
+                status, res = transaction.batch_update(statements)
+                many_result_set.add_iter(res)
+
+                if status.code != OK:
+                    self.connection._transaction = None
```
olavloite, line 278: You should explicitly rollback the transaction here. The transaction will be aborted by Cloud Spanner after 10 seconds if you just keep it open here, but that would mean that it will unnecessarily hold on to locks for 10 seconds.

(This comment is void if we use `run_in_transaction`, as that will handle the transaction for you.)

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -258,9 +260,56 @@ def executemany(self, operation, seq_of_params):
...
+
+                        res_checksum = ResultsChecksum()
+                        if not retried:
+                            self.connection._statements.append(
+                                (statements, res_checksum)
+                            )
+
+                        status, res = transaction.batch_update(statements)
+                        many_result_set.add_iter(res)
+                        res_checksum.consume_result(res)
```
olavloite, line 296: The status should be added to the checksum as well. Consider the following situation:
1. During the initial attempt batch_update returns [1, 1, 1], Status.INVALID_ARGUMENT
2. During a retry attempt batch_update returns [1, 1, 1], Status.ALREADY_EXISTS

These two results should not be considered equal and the retry should be aborted.

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -258,9 +260,56 @@ def executemany(self, operation, seq_of_params):
...
+
+                        if status.code == ABORTED:
+                            self.connection._transaction = None
+                            raise Aborted(status.details)
+                        elif status.code != OK:
+                            self.connection._transaction = None
```
olavloite, line 302: The transaction should not be invalidated here. If the status for example is `ALREADY_EXISTS`, the client application should be allowed to execute a new insert statement with a different key. Also, any other error code (other than `ABORTED`) is something that the client application should be allowed to inspect and possibly recover from. If it does not want to recover from it, then the client application should rollback the transaction.

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -258,9 +260,56 @@ def executemany(self, operation, seq_of_params):
...
+            statements = []
+
+            for params in seq_of_params:
+                sql, params = parse_utils.sql_pyformat_args_to_spanner(
+                    operation, params
+                )
+                statements.append((sql, params, get_param_types(params)))
+
+            if self.connection.autocommit:
+                transaction = self.connection.transaction_checkout()
```
olavloite, line 273: I think it would be better to handle this in the same way as that we are handling single statements in autocommit, which means just using the `run_in_transaction` function in the client library: https://github.com/googleapis/python-spanner/blob/62b937d6889fc56a64e5073985bab8ccd5d8c24d/google/cloud/spanner_dbapi/cursor.py#L232

tests/unit/spanner_dbapi/test_connection.py
```diff
@@ -157,9 +157,6 @@ def test_transaction_checkout(self):
         mock_transaction.committed = mock_transaction.rolled_back = False
         self.assertEqual(connection.transaction_checkout(), mock_transaction)
 
-        connection._autocommit = True
```
olavloite, line 160: If you use `run_in_transaction` for batch updates in autocommit mode, then you don't have to change this behavior.

google/cloud/spanner_dbapi/cursor.py
IlyaFaer: Done

google/cloud/spanner_dbapi/cursor.py
IlyaFaer: Hm-m, yeah, it makes sense. Done

google/cloud/spanner_dbapi/cursor.py
IlyaFaer: Agreed, done

google/cloud/spanner_dbapi/cursor.py
IlyaFaer: I had a thought about it, but it will not give anything, you know, significant, while testing will become more complex (because transactions will be initiated deeper in the API). Also it'll be little bit harder to get part of the result in case of an error failed during the batch execution.

So, it looks like a change to the same.

olavloite: I don't agree that it is the same. If you use `run_in_transaction`, aborted transactions will automatically be retried by the client library. In this implementation any aborted transaction will be propagated as an error to the client application. That means that if you use `execute` in autocommit mode, statements will automatically be retried if the underlying transaction is aborted, while if you use `execute_many` in autocommit mode, the aborted error will be propagated to the client application. I don't think that is a very good API.
The other option would be to use the checksum retry logic for autocommit as well, but that seems like over-complicating things to me.

tests/unit/spanner_dbapi/test_cursor.py
IlyaFaer: Done

google/cloud/spanner_dbapi/connection.py
```diff
@@ -175,43 +178,58 @@ def _rerun_previous_statements(self):
...
+                statements, checksum = statement
+
+                transaction = self.transaction_checkout()
+                status, res = transaction.batch_update(statements)
+
+                if status.code == ABORTED:
+                    self.connection._transaction = None
+                    raise Aborted(status.details)
+                elif status.code != OK:
+                    raise OperationalError(status.details)
```
olavloite, line 191: This should not be propagated to the client application. The status code should be included in the checksum check, and if the statement returned the same status code as the during the initial attempt, then everything is fine.

google/cloud/spanner_dbapi/connection.py
```diff
@@ -175,43 +177,56 @@ def _rerun_previous_statements(self):
...
-            if statement != self._statements[-1]:
-                for res in res_iter:
-                    retried_checksum.consume_result(res)
-
-                _compare_checksums(statement.checksum, retried_checksum)
-            # executing the failed statement
+            if isinstance(statement, list):
+                statements, checksum = statement
+
+                transaction = self.transaction_checkout()
```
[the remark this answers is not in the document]
IlyaFaer: The `transaction_checkout()` method uses `_session_checkout()` method to get session:
https://github.com/googleapis/python-spanner/blob/916ec7ac91945b50cc3dcb12792f3912cce07482/google/cloud/spanner_dbapi/connection.py#L226

And `_session_checkout()` method checks if a session object is already obtained by the connection. If not, it'll be taken from the pool, otherwise the previously taken session will be used:
https://github.com/googleapis/python-spanner/blob/916ec7ac91945b50cc3dcb12792f3912cce07482/google/cloud/spanner_dbapi/connection.py#L136

So, it looks like it should work fine. I'll take a closer look at the tests, probably it already checked somewhere - if not, I think we can add a check.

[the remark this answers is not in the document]
olavloite: You're right, the `_session_checkout()` function seems to handle that correctly.

olavloite (approved):

larkee (commented):
LGTM, but would you mind updating the DML in the `test_execute_many` [system test](https://github.com/googleapis/python-spanner/blob/master/tests/system/test_system_dbapi.py#L346-L353) so that we are testing this feature against the emulator and production?

larkee (approved):

PR #428, opened 2021-07-26
samples: add default leader options samples

zoercai:
Samples for https://github.com/googleapis/python-spanner/pull/399. Implementation code will be gone once https://github.com/googleapis/python-spanner/pull/399 is merged.


larkee (changes requested):
Currently, you are creating an instance within the `spanner_create_database_with_default_leader` and `spanner_update_database_with_default_leader` samples. 

This is an issue as if there is a failure later on in the sample, the instance does not get cleaned up correctly. Additionally the second sample fails as it tried to create the same instance.

The instance creation should be removed from the samples and you should instead be using `sample_instance`, or a pytest fixture that uses `sample_instance`. This will ensure that the instance clean up is handled correctly.

larkee (approved):
LGTM once the implementation is merged and released :+1:

larkee (commented):
Having just written the samples for PHP I noticed some inconsistencies. Just a few changes for the `list_*` samples

hengfengli:
@zoercai Why do we get an error `Instance already exist`? 

```
value = None
from_value = <_InactiveRpcError of RPC that terminated with:
	status = StatusCode.ALREADY_EXISTS
	details = "Instance already exist...message":"Instance already exists: projects/precise-truck-742/instances/google-cloud-1628122209464","grpc_status":6}"
>

>   ???
E   google.api_core.exceptions.AlreadyExists: 409 Instance already exists: projects/precise-truck-742/instances/google-cloud-1628122209464
```

PR #475, opened 2021-08-02
feat(db_api): add an ability to set ReadOnly/ReadWrite connection mode

IlyaFaer:
Proposal for the user's question: https://github.com/cloudspannerecosystem/python-spanner-sqlalchemy/issues/97
Will require small changes in SQLAlchemy dialect as well.


google/cloud/spanner_dbapi/connection.py
```diff
@@ -121,6 +127,31 @@ def instance(self):
...
+
+        Args:
+            value (bool): True for ReadOnly mode, False for ReadWrite.
+        """
+        if self.inside_transaction:
+            raise ValueError(
+                "Connection read/write mode can't be changed while a transaction is in progress. "
+                "Commit or rollback the current transaction and try again."
+            )
+        self._read_only = value
```
IlyaFaer, line 153: It's done in the same way JDBC works: connection mode can be changed, but only if there is no active transaction on it.

google/cloud/spanner_v1/transaction.py
```diff
@@ -100,7 +105,10 @@ def begin(self):
         database = self._session._database
         api = database.spanner_api
         metadata = _metadata_with_prefix(database.name)
-        txn_options = TransactionOptions(read_write=TransactionOptions.ReadWrite())
+        if read_only:
+            txn_options = TransactionOptions(read_only=TransactionOptions.ReadOnly())
```
olavloite, line 109: This will begin a read-only transaction on this connection. It is important to realize (and maybe also document) a couple of things when using a read-only transaction:
1. A read-only transaction will only read data at a specific point in time during the entire transaction. That means that when a user starts a read-only transaction, they will never see any new changes to the database until the read-only transaction has been closed and a new one begins. So if I start a new read-only transaction at time T1, and someone else executes and commits a read/write transaction at T2, the read-only transaction will still only see the data at T1 and not the data from T2. This can be surprising for many users, and it should be pointed out that most users who want to just read data should probably use autocommit mode, and not read-only transactional mode.
2. Read-only transactions cannot be committed or rolled back. You must therefore check the type of transaction that is in use when `commit` or `rollback` is called on the connection, and only do something with the transaction if it is a read/write transaction. (And of course return the session to the pool if it is a read-only transaction.)

tests/system/test_system_dbapi.py
```diff
@@ -426,6 +427,28 @@ def test_DDL_commit(self):
...
+    UPDATE contacts
+    SET first_name = 'updated-first-name'
+    WHERE first_name = 'first-name'
+    """
+            )
+
+        cur.execute("SELECT * FROM contacts")
+
+        with self.assertRaises(exceptions.FailedPrecondition):
+            conn.commit()
```
olavloite, line 450: This should not raise an exception. This should remove the read-only transaction from the connection and return the session to the pool. If there is no possibility of 'committing' (or otherwise ending) a read-only transaction, the connection will __never__ see any new data after the read-only transaction has started.

tests/system/test_system_dbapi.py
```diff
@@ -426,6 +427,28 @@ def test_DDL_commit(self):
         cur.execute("DROP TABLE Singers")
         conn.commit()
 
+    def test_read_only(self):
```
olavloite, line 430: It would be good to add an extra system test that shows the following:
1. A read-only transaction can read data that was present in the database when the read-only transaction started.
2. If data is added using a separate read/write transaction (i.e. another user added data to the database), those records are __not__ visible to the read-only transaction in step 1.
3. __After__ committing/rolling back (closing) the read-only transaction, and starting a new read-only transaction, the data from step 2 should be visible to the read-only transaction.

olavloite (changes requested):

tests/system/test_system_dbapi.py
IlyaFaer: It is something that comes from the backend with a message:
`400 Cannot commit or rollback a read-only or partitioned-dml transaction`. So, it's not us who raise this exception.

tests/system/test_system_dbapi.py
IlyaFaer: Done

google/cloud/spanner_v1/transaction.py
IlyaFaer: Okay, I've added changes for item 2 and a system test for item 1.
Looks like we need to find a proper place for mentioning all of it. I don't see any docs for connection API...

google/cloud/spanner_dbapi/connection.py
```diff
@@ -48,9 +48,14 @@ class Connection:
 
     :type database: :class:`~google.cloud.spanner_v1.database.Database`
     :param database: The database to which the connection is linked.
+
+    :type read_only: bool
+    :param read_only:
+        Flag to designate if the connection must use only ReadOnly
+        transaction type.
```
olavloite, line 55:
```suggestion
        Flag to indicate that the connection may only execute queries and no update or DDL statements.
        If autocommit is true, the connection will use a single use read-only transaction with strong timestamp
        bound for each new statement, and will immediately see any changes that have been committed by
        any other transaction.
        If autocommit is false, the connection will automatically start a new multi use read-only transaction
        with strong timestamp bound when the first statement is executed. This read-only transaction will be
        used for all subsequent statements until either commit() or rollback() is called on the connection. The
        read-only transaction will read from a consistent snapshot of the database at the time that the
        transaction started. This means that the transaction will not see any changes that have been
        committed by other transactions since the start of the read-only transaction. Commit or rolling back
        the read-only transaction is semantically the same, and only indicates that the read-only transaction
        should end a that a new one should be started when the next statement is executed.
```

google/cloud/spanner_dbapi/connection.py
```diff
@@ -121,6 +127,31 @@ def instance(self):
...
         return self._instance
 
+    @property
+    def read_only(self):
+        """Flag: the connection can be used only for database reads.
+
+        Returns:
+            bool:
+                True, if the connection intended to be used for
+                database reads only.
```
olavloite, line 137:
```suggestion
                True if the connection may only be used for database reads.
```

google/cloud/spanner_v1/transaction.py
```diff
@@ -80,9 +80,14 @@ def _make_txn_selector(self):
...
         return TransactionSelector(id=self._transaction_id)
 
-    def begin(self):
+    def begin(self, read_only=False):
         """Begin a transaction on the database.
 
+        :type read_only: bool
+        :param read_only:
+            (Optional) If True, ReadOnly transaction type will be
+            begun, ReadWrite otherwise.
```
olavloite, line 89:
```suggestion
            (Optional) If True, a read-only transaction with strong timestamp bound will be started.
            Otherwise a read/write transaction will be started.
```

google/cloud/spanner_v1/transaction.py
olavloite: Thanks, this looks great. I've added a couple of suggestion for documentation in the code.

Also, there is one more thing that we could simplify for read-only transactions: A read-only transaction will never be aborted by Cloud Spanner, so keeping track of a running checksum for the returned result sets is not necessary.  We could therefore skip that when we see that the transaction is in read-only mode.

google/cloud/spanner_dbapi/connection.py
```diff
@@ -177,7 +217,7 @@ def _rerun_previous_statements(self):
         for statement in self._statements:
             res_iter, retried_checksum = self.run_statement(statement, retried=True)
             # executing all the completed statements
-            if statement != self._statements[-1]:
+            if statement != self._statements[-1] and not self.read_only:
```
olavloite, line 220: I think it would be better to change the `except Aborted` blocks in the different methods so that they never call `retry_transaction` when it is a read-only transaction. Instead the `Aborted` error should just be propagated to the caller when it is a read-only transaction (and in theory it should also never happen).

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -290,7 +290,7 @@ def fetchall(self):
         res = []
         try:
             for row in self:
-                if not self.connection.autocommit:
+                if not self.connection.autocommit and not self.connection.read_only:
                     self._checksum.consume_result(row)
                 res.append(row)
         except Aborted:
```
olavloite, line 296: Add an `if` statement to this and other `except Aborted` blocks to prevent `retry_transaction` from being called when it is a read-only transaction.

larkee (changes requested):
Read-only transaction in the Python client are represented by the `Snapshot` [class](https://github.com/googleapis/python-spanner/blob/master/google/cloud/spanner_v1/snapshot.py#L532).

Please do not modify `transaction.py`. Use `Snapshot` instead.

IlyaFaer:
@larkee, hm, okay. If we want to use the `Snapshot()` class, we probably could re-use the existing method for single-use reads. Let me know if you think it's better to use a snapshot as a transaction for several sequential reads.

larkee:
> @larkee, hm, okay. If we want to use the `Snapshot()` class, we probably could re-use the existing method for single-use reads. Let me know if you think it's better to use a snapshot as a transaction for several sequential reads.

By default, a snapshot is a single-use read-only transaction so it would only be valid for one read operation. However, there is an option make a multi-use snapshot in which all reads are from the same snapshot of data i.e. data is consistent.

Multi-use snapshot example: https://cloud.google.com/spanner/docs/transactions#ro_transaction_example
Single-use snapshot example: https://cloud.google.com/spanner/docs/reads#single_read_methods

With this in mind, I think using `multi_use` snapshots should only be used when `read_only=True` and `AUTOCOMMIT=False`. WDYT?

olavloite:
> With this in mind, I think using `multi_use` snapshots should only be used when `read_only=True` and `AUTOCOMMIT=False`. WDYT?

Yes, that is the intention here. So the implementation should roughly be:

1. `read_only=True` and `auto_commit=False`: Use `multi_use` snapshots and only allow queries/reads. Calling `commit()` or `rollback()` should close the current multi-use snapshot, and the next statement that is executed should create a new multi-use snapshot.
2. `auto_commit=True` (and regardless of the state of `read_only`): Always use single-use snapshots for queries/reads.

larkee (commented):
LGTM, however I would like some system tests to confirm it is WAI

google/cloud/spanner_dbapi/connection.py
```diff
@@ -187,15 +229,16 @@ def _rerun_previous_statements(self):
                     self.connection._transaction = None
                     raise Aborted(status.details)
 
-                retried_checksum = ResultsChecksum()
-                retried_checksum.consume_result(res)
-                retried_checksum.consume_result(status.code)
+                if not self.read_only:
```
olavloite, line 232: This should not be necessary. The _rerun_previous_statements method should never be called when we are in read-only mode (and also not when in auto-commit mode).

Re-running a list of statements is only necessary when we are in a multi-use read/write transaction.

google/cloud/spanner_dbapi/connection.py
```diff
@@ -187,15 +229,16 @@ def _rerun_previous_statements(self):
...
+                    retried_checksum.consume_result(res)
+                    retried_checksum.consume_result(status.code)
 
-                _compare_checksums(checksum, retried_checksum)
+                    _compare_checksums(checksum, retried_checksum)
             else:
                 res_iter, retried_checksum = self.run_statement(statement, retried=True)
                 # executing all the completed statements
-                if statement != self._statements[-1]:
+                if statement != self._statements[-1] and not self.read_only:
```
olavloite, line 241: Here also: This should not be necessary. We should never call _rerun_previous_statements unless the transaction is a multi-use read/write transaction.

google/cloud/spanner_dbapi/connection.py
```diff
@@ -231,6 +268,19 @@ def transaction_checkout(self):
...
+
+        Initiate a new multi-use snapshot, if there is no snapshot in
+        this connection yet. Return the existing one otherwise.
+
+        :rtype: :class:`google.cloud.spanner_v1.snapshot.Snapshot`
+        :returns: A Cloud Spanner snapshot object, ready to use.
+        """
+        if not self._snapshot:
+            self._snapshot = Snapshot(self._session_checkout(), multi_use=True)
+        return self._snapshot
```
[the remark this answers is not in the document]
olavloite: I think it would be good in that case to add a check at the beginning of the method that actually checks that the state of the connection is `read_only==True && autocommit==False`. That prevents any accidental errors in the future when making changes.

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -394,38 +398,48 @@ def setoutputsize(self, size, column=None):
...
         self._raise_if_closed()
 
+    def _handle_DQL_with_snapshot(self, snapshot, sql, params):
+        # Reference
+        #  https://googleapis.dev/python/spanner/latest/session-api.html#google.cloud.spanner_v1.session.Session.execute_sql
+        sql, params = parse_utils.sql_pyformat_args_to_spanner(sql, params)
+        res = snapshot.execute_sql(
+            sql, params=params, param_types=get_param_types(params)
+        )
+        if type(res) == int:
```
olavloite, line 408: Doesn't this mean that the statement was an update statement? In that case, it should not be possible to ever enter this if-block, as an update statement cannot be executed on a read-only snapshot.
Or is this some specific handling for `SELECT COUNT(*)` statements?

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -394,38 +398,48 @@ def setoutputsize(self, size, column=None):
...
+            # many items returns a RuntimeError if .fetchone() is
+            # invoked and vice versa.
+            self._result_set = res
+            # Read the first element so that the StreamedResultSet can
+            # return the metadata after a DQL statement. See issue #155.
+            while True:
+                try:
+                    self._itr = PeekIterator(self._result_set)
+                    break
+                except Aborted:
```
olavloite, line 427: A read-only snapshot will never be aborted by Cloud Spanner, so this is not necessary.

google/cloud/spanner_dbapi/cursor.py
IlyaFaer: I don't think it's a relevant part of code at all, it was probably copy-pasted from somewhere else. `Snapshot.execute_sql()` returns `StreamedResultSet`, and at the first look I don't see how an `int` can get here.

Still, in this PR I wasn't going to change anything, I've only rewritten the method to make it able to pass a snapshot as an arg (not only use it as a context). So, there is a note for this `if` statement, but let's keep it for another PR.

google/cloud/spanner_dbapi/cursor.py
IlyaFaer: Yeah, it also looks like something copied from another part of code (where the `execute_sql()` method is called not with `Snapshot`). Will take a look.

google/cloud/spanner_dbapi/cursor.py
olavloite: This comment has been marked as resolved, but no changes were made. Do you disagree with my suggestion, or is this something you are still working on?

google/cloud/spanner_dbapi/connection.py
```diff
@@ -259,14 +314,18 @@ def commit(self):
 
         This method is non-operational in autocommit mode.
         """
+        self._snapshot = None
+
```
larkee, line 318: I think it would make more sense to add the `read_only` check here and exit early. If the connection is read only, then none of the code below should be relevant. WDYT?

This also applies to `rollback()`.

google/cloud/spanner_dbapi/connection.py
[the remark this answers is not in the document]
larkee: This was marked as resolved but no changes were made. Please address.

google/cloud/spanner_dbapi/connection.py
olavloite: The `self._release_session()`call would also be needed in the early exit branch in that case.

google/cloud/spanner_dbapi/connection.py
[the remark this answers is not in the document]
IlyaFaer: Done

skuruppu:
@IlyaFaer is this ready for another round of reviews?

IlyaFaer:
@skuruppu, yeah, I think everything is fixed. The only unresolved question was answered by Knut, so I assume it's also resolved.

olavloite:
@IlyaFaer 

There are a couple of comments that were marked as resolved, but with no action taken, such as:
* https://github.com/googleapis/python-spanner/pull/475/files#r689372738
* https://github.com/googleapis/python-spanner/pull/475/files#r684914316

Would you mind taking another look at those?

Also, @larkee asked for a system test to be added to show the working of the feature. Could you take a look at that as well?

IlyaFaer:
> There are a couple of comments that were marked as resolved, but with no action taken, such as:
https://github.com/googleapis/python-spanner/pull/475/files#r689372738
https://github.com/googleapis/python-spanner/pull/475/files#r684914316

I've written a comment for all the related comments: https://github.com/googleapis/python-spanner/pull/475#discussion_r689369342
So, I'm going to work on it, but in a separate PR, as it doesn't seem to be related to this one, but will require not the smallest changes.

The system test has been lost! It was actually added, but because of (not mine) big system tests refactoring PR it was erased during merge.
https://github.com/googleapis/python-spanner/pull/475/commits/b8eefed65d660dd7cd3cef27d5a61f5caf18edad

olavloite:
> I've written a comment for all the related comments: [#475 (comment)](https://github.com/googleapis/python-spanner/pull/475#discussion_r689369342)

That is true for https://github.com/googleapis/python-spanner/pull/475/files#r689320974, but not for the two other comments in this case which are directly related to this change:
1. This PR introduces support for multi-use read-only transactions.
2. The existing read/write transactions use a checksum and retry mechanism to handle aborted errors.
3. The newly introduced read-only transactions now also use the checksum and retry mechanism that was in place for read/write transactions.
4. Read-only transactions however will never be aborted by Cloud Spanner, which means that keeping track of a checksum and checking for aborted errors for read-only transactions is not necessary.

So I think this PR should take into account that read-only transactions will never be aborted by Cloud Spanner, they should not keep track of a result set checksum, and they should not be retried in case of an aborted error.

IlyaFaer:
@olavloite, the PR is adding ReadOnly transactions support, there are users, who wait for this feature. The error with `Aborted` expectation was always there - it's not optimal, yes, but lived with it and we can live little bit more, while lack of ReadOnly transactions support can block users' projects. As I see it, we should merge this PR ASAP (it's been 1.5 months in progress already), so that people at least could use the package. The error will be fixed anyway, it just should not delay more significant changes (and hence customers' projects).

skuruppu:
I would be a bit worried about doing the extra performance cost of keeping track of checksums for read-write transactions though. I wouldn't be comfortable releasing that. Is it hard to fix @IlyaFaer?

IlyaFaer:
@skuruppu, of course, it's not. I don't want to sound boring, but I can't see why you both think it's that performance influencing. We don't calculate checksums in `read_only` - the `if`s are already added (https://github.com/googleapis/python-spanner/pull/475/commits/8332d4977e658d7a47971932c1d19502bcb9d7fb). As for retrying `read_only`, well, `read_only` transactions don't fail with `Aborted`, so they're not gonna be retried anyway. Thus, I don't see why these changes should be urgent and delay other fixes...

Anyway it's pushed.

olavloite:
> ... but I can't see why you both think it's that performance influencing. We don't calculate checksums in read_only - the ifs are already added (8332d49).

Sorry, that was my fault. I didn't realize that you had already created that change, as the [comment that you referenced](https://github.com/googleapis/python-spanner/pull/475#discussion_r689369342) above didn't say anything about that.

> ... Thus, I don't see why these changes should be urgent and delay other fixes...

In my opinion, the current implementation would have been slightly confusing for a potential new developer. The retry mechanism is not needed for read-only transactions. I think that it is better that it is completely clear in the code to prevent anyone from accidentally re-enabling it in the future. (This last commit for example has also removed an unnecessary test case that tested the retry mechanism for a query in autocommit mode; a situation that could not happen, but still was part of the code base.)

Anyways, I'm happy enough with this as-is, but we should add a system test ASAP to actually show that it is working as intended. If @skuruppu and/or @larkee are OK with that, then we could do that in a separate PR.

skuruppu:
Ultimately @larkee needs to approve this as the code owner so please do add any tests that they ask for before merging.

larkee:
@IlyaFaer I second what @olavloite said. The refactor is less about performance (since it should be roughly equivalent) but is more about readability and reducing confusion for new developers.

However, if you create a tracking issue for the refactor work, I will approve and merge this PR :+1:

IlyaFaer:
I think all comments are already processed.

I've pushed changes with adding `if` statements into `fetch` methods to avoid retrying `read_only` transactions, etc. (see https://github.com/q-logic/python-spanner/commit/ac8c4b2b62cd228d687decda4ec54b32553164e4). So, I don't think any refactoring left here.

I also returned back that deleted-by-merge-of-another-commit system test. It checks that `UPDATE` operation will cause an error, while `SELECT` will work fine.

I think it's worth adding a unit test, which'll check that `read_only` transactions are not retried on `Aborted` exception. Otherwise, I think we don't have anything for a separate issue.

larkee (approved):

olavloite (approved):

IlyaFaer:
Looking at `automerge` label set by googler, I suppose it's not a problem if I'll merge it by myself (while tests are green and there are no changes in the `main` branch).

PR #479, opened 2021-08-06
tests: allow prerelease dependency versions under Python 3.9

tseaver:
Also, drop use of 'pytz', which is no longer depended on by `google-api-core` / `google-cloud-core`.

Instead, use either `datetime.timezone.utc` or `google.cloud._helpers.UTC`, depending on usage.


larkee (approved):

tseaver:
The error here is because we patch `time.time`, which ends up being called more than once because OT is emitting an unexpected warning message:

```python
google/cloud/spanner_v1/session.py:371: in run_in_transaction
    txn.commit(
google/cloud/spanner_v1/transaction.py:161: in commit
    with trace_call("CloudSpanner.Commit", self._session, trace_attributes):
/usr/local/lib/python3.9/contextlib.py:117: in __enter__
    return next(self.gen)
google/cloud/spanner_v1/_opentelemetry_tracing.py:55: in trace_call
    span.set_status(Status(StatusCode.OK))
.nox/unit-3-9/lib/python3.9/site-packages/opentelemetry/trace/status.py:57: in __init__
    logger.warning(
/usr/local/lib/python3.9/logging/__init__.py:1446: in warning
    self._log(WARNING, msg, args, **kwargs)
/usr/local/lib/python3.9/logging/__init__.py:1575: in _log
    record = self.makeRecord(self.name, level, fn, lno, msg, args,
/usr/local/lib/python3.9/logging/__init__.py:1544: in makeRecord
    rv = _logRecordFactory(name, level, fn, lno, msg, args, exc_info, func,
/usr/local/lib/python3.9/logging/__init__.py:283: in __init__
    ct = time.time()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

    def _time(_results=[1, 1.5]):
>       return _results.pop(0)
E       IndexError: pop from empty list

tests/unit/test_session.py:1192: IndexError
------------------------------ Captured log call -------------------------------
WARNING  opentelemetry.trace.status:status.py:57 description should only be set when status_code is set to StatusCode.ERROR
```

We can clearly work around this issue by making our `_time` mock an infinite generator. However,  I don't believe the traceback which triggered the error here.  This line:

```python
google/cloud/spanner_v1/_opentelemetry_tracing.py:55: in trace_call
    span.set_status(Status(StatusCode.OK))
```
clearly does *not* pass a non-default `description` to the `Status` constructor (which is what the warning is complaining about).

I wish there was a way for us to see the list of installed packages in the virtual environment.

PR #484, opened 2021-08-11
fix(samples): batch_update() results processing error

IlyaFaer:
The original `batch_update()` method [returns two values](https://github.com/googleapis/python-spanner/blob/5629bacf780cf79c3af4cda3de50986c82e18c0d/google/cloud/spanner_v1/transaction.py#L315-L321): batch status and row counts, while the sample accepts only one value from the method call (and checks its length, which is always equal to 2).

### DIFF
--- a/samples/samples/snippets.py
+++ b/samples/samples/snippets.py
@@ -1509,6 +1509,8 @@ def delete_data_with_partitioned_dml(instance_id, database_id):
 def update_with_batch_dml(instance_id, database_id):
     """Updates sample data in the database using Batch DML. """
     # [START spanner_dml_batch_update]
+    from google.rpc.code_pb2 import OK
+
     # instance_id = "your-spanner-instance"
     # database_id = "your-spanner-db-id"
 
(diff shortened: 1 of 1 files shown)


skuruppu (approved):

samples/samples/snippets.py
```diff
@@ -1529,7 +1529,11 @@ def update_with_batch_dml(instance_id, database_id):
     )
 
     def update_albums(transaction):
-        row_cts = transaction.batch_update([insert_statement, update_statement])
+        status, row_cts = transaction.batch_update([insert_statement, update_statement])
+
+        # check response status and process error (if occurred)
+        # if status.code != rpc.code_pb2.OK:
+        # ...
```
larkee, line 1536: I have concerns about this comment creating incorrect expectations. Yes, you can check the status here and do something different non-OK statuses. However, the error will still be raised by `run_in_transaction` when `commit()` is called unless you raise something else within the transaction logic.

larkee: Perhaps we should add the check in the sample to show best practices and include comments explaining the peculiarities, e.g.
```
if status.code != OK:
  # Do handling here.
  # Note: the error will still be raised when commit is called by `run_in_transaction`.
  return
```

WDYT, @IlyaFaer?

IlyaFaer: @larkee, yeah, it can be done

larkee (approved):

PR #543, opened 2021-08-27
fix(db_api): move connection validation into a separate method

IlyaFaer:
Fixes https://github.com/googleapis/python-spanner-sqlalchemy/issues/104


tests/unit/spanner_dbapi/test_connection.py
```diff
@@ -624,3 +624,46 @@ def test_retry_transaction_w_empty_response(self):
...
+        snapshot_ctx = mock.Mock()
+        snapshot_ctx.__enter__ = mock.Mock(return_value=snapshot_obj)
+        snapshot_ctx.__exit__ = exit_func
+        snapshot_method = mock.Mock(return_value=snapshot_ctx)
+
+        connection.database.snapshot = snapshot_method
+
+        connection.validate()
+
+    def test_validate_fail(self):
```
larkee, line 648: This test only checks that an error is raised if the backend returns an incorrect result. There is no test for an invalid instance or database, or if the connection is closed.

google/cloud/spanner_dbapi/connection.py
```diff
@@ -349,6 +349,24 @@ def run_statement(self, statement, retried=False):
...
+        invalid, target database is not found, or the request result
+        is incorrect.
+
+        :raises: :class:`InterfaceError`: if this connection is closed.
+        :raises: :class:`OperationalError`: if the request result is incorrect.
+        """
+        self._raise_if_closed()
+
+        with self.database.snapshot() as snapshot:
+            if [[1]] != list(snapshot.execute_sql("SELECT 1")):
```
larkee, line 367: nit: prefer `list(snapshot.execute_sql("SELECT 1")) != [[1]]`

google/cloud/spanner_dbapi/connection.py
```diff
@@ -349,6 +349,24 @@ def run_statement(self, statement, retried=False):
...
+                raise OperationalError("The connection is invalid")
```
larkee, line 368: This error message could be more helpful; currently there is no indication of why the connection is invalid.

larkee (changes requested):
This PR will not behave as intended against production. Please add system tests.

tests/unit/spanner_dbapi/test_connection.py
IlyaFaer: I agree on adding a test for a closed connection, but do we need tests for an invalid instance/database? It'll be an error from deeper `execute_sql()` call, and we do nothing with this error. Meaning it doesn't seem to be strong related to the `ping()` method.

larkee: I'd prefer us to know exactly what the error surfaced to the user looks like so we can be sure it's clear to the user what went wrong.

IlyaFaer: Well, we don't do anything with the exception. If it's raised, it's raised.
Anyway, here is the test.

google/cloud/spanner_dbapi/connection.py
```diff
@@ -349,6 +349,28 @@ def run_statement(self, statement, retried=False):
...
+        """
+        Execute a minimal request to check if the connection
+        is valid and the related database is reachable.
+
+        Raise an exception in case if the connection is closed,
+        invalid, target database is not found, or the request result
+        is incorrect.
+
+        :raises: :class:`InterfaceError`: if this connection is closed.
+        :raises: :class:`OperationalError`: if the request result is incorrect.
```
larkee, line 362: Please also include `NotFound` if instance or database are non-existent. The previous paragraph mentions that an exception is raised in this case so it's good to explicitly state what this exception will be.

larkee (approved):
LGTM with the docstring addition

google/cloud/spanner_dbapi/connection.py
IlyaFaer: Done

PR #572, opened 2021-09-16
perf: use batch_update instead of sending a request for each statement

larkee:
This PR will improve the performance for queries that generate multiple statements by sending them together rather than one-by-one.


larkee:
Closed in favor of #630

PR #584, opened 2021-09-21
feat(db_api): support stale reads

IlyaFaer:
Based on #475 and requires it to be merged first.


google/cloud/spanner_dbapi/connection.py
```diff
@@ -165,6 +166,48 @@ def read_only(self, value):
...
         self._read_only = value
 
+    @property
+    def staleness(self):
+        """Current read staleness option value of this `Connection`.
+
+        Returns:
+            dict: Staleness type and value.
+        """
+        return self._staleness
```
larkee, line 176: In the interest of avoiding having to add `or {}` to a bunch of checks, perhaps it makes more sense to do one of the following:

1) Keep the default for `self._staleness` as `None` and update this line to `return self._staleness or {}`
2) Change the default for `self._staleness` to `{}` and update this method to return `None` if `self._staleness is {}`

Which one we use depends on whether getting `None` or an empty dict is more useful/clearer when calling `conn.staleness`

tests/unit/spanner_dbapi/test_connection.py
```diff
@@ -763,3 +755,121 @@ def test_validate_closed(self):
...
+        with self.assertRaises(ValueError):
+            connection.staleness = {"something": 4}
+
+        # no expected staleness types
+        with self.assertRaises(ValueError):
+            connection.staleness = {}
+
+    def test_staleness_inside_transaction(self):
+        """
+        Check that it's impossible to change the `steleness`
```
larkee, line 777:
```suggestion
        Check that it's impossible to change the `staleness`
```

tests/unit/spanner_dbapi/test_connection.py
```diff
@@ -763,3 +755,121 @@ def test_validate_closed(self):
...
+        """
+        connection = self._make_connection()
+        connection._transaction = mock.Mock(committed=False, rolled_back=False)
+
+        with self.assertRaises(ValueError):
+            connection.staleness = {"read_timestamp": datetime.datetime(2021, 9, 21)}
+
+    def test_staleness_multi_use(self):
+        """
+        Check that `steleness` option is correctly
```
larkee, line 788:
```suggestion
        Check that `staleness` option is correctly
```

google/cloud/spanner_dbapi/connection.py
```diff
@@ -165,6 +166,48 @@ def read_only(self, value):
...
+                "read_timestamp",
+                "min_read_timestamp",
+                "max_staleness",
+                "exact_staleness",
+            ):
+                if opt in value and value[opt]:
+                    one_of_expected = True
+                    break
+
+            if not one_of_expected:
```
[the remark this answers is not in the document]
IlyaFaer: Done

larkee (approved):

skuruppu:
@IlyaFaer what do we need to do to get this merged? Are the test failures related to this change or are they flaky?

IlyaFaer:
@skuruppu, looks like there was a problem with lint (I've updated the `black` linting package locally, but kokoro seems to be using a particular version). Fixed it, let's see if everything is okay now.

PR #593, opened 2021-09-27
fix: add support for json data type

### LINKED ISSUE #592: bug: json data insert int db is failing
json data insert int db is failing

#### Environment details

  - OS type and version: Mac os latest
  - Python version: `python --version` 3.8
  - `google-cloud-spanner` 3.10

#### Steps to reproduce

  1. Insert json data in db using `cursor.execute` method
  2. Pass the json as dictionary without formatting it as string.

#### Code example

```python
    # Create table
    cur = conn.cursor()
    cur.execute(
        """
        CREATE TABLE JsonDetails (
            DataId     INT64 NOT NULL,
            Details    JSON,
        ) PRIMARY KEY (DataId)
    """
    )
    conn.close()

    # Insert data to table
    conn = Connection(shared_instance, dbapi_database)
    conn.autocommit = True
    cur = conn.cursor()
    cur.execute(
        sql="INSERT INTO JsonDetails (DataId, Details) VALUES (%s, %s)",
        args=(123, {"name": "Jakob", "age": "26"}),
    )

```

#### Stack trace
```
except InvalidArgument as e:
>           raise ProgrammingError(e.details if hasattr(e, "details") else e)
E           django.db.utils.ProgrammingError: 400 Value of type STRING cannot be assigned to value, which has type JSON [at 1:33]\nUPDATE tests_detail SET value = @a0 WHERE tests_detail.id = @a1\n
```

Thanks!

vi3k6i5:
Add support for json datatype, but adding mapping from dictionary to json for db insert.

Fixes #592 🦕


asthamohta (approved):
LGTM

larkee (commented):
Could you link to evidence that DB-API expects JSON to be passed in as a dict?

This code LGTM but I'm cautious about using dict unless there is precedent

vi3k6i5:
> Could you link to evidence that DB-API expects JSON to be passed in as a dict?
> 
> This code LGTM but I'm cautious about using dict unless there is precedent

If we pass a string from django, then type casting will be to "STRING", which causes this error. DB-api needs to know that json has to be passed, but that conversion happens based on instance type.

```
except InvalidArgument as e:
>           raise ProgrammingError(e.details if hasattr(e, "details") else e)
E           django.db.utils.ProgrammingError: 400 Value of type STRING cannot be assigned to value, which has type JSON [at 1:33]\nUPDATE tests_detail SET value = @a0 WHERE tests_detail.id = @a1\n
```

larkee:
> > Could you link to evidence that DB-API expects JSON to be passed in as a dict?
> > This code LGTM but I'm cautious about using dict unless there is precedent
> 
> If we pass a string from django, then type casting will be to "STRING", which causes this error. DB-api needs to know that json has to be passed, but that conversion happens based on instance type.

Hmm.. I see your point. Perhaps we should follow the route of [`pyscopg`](https://www.psycopg.org/docs/extras.html?highlight=json#json-adaptation) and add a custom type class `Json`. The benefit of doing so is we could allow JSON values to be specified as either an object or a string:
```
# Use object for JSON.
cur.execute(
        sql="INSERT INTO JsonDetails (DataId, Details) VALUES (%s, %s)",
        args=(123, Json({"name": "Jakob", "age": "26"})),
    )

# Use string for JSON.
cur.execute(
        sql="INSERT INTO JsonDetails (DataId, Details) VALUES (%s, %s)",
        args=(123, Json('{"age":"26","name":"Jakob"}')),
    )
```
WDYT?

vi3k6i5:
> > > Could you link to evidence that DB-API expects JSON to be passed in as a dict?
> > > This code LGTM but I'm cautious about using dict unless there is precedent
> > 
> > 
> > If we pass a string from django, then type casting will be to "STRING", which causes this error. DB-api needs to know that json has to be passed, but that conversion happens based on instance type.
> 
> Hmm.. I see your point. Perhaps we should follow the route of [`pyscopg`](https://www.psycopg.org/docs/extras.html?highlight=json#json-adaptation) and add a custom type class `Json`. The benefit of doing so is we could allow JSON values to be specified as either an object or a string:
> 
> ```
> # Use object for JSON.
> cur.execute(
>         sql="INSERT INTO JsonDetails (DataId, Details) VALUES (%s, %s)",
>         args=(123, Json({"name": "Jakob", "age": "26"})),
>     )
> 
> # Use string for JSON.
> cur.execute(
>         sql="INSERT INTO JsonDetails (DataId, Details) VALUES (%s, %s)",
>         args=(123, Json('{"age":"26","name":"Jakob"}')),
>     )
> ```
> 
> WDYT?

This makes sense, but then users would be tied to using our data type in their code. 

The mapping would still have to be added
```json_type: spanner.param_types.JSON,```

and the conversion from `json` to `json_dump`
```
    if isinstance(value, json_type):
        return Value(
            string_value=json.dumps(value, sort_keys=True, separators=(",", ":"),)
        )
```

**Comment by IlyaFaer** [comment]
Apologies for bursting in without an invitation, but it should be mentioned we support such a params style:
https://github.com/googleapis/python-spanner/blob/f59d08bf5562e52fbe1535047ddafaa399814b3e/tests/unit/spanner_dbapi/test_parse_utils.py#L276-L279

Also there is a request to support `pyformat` formatting style, which'll also use `dict`:
https://www.python.org/dev/peps/pep-0249/#paramstyle

Considering it, using `dict` as a JSON data type can be little bit confusing. I second @larkee proposal to add a separate JSON data type class.

**Comment by vi3k6i5** [comment]
> Apologies for bursting in without an invitation, but it should be mentioned we support such a params style:
> https://github.com/googleapis/python-spanner/blob/f59d08bf5562e52fbe1535047ddafaa399814b3e/tests/unit/spanner_dbapi/test_parse_utils.py#L276-L279
> 
> Also there is a request to support `pyformat` formatting style, which'll also use `dict`:
> https://www.python.org/dev/peps/pep-0249/#paramstyle
> 
> Considering it, using `dict` as a JSON data type can be little bit confusing. I second @larkee proposal to add a separate JSON data type class.

Makes sense, will add a separate JSON type for the same.

**Comment by vi3k6i5** [comment]
@larkee @IlyaFaer Not able to find the right place to define the json type, any suggestions ?

**Comment by IlyaFaer** [comment]
I guess it should live somewhere in https://github.com/googleapis/python-spanner/tree/main/google/cloud/spanner_v1/types, as a  separate file (because in other files there are protos).

Also there is a [types.py](https://github.com/googleapis/python-spanner/blob/main/google/cloud/spanner_dbapi/types.py) file for DB API, but writing a class into this file looks like it's going to be used only by the DB API.

**Comment by vi3k6i5** [comment]
@IlyaFaer : Thank you for the suggestion, added a new object called JsonObject and moved it to types.

**Review by asthamohta — APPROVED** [review]

File: google/cloud/spanner_v1/types/data_types.py [hunk 1]
```diff
@@ -0,0 +1,25 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Custom data types for spanner."""
+
+
+class JsonObject(dict):
+    """
+    JsonObject type help format Django jsonfield to compatible Cloud Spanner's
```
IlyaFaer, line 20: `jsonfield` - looks like a typo

tests/system/test_dbapi.py
```diff
@@ -328,6 +328,53 @@ def test_DDL_autocommit(shared_instance, dbapi_database):
...
+    cur = conn.cursor()
+    cur.execute(
+        """
+        CREATE TABLE JsonDetails (
+            DataId     INT64 NOT NULL,
+            Details    JSON,
+        ) PRIMARY KEY (DataId)
+    """
+    )
+    conn.close()
```
IlyaFaer, line 348: You probably don't need to `close()` connection and open new for every operation. It should be okay to use single `Connection` and `Cursor` objects for the whole test.

google/cloud/spanner_v1/types/data_types.py
vi3k6i5: jsonfield is a datatype in django, corrected the case for the same.

tests/system/test_dbapi.py
vi3k6i5: removed extra connection creation and reused the 1st connection.

larkee:
> I guess it should live somewhere in https://github.com/googleapis/python-spanner/tree/main/google/cloud/spanner_v1/types, as a separate file (because in other files there are protos).

Ahh, unfortunately, that is a generated directory which means we shouldn't put handwritten code there as it will be overwritten/removed whenever the code is regenerated.

I think the best option is to move `data_types.py` to `google/cloud/spanner_v1` and add `JsonObject` to the `google/cloud/spanner_v1/__init__` file so it can be easily imported e.g.
```
from google.cloud.spanner_v1 import JsonObject
```

vi3k6i5:
@larkee : Done, Thanks.

larkee (approved):

PR #594, opened 2021-09-27
chore: exclude 'CODEOWNERS' from templated files

tseaver:
See: https://github.com/googleapis/synthtool/pull/1201


parthea:
@tseaver  I'm going to close this PR because we've got a compatible change in https://github.com/googleapis/python-spanner/pull/642.

PR #603, opened 2021-09-30
feat(db_api): make rowcount property NotImplemented

IlyaFaer:
The original issue: https://github.com/googleapis/python-spanner-sqlalchemy/issues/121
As we're not able to implement `rowcount` property according to the PEP-249, erasing its functionality (which is actual for only single one case - UPDATE query in `autocommit` mode).


larkee (approved):

vi3k6i5:
Updates with autocommit is default in Django, which breaks with these changes, can we undo these changes? @IlyaFaer

IlyaFaer:
@vi3k6i5, these changes were changed not very long ago, here #628

vi3k6i5:
The fix in #628 is not working for Django. In Django model object update in auto-commit mode checks for success by getting the `_row_count` value which is now returning -1 and thus Django thinks that the update failed.

vi3k6i5:
Not able to run updates because of this change:
`
author_kent = Author(
    first_name="Arthur",
    last_name="Kent",
    rating=Decimal("4.1"),
)
author_kent.save() # Works file
author_kent.rating = Decimal("4.2")
author_kent.save() # Fails with following error
`

Error
`
E           grpc._channel._InactiveRpcError: <_InactiveRpcError of RPC that terminated with:
E           	status = StatusCode.ALREADY_EXISTS
E           	details = "Failed to insert row with primary key ({pk#id:837916644807504432}) due to previously existing row"
E           	debug_error_string = "{"created":"@1643794129.022745000","description":"Error received from peer ipv4:0.0.0.0:9010","file":"src/core/lib/surface/call.cc","file_line":1075,"grpc_message":"Failed to insert row with primary key ({pk#id:837916644807504432}) due to previously existing row","grpc_status":6}"
E           >

../../.virtualenvs/temp_2/lib/python3.9/site-packages/grpc/_channel.py:849: _InactiveRpcError
`


Reason for the failure is pretty simple:
1. During `author_kent.save()` Django tries to first update the value always, if that is successful then the query is ended else it tries to insert.
2. The way django checks if a update was successful or not is by checking row_count > 0 [Link](https://github.com/django/django/blob/main/django/db/models/sql/compiler.py#L1612)

Since we are returning -1 for row_count, django assumes the update has failed and tries to insert data again which fails because there is already an entry in db with the same primary key.

Solution would be to revert the row_count changes.

Can you elaborate where the row_count returns incorrect result which does not match the expectation set in [PEP 249](https://www.python.org/dev/peps/pep-0249/#rowcount) 

Specifically [point 7](https://www.python.org/dev/peps/pep-0249/#id46) and [point 9](https://www.python.org/dev/peps/pep-0249/#id48). 

Thanks

PR #640, opened 2021-11-11
chore: changing default region to us-west1

asthamohta:


larkee (approved):

larkee (changes requested):
The backup tests also use an additional instance with a different config:
https://github.com/googleapis/python-spanner/blob/8ca868c3b3f487c1ef4f655aedd0ac2ca449c103/tests/system/test_backup_api.py#L56
We should probably prevent that one from choosing `us-central1` for now as well.

We should also make this change for the samples:
https://github.com/googleapis/python-spanner/blob/main/samples/samples/conftest.py#L90

vi3k6i5 (approved):
LGTM

larkee (approved):

Issue #625, opened 2021-10-12
Add a py.typed file for submodule


crwilcox:
To be in accordance with [PEP 561](https://www.python.org/dev/peps/pep-0561/) we should add a py.typed file to google/cloud/datastore/py.typed

The file can be empty but does need to be present for some type checkers.

Also, for this to work, google/cloud/spanner.py needs to be restructured to google/cloud/spanner/__init__.py

See https://github.com/googleapis/python-firestore/issues/447 for additional context

PR #646, opened 2021-11-26
feat: python typing

IlyaFaer:
Towards #625 

`py.typed` file is already added into the repository and is included in the package. Implementing a new `nox` session to check types with `mypy`.


google/cloud/spanner_v1/__init__.py
```diff
@@ -18,7 +18,7 @@
 
 import pkg_resources
 
-__version__ = pkg_resources.get_distribution("google-cloud-spanner").version
+__version__: str = pkg_resources.get_distribution("google-cloud-spanner").version
```
tseaver, line 21: This annotation shouldn't be needed, since you have installed `types-setuptools` in the nox session.

tseaver (changes requested):
Ideally we would be testing `google/` and `tests/`, and without the `--ignore-missing-imports` flag (the `--no-incremental` flat shouldn't be needed, either -- one can just remove `.mypy_cache` to blow the cache away).  That would likely require adding some more `types-` packages (e.g. `types-protobuf`, `types-mock`).

We would then be adding `py.typed` to the remaining subdirs of `google/cloud/`, in particular, to `google/cloud/spanner_dbapi/`.

asthamohta (approved):

PR #668, opened 2022-01-20
chore: update supported python versions in README to >= 3.6

rajatbhatta:
Update supported python versions in README to >= 3.6, in line with setup.py which indicates python_requires=">=3.6".

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -56,7 +56,7 @@ dependencies.
 
 Supported Python Versions
 ^^^^^^^^^^^^^^^^^^^^^^^^^
-Python >= 3.5
+Python >= 3.6
 
 Deprecated Python Versions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/README.rst
+++ b/README.rst
@@ -60,7 +60,8 @@ Python >= 3.6
 
 Deprecated Python Versions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
-Python == 2.7. Python 2.7 support will be removed on January 1, 2020.
+Python == 2.7.
+Python == 3.5.
 
 
 Mac/Linux


README.rst
```diff
@@ -56,7 +56,7 @@ dependencies.
 
 Supported Python Versions
 ^^^^^^^^^^^^^^^^^^^^^^^^^
-Python >= 3.5
+Python >= 3.6
```
ansh0l, line 59: Would this mean all customers using python 3.5 face incompatibility with library?

rajatbhatta: I think so. 

Support for Python 3.5 and earlier versions was removed as part of this commit: https://github.com/googleapis/python-spanner/pull/147/commits/606258bc728e5eed2b0b032a6428e1b77c39b3ce.

Adding @vi3k6i5 to comment/confirm the same.

IlyaFaer: Users will get problems if they'll try to upgrade to the new version of the `python-spanner` package. See the setup param for details:
https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#python-requires

So, until they'll try to upgrade, they're okay.
Plus to this, I don't really think it's a big move to upgrade from 3.5 to 3.6. It's not jumping from 2 to 3, it's a comparatively small step, which, most likely, will not require manual work.

vi3k6i5: Makes sense. This is anyways just a documentation fix, the setup.py file was updated a while back.

vi3k6i5: Maybe we should add Python 3.5 to Deprecated Python Versions below ? @ansh0l @IlyaFaer ?

vi3k6i5 (approved):
LGTM

README.rst
IlyaFaer: Yeah, I guess, let's add

rajatbhatta: This is done.

ansh0l: sounds good

ansh0l (approved):
lgtm

PR #694, opened 2022-03-28
fix: add session.exists() check in PingingPool.get() method.

vi3k6i5:
`PingingPool` currently checks if session exists in `.get()` method only when `_NOW() > ping_after`.  If the session is used in any query and then returned the pool then the ping_after is reset to `_NOW() _  + delta`. If the session is deleted in the backend and the checkout gives a not found error the session will still get returned to the pool with a reset for `ping_after = _NOW() _  + delta`.

As described in the comment in `get()` method.

> Using session.exists() guarantees the returned session exists.
> session.ping() uses a cached result in the backend which could
> result in a recently deleted session being returned.

Example:
An Application using `PingingPool` with 1 Session. The session has been killed on the server side, but the `ping` method has not cleared out the session from the pool yet because [ping_after > _NOW()](https://github.com/googleapis/python-spanner/blob/main/google/cloud/spanner_v1/pool.py#L435) is not `True`.

Now `pool.get()` will return the session, and [SnapshotCheckout](https://github.com/googleapis/python-spanner/blob/828df62d1e14dd83003897c5b9ad58592fdff852/google/cloud/spanner_v1/database.py#L858) will use the session and fail and return the session back into the [pool](https://github.com/googleapis/python-spanner/blob/828df62d1e14dd83003897c5b9ad58592fdff852/google/cloud/spanner_v1/database.py#L865).

Put Method will put the session in the pool with a new [wait_time](https://github.com/googleapis/python-spanner/blob/main/google/cloud/spanner_v1/pool.py#L402), and the process will continue. 

To avoid this inside the `__exit__` method of `SessionCheckout` we should check if `NOT_FOUND` error was raised or not. If that error was raised then create a new session and push it inside the pool.

### DIFF
--- a/google/cloud/spanner_v1/pool.py
+++ b/google/cloud/spanner_v1/pool.py
@@ -387,15 +387,14 @@ class PingingPool(AbstractSessionPool):
         if timeout is None:
             timeout = self.default_timeout
 
-        ping_after, session = self._sessions.get(block=True, timeout=timeout)
+        _, session = self._sessions.get(block=True, timeout=timeout)
 
-        if _NOW() > ping_after:
-            # Using session.exists() guarantees the returned session exists.
-            # session.ping() uses a cached result in the backend which could
-            # result in a recently deleted session being returned.
-            if not session.exists():
-                session = self._new_session()
-                session.create()
+        # Using session.exists() guarantees the returned session exists.
+        # session.ping() uses a cached result in the backend which could
+        # result in a recently deleted session being returned.
+        if not session.exists():
+            session = self._new_session()
+            session.create()
 
         return session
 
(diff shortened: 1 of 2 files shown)


asthamohta (approved):

larkee (changes requested):
This change means that every session will be checked before use which sounds great but the added latency defeats the purpose of the pinging pool and will slow down existing applications.

I think having logic for catching the exception in `SessionCheckout` might make more sense.

beaverlyhills:
I agree that handling exceptions in the `__exit__` blocks of all session checkout context managers is the way to go.

I would also suggest to not update the `wait_time` in `put()` method as there is no guarantee that any queries were actually performed by the caller. I wonder if that was the root cause of the issue we observed in our system, rather than Spanner server forcefully closing the session.

Callers would be responsible for handling `NotFound` exception and doing the retry with another session, which is something worth mentioning in the documentation of `PingingPool` as it does not guarantee that returned sessions are always valid.

vi3k6i5:
@beaverlyhills @larkee : Created a separate PR with the relevant changes: https://github.com/googleapis/python-spanner/pull/718 For some reason this PR was not getting updated.

PR #743, opened 2022-06-07
feat: FGAC changes with Samples

asthamohta:
This PR contains generated code as well, please review only the handwritten layer and samples


google/cloud/spanner_v1/pool.py
```diff
@@ -472,11 +503,22 @@ class TransactionPingingPool(PingingPool):
...
     """
 
-    def __init__(self, size=10, default_timeout=10, ping_interval=3000, labels=None):
+    def __init__(
+        self,
+        size=10,
+        default_timeout=10,
+        ping_interval=3000,
+        labels=None,
+        database_role=None,
```
IlyaFaer, line 512: The `database_role` argument docstrings should be added everywhere the argument was added

google/cloud/spanner_v1/database.py
```diff
@@ -559,16 +576,23 @@ def execute_pdml():
...
+        :param database_role: (Optional) user-assigned database_role for the session.
+
         :rtype: :class:`~google.cloud.spanner_v1.session.Session`
         :returns: a session bound to this database.
         """
-        return Session(self, labels=labels)
+        role = self._database_role
+        # If role is specified then that role is used instead.
+        if database_role:
+            role = database_role
```
rajatbhatta, line 594:
```suggestion
        # If role is specified in param, then that role is used
        # instead.
        role = database_role or self._database_role
```

samples/samples/snippets.py
```diff
@@ -349,23 +351,23 @@ def insert_data(instance_id, database_id):
...
-                (1, 1, u"Total Junk"),
-                (1, 2, u"Go, Go, Go"),
-                (2, 1, u"Green"),
-                (2, 2, u"Forever Hold Your Peace"),
-                (2, 3, u"Terrified"),
+                (1, 1, "Total Junk"),
+                (1, 2, "Go, Go, Go"),
+                (2, 1, "Green"),
+                (2, 2, "Forever Hold Your Peace"),
+                (2, 3, "Terrified"),
```
rajatbhatta, line 370: Why were these changed from unicode strings to normal strings? Does having unicode strings affect the newly written samples in any way?

tests/system/test_database_api.py
```diff
@@ -159,6 +161,48 @@ def test_create_database_with_default_leader_success(
...
+    databases_to_delete.append(temp_db)
+    create_op.result(DBAPI_OPERATION_TIMEOUT)
+    policy = temp_db.get_iam_policy(3)
+
+    assert policy.version == 0
+    assert policy.etag == b"\x00 \x01"
+
+    new_binding = policy_pb2.Binding(
+        role="roles/spanner.fineGrainedAccessUser",
+        members=["user:asthamohta@google.com"],
```
rajatbhatta, line 190: Should we explicit mention human users ("user:asthamohta@google.com")? Maybe we should use a non-human user?

rajatbhatta (changes requested):

samples/samples/snippets_test.py
```diff
@@ -653,3 +657,37 @@ def test_set_request_tag(capsys, instance_id, sample_database):
...
+
+@pytest.mark.dependency(depends=["add_and_drop_database_roles"])
+def test_list_database_roles(capsys, instance_id, sample_database):
+    snippets.list_database_roles(instance_id, sample_database.database_id)
+    out, _ = capsys.readouterr()
+    assert "new_parent" in out
+
+
+@pytest.mark.dependency(depends=["add_and_drop_database_roles"])
+def test_lenable_fine_grained_access(capsys, instance_id, sample_database):
```
asthamohta, line 685: fix type

samples/samples/snippets.py
asthamohta: This was automatically done by the linting application as per a new version

tests/system/test_database_api.py
asthamohta: Yes we plan on changing this for GA to a value that user can set up

PR #759, opened 2022-06-29
test: postgresql tests fix

asthamohta:
The following runs PG tests in a parametrised manner and fixes the errors in the tests


noxfile.py
```diff
@@ -269,6 +270,10 @@ def system(session):
 
     install_systemtest_dependencies(session, "-c", constraints_path)
 
+    # If POSTGRESQL tests and Emulator, skip the tests
+    if os.environ.get("SPANNER_EMULATOR_HOST") and database_dialect == "POSTGRESQL":
+        session.skip("Postgresql is not supported by Emulator yet.")
```
IlyaFaer, line 275: This `if` can be moved upper to avoid checking folders in case of PostgreSQL and emulator. Not a significant thing though

tests/system/conftest.py
```diff
@@ -57,6 +57,14 @@ def not_postgres(database_dialect):
         )
 
 
+@pytest.fixture(scope="session")
+def not_google_standard_sql(database_dialect):
+    if database_dialect == DatabaseDialect.GOOGLE_STANDARD_SQL:
+        pytest.skip(
+            f"{_helpers.DATABASE_DIALECT_ENVVAR} set to GOOGLE_STANDARD_SQL in environment."
+        )
```
IlyaFaer, line 65: I'm not 100% sure about this fixture. It's for a session, but if I'm not mistaken, it still needs to be used by a test to be actually executed. Like `not_postgres()` fixture, for example - it's the same, but it's called by several tests (by passing the fixture name as an argument of a test). But this one seems to be not used anywhere, thus, never executed.

.kokoro/build.sh
```diff
@@ -55,6 +55,8 @@ fi
 
 # If NOX_SESSION is set, it only runs the specified session,
 # otherwise run all the sessions.
+if [[ -n "${SPANNER_DATABASE_DIALECT:-}" ]]; then
+    export SPANNER_DATABASE_DIALECT = ${SPANNER_DATABASE_DIALECT:-}
```
IlyaFaer, line 59: `fi` should be added after this line to close the `if` statement

tests/system/conftest.py
asthamohta: Hey Ilya, so i will be using this when I introduce jsonb functionality which is only supported in postgres

ansh0l (approved):
LGTM

PR #771, opened 2022-07-15
fix: if JsonObject serialized to None then return `null_value` instead of `string_value`

o-aleks:


asthamohta:
hi @o-aleks, what is this PR for?

o-aleks:
Hi @asthamohta, 

There is a bug in `_make_value_pb` function. 
If the value is an instance of JsonObject and it is None after serializing, then it should be translated to NULL_VALUE, but in the original code, it is a string.

My fix is:
```
def _make_value_pb(value):
	...
    if isinstance(value, JsonObject):
        value = value.serialize()
        if value is None:
            return Value(null_value="NULL_VALUE")
        else:
            return Value(string_value=value)
```

Could you please check it and approve the PR?
Thank you.

IlyaFaer:
Should be okay. There is an `if` statement upper in the method, which does the same for all the data types, but `JsonObject` is more complicated than other types, so it passes the condition (and we can't change this behavior for the class, because there is no magic method for `is None` statements).

A test should be added though. @o-aleks, will you find time to add a test? Or I can do it in a separate issue.

o-aleks:
> A test should be added though. @o-aleks, will you find time to add a test? Or I can do it in a separate issue.
Oh, yes, of course, I'll add the test.

IlyaFaer (approved):

IlyaFaer:
Thanks, @o-aleks, looks good, and tests are passing fine. I'm merging it.

PR #794, opened 2022-08-30
samples: changes to json samples updating for JsonObject and linting

asthamohta:


samples/samples/noxfile.py
```diff
@@ -180,6 +180,7 @@ def blacken(session: nox.sessions.Session) -> None:
 # format = isort + black
 #
 
+
```
IlyaFaer, line 183: Noxfile should not be updated manually (or by linters), because it's autogenerated. There is a note at the top of the file. The only exception is noxfile in `sqlalchemy_spanner`, because we written it by ourselves to create very special test sessions.

samples/samples/snippets.py
```diff
@@ -1070,17 +1068,17 @@ def update_data_with_json(instance_id, database_id):
...
                         [
-                            {"name": "room 1", "open": True},
-                            {"name": "room 2", "open": False},
+                            JsonObject({"name": "room 1", "open": True}),
+                            JsonObject({"name": "room 2", "open": False}),
                         ]
                     ),
                 ),
-                (19, json.dumps({"rating": 9, "open": True})),
+                (19, JsonObject({"rating": 9, "open": True})),
```
IlyaFaer, line 1078: I think one of the `JsonObject`s here can be written in another style, like:
```python
JsonObject(rating=9, open=True)
```
With a comment that this style of arguments is also supported - some users may consider it more convenient.

IlyaFaer (approved):

rajatbhatta (approved):

PR #804, opened 2022-09-12
feat: Adding reason, domain, metadata & error_details fields in Custom Exceptions for additional info

gauravpurohit06:
Adding following properties under DBAPI exceptions to provide additional error info.
-reason
-domain
-metadata
-details

### DIFF
--- a/google/cloud/spanner_dbapi/exceptions.py
+++ b/google/cloud/spanner_dbapi/exceptions.py
@@ -36,7 +36,7 @@ class Error(Exception):
     def reason(self):
         """The reason of the error.
         Reference:
-            https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto#L112
+            https://cloud.google.com/apis/design/errors#error_info
         Returns:
             Union[str, None]: An optional string containing reason of the error.
         """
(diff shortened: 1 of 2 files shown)


google/cloud/spanner_dbapi/exceptions.py
```diff
@@ -27,7 +29,65 @@ class Error(Exception):
     Does not include :class:`Warning`.
     """
 
-    pass
+    def _is_error_cause_instance_of_google_api_exception(self):
```
IlyaFaer, line 32: It probably should be done in `__init__`. Cause of an error doesn't change, so it's a static property - better designate it once and keep as a `bool` attribute than check every time.

google/cloud/spanner_dbapi/exceptions.py
```diff
@@ -27,7 +29,65 @@ class Error(Exception):
...
+        return isinstance(self.__cause__, GoogleAPICallError)
+
+    @property
+    def reason(self):
+        """The reason of the error.
+        Reference:
+            https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto#L112
```
asthamohta, line 39: We normally don't refer to github proto links in reference. If there are any google documentation pages, you can add those instead.

google/cloud/spanner_dbapi/exceptions.py
gauravpurohit06: We can't do it under the `__init__ ` method, because `__cause__` gets set after the object is created. 
so `raise EXCEPTION from CAUSE) is equivalent to 
`exc = EXCEPTION
exc.__cause__ = CAUSE
raise exc`

google/cloud/spanner_dbapi/exceptions.py
gauravpurohit06: Yes, we can replace it with following document: 
https://cloud.google.com/apis/design/errors#error_info

asthamohta (approved):

PR #838, opened 2022-10-07
feat: support request options in !autocommit mode

IlyaFaer:
Towards https://github.com/googleapis/python-spanner-sqlalchemy/issues/248

### DIFF
--- a/google/cloud/spanner_dbapi/connection.py
+++ b/google/cloud/spanner_dbapi/connection.py
@@ -184,6 +184,21 @@ class Connection:
         self._read_only = value
 
     @property
+    def request_options(self):
+        """Options for the next SQL operations.
+
+        Returns:
+            google.cloud.spanner_v1.RequestOptions:
+                Request options.
+        """
+        if self.request_priority is None:
+            return
+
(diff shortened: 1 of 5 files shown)


google/cloud/spanner_dbapi/connection.py
```diff
@@ -183,6 +183,21 @@ def read_only(self, value):
...
+        Returns:
+            google.cloud.spanner_v1.RequestOptions:
+                Request options.
+        """
+        if self.request_priority is None:
+            return
+
+        req_opts = RequestOptions(priority=self.request_priority)
+        self.request_priority = None
+        return req_opts
```
IlyaFaer, line 199: Request options are used only for next one request

asthamohta: Is there a reason for this?

IlyaFaer: Spanner is designed this way, otherwise there is no sense in request options.
Let's say there is a request priority. If it'll be set for a connection/session, than all the requests will have the same priority. It's technically the same that not setting priority at all.

asthamohta (approved):

PR #867, opened 2022-12-13
feat: fgac support and samples

asthamohta:
`enable_fine_grained_access` sample requires service account to work, so skipping integration tests for the same.


rajatbhatta (approved):
LGTM, with a few minor comments.

PR #874, opened 2022-12-26
tests: adding support for array testing in postgres

asthamohta:


tests/system/test_session_api.py
```diff
@@ -180,12 +178,33 @@
...
+    PostGresAllTypesRowData(pkey=205, float_array=[]),
+    PostGresAllTypesRowData(pkey=206, string_array=[]),
+    PostGresAllTypesRowData(pkey=207, timestamp_array=[]),
+    PostGresAllTypesRowData(pkey=208, numeric_array=[]),
+    PostGresAllTypesRowData(pkey=209, jsonb_array=[]),
+    # non-empty array values, including nulls
+    PostGresAllTypesRowData(pkey=301, int_array=[123, 456, None]),
+    PostGresAllTypesRowData(pkey=302, bool_array=[True, False, None]),
+    PostGresAllTypesRowData(pkey=303, bytes_array=[BYTES_1, BYTES_2, None]),
+    PostGresAllTypesRowData(pkey=304, date_array=[SOME_DATE, None]),
```
rajatbhatta, line 202: Should we add another date here, similar to how we have for other arrays?
```suggestion
    PostGresAllTypesRowData(pkey=304, date_array=[SOME_DATE, SOME_DATE, None]),
```

tests/system/test_session_api.py
```diff
@@ -180,12 +178,33 @@
...
+    PostGresAllTypesRowData(pkey=305, float_array=[3.1415926, 2.71828, None]),
```
rajatbhatta, line 203: Should we check for `POS_INF`, and `NEG_INF` as well?

rajatbhatta (approved):
LTGM.

PR #909, opened 2023-03-07
feat: Proto Columns Feature

harshachinta:
This PR adds code changes for Proto Columns feature support in Python Client along with sample snippets.


google/cloud/spanner_v1/_helpers.py
```diff
@@ -17,9 +17,12 @@
 import datetime
 import decimal
 import math
+import base64
 
 from google.protobuf.struct_pb2 import ListValue
 from google.protobuf.struct_pb2 import Value
+from google.protobuf.message import Message
```
asthamohta, line 24: You need to expose these types in the init file for customers to be able to use this data type

google/cloud/spanner_v1/_helpers.py
```diff
@@ -170,6 +173,13 @@ def _make_value_pb(value):
             return Value(null_value="NULL_VALUE")
         else:
             return Value(string_value=value)
+    if isinstance(value, Message):
+        value = value.SerializeToString()
+        if value is None:
+            return Value(null_value="NULL_VALUE")
+        else:
+            b = base64.b64encode(value)
```
asthamohta, line 181: Can we have this in a single line: `return Value(string_value=base64.b64encode(value))`

google/cloud/spanner_v1/_helpers.py
```diff
@@ -198,7 +208,7 @@ def _make_list_value_pbs(values):
     return [_make_list_value_pb(row) for row in values]
 
 
-def _parse_value_pb(value_pb, field_type):
+def _parse_value_pb(value_pb, field_type, field_name, column_info=None):
```
asthamohta, line 211: Assign default value to field_name

google/cloud/spanner_v1/_helpers.py
```diff
@@ -207,6 +217,12 @@ def _parse_value_pb(value_pb, field_type):
     :type field_type: :class:`~google.cloud.spanner_v1.types.Type`
     :param field_type: type code for the value
 
+    :type field_name: str
```
asthamohta, line 220: Add optional in the documentation

google/cloud/spanner_v1/data_types.py
```diff
@@ -71,3 +75,109 @@ def serialize(self):
             return json.dumps(self._array_value, sort_keys=True, separators=(",", ":"))
 
         return json.dumps(self, sort_keys=True, separators=(",", ":"))
+
+
+def __proto_message(bytes_val, proto_message_object):
```
asthamohta, line 80: Why are we using double underscores here? My understanding is that the intent was for it to be a private fxn and hence should be a single underscore.
Reference: [link](https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-single-and-double-underscore-before-an-object-name)

google/cloud/spanner_v1/data_types.py
```diff
@@ -71,3 +75,109 @@ def serialize(self):
...
+    Raises:
+        ValueError: if the input int_list is not of type list
+    """
+    if isinstance(int_value, types.NoneType):
+        return None
+
+    if not isinstance(proto_enum_object, EnumTypeWrapper):
+        raise ValueError("Input proto_enum_object should be of type EnumTypeWrapper")
+
+    if not isinstance(int_value, (int, list)):
```
asthamohta, line 177: You are only checking for int and list here, should be int and list of int

google/cloud/spanner_v1/param_types.py
```diff
@@ -71,3 +73,35 @@ def Struct(fields):
...
+
+
+def ProtoMessage(proto_message_object):
+    """Construct a proto message type description protobuf.
+
+    :type proto_message_object: :class:`google.protobuf.message.Message`
+    :param proto_message_object: the proto message instance
+
+    :rtype: :class:`type_pb2.Type`
+    :returns: the appropriate struct-type protobuf
```
asthamohta, line 85: nit: proto-message-type

google/cloud/spanner_v1/param_types.py
```diff
@@ -71,3 +73,35 @@ def Struct(fields):
...
+
+
+def ProtoEnum(proto_enum_object):
+    """Construct a proto enum type description protobuf.
+
+    :type proto_enum_object: :class:`google.protobuf.internal.enum_type_wrapper.EnumTypeWrapper`
+    :param proto_enum_object: the proto enum instance
+
+    :rtype: :class:`type_pb2.Type`
+    :returns: the appropriate struct-type protobuf
```
asthamohta, line 101: nit: proto-enum-type

samples/samples/snippets.py
```diff
@@ -278,6 +280,46 @@ def create_database_with_default_leader(instance_id, database_id, default_leader
 # [END spanner_create_database_with_default_leader]
 
 
+# [START spanner_create_database_with_proto_descriptors]
+def create_database_with_proto_descriptors(instance_id, database_id):
+    """Creates a database with proto descriptors and tables with proto columns for sample data."""
+    spanner_client = spanner.Client(client_options={"api_endpoint": "staging-wrenchworks.sandbox.googleapis.com"})
```
asthamohta, line 286: Remove staging env

samples/samples/snippets.py
```diff
@@ -306,16 +348,57 @@ def update_database_with_default_leader(instance_id, database_id, default_leader
...
     """Gets the database DDL statements."""
-    spanner_client = spanner.Client()
+    spanner_client = spanner.Client(client_options={"api_endpoint": "staging-wrenchworks.sandbox.googleapis.com"})
     instance = spanner_client.instance(instance_id)
     database = instance.database(database_id)
     ddl = spanner_client.database_admin_api.get_database_ddl(database=database.name)
     print("Retrieved database DDL for {}".format(database_id))
     for statement in ddl.statements:
         print(statement)
+    print(ddl.proto_descriptors)
```
asthamohta, line 401: Change checks in snippets_test.py to check for correct value

samples/samples/testdata/singer_pb2.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
```
asthamohta, line 1: Add license to all new files

google/cloud/spanner_v1/_helpers.py
harshachinta: @asthamohta 
`field_name` is **not** an optional parameter and it is similar to `field_type`. The `field_name` is derived from the same dict as that of `field_type`. You can refer it [here](https://github.com/googleapis/python-spanner/pull/909/files#diff-5fcb4175b487a4a9418b28ee746c4397a8862ded6389818dd5c501c608ee8504R102-R103).
I guess this method is internal by the naming convention and hence it would not be a breaking change.
Just want to recheck if this method is public or used externally elsewhere by users? If yes then we can rethink.

google/cloud/spanner_v1/_helpers.py
harshachinta: As mentioned in [comment](https://github.com/googleapis/python-spanner/pull/909#discussion_r1135549179), this is not an optional field.

google/cloud/spanner_v1/data_types.py
harshachinta: This is a bit confusing as different sources say it differently.
Source: https://www.geeksforgeeks.org/private-methods-in-python/
Text: to define a private method prefix the member name with the double underscore “__”.

However, I will modify it to contain single underscore to follow the standard convention used here.

samples/samples/testdata/singer_pb2.py
harshachinta: singer_pb2.py file is autogenerated by protocol buffer compiler. The similar generated files such as [spanner_pb2.py](https://github.com/googleapis/python-spanner/blob/main/benchmark/benchwrapper/proto/spanner_pb2.py), [spanner_pb2_grpc.py](https://github.com/googleapis/python-spanner/blob/main/benchmark/benchwrapper/proto/spanner_pb2_grpc.py) do not have license added in them.

google/cloud/spanner_v1/_helpers.py
harshachinta: I guess init file has only the types that exist in python-spanner client library. The Message is a type from google.protobuf.message and I don't think this should be added.
Additionally, customers do not use this type directly. They use the references of proto message and enum from the auto generated code ([singer_pb2.py](https://github.com/googleapis/python-spanner/pull/909/files#diff-9ba15bb21b15c25ea6a17fb3c6ec868cfb26dcbeb403f50b21642f44433d2684)). For ex, the sample for this would look like [insert sample](https://github.com/googleapis/python-spanner/pull/909/files#diff-62f4ce45093b6508acb281948c5d4ab7121da9db4d7ad1eb1dd69d58c42a0f1eR2521) where the customer does not use Message directly.

samples/samples/snippets.py
harshachinta: `ddl.proto_descriptors` is of type bytes and it would be difficult to check for correct value. Do you prefer removing this line instead?

google/cloud/spanner_v1/_helpers.py
asthamohta: Okay got it.

google/cloud/spanner_v1/data_types.py
asthamohta: Yes, thankyou

samples/samples/snippets.py
asthamohta: You don't have to check the whole value, just check for a keyword that would be present in the output

harshachinta: the output is a byte string which is not very readable. let me figure it out.

asthamohta (approved):

samples/samples/testdata/singer_pb2.py
harshachinta: Done

google/cloud/spanner_v1/data_types.py
```diff
@@ -71,3 +75,109 @@ def serialize(self):
...
+    if isinstance(int_val, types.NoneType):
+        return None
+
+    if not isinstance(int_val, int):
+        raise ValueError("Expected input int_val to be a integer")
+
+    return proto_enum_object.Name(int_val)
+
+
+def get_proto_message(bytes_string, proto_message_object):
```
gauravpurohit06, line 128: Can we use more meaningful name for `bytes_string ` if the type can be list?

google/cloud/spanner_v1/_helpers.py
```diff
@@ -234,18 +249,38 @@ def _parse_value_pb(value_pb, field_type):
...
     elif type_code == TypeCode.NUMERIC:
         return decimal.Decimal(value_pb.string_value)
     elif type_code == TypeCode.JSON:
         return JsonObject.from_str(value_pb.string_value)
+    elif type_code == TypeCode.PROTO:
+        bytes_value = base64.b64decode(value_pb.string_value)
+        if column_info is not None and column_info.get(field_name) is not None:
+            proto_message = column_info.get(field_name)
+            if isinstance(proto_message, Message):
+                proto_message = proto_message.__deepcopy__()
```
gauravpurohit06, line 273: Why are we doing `__deepcopy__ ` here? Usually the user will be passing the default object which won't be having any value associated with it? 
- Does `ParseFromString ` method completely ignores the value which are present internally and create a new object out of `bytes` or merge in the existing object?

samples/samples/snippets.py
```diff
@@ -2428,6 +2511,165 @@ def enable_fine_grained_access(
     # [END spanner_enable_fine_grained_access]
 
 
+# [START spanner_insert_proto_columns_data_with_dml]
+def insert_proto_columns_data_with_dml(instance_id, database_id):
```
gauravpurohit06, line 2515: The samples should be consistent across the languages, Can you please recheck all the samples w.r.t to Go & Java Client?

gauravpurohit06 (changes requested):

samples/samples/snippets.py
harshachinta: Java has samples only for mutations. However, I think we should have samples that shows both DML and mutations for read and write. Let's discuss offline on this and conclude the samples needed.

google/cloud/spanner_v1/session.py
```diff
@@ -245,6 +250,7 @@ def execute_sql(
         query_mode=None,
         query_options=None,
         request_options=None,
+        column_info=None,
```
gauravpurohit06, line 253: Please move `column_info ` in the last otherwise it would be a breaking change if some customer is passing the method arguments in order and not using dictionary to pass arguments.

google/cloud/spanner_v1/snapshot.py
```diff
@@ -167,6 +167,7 @@ def read(
         limit=0,
         partition=None,
         request_options=None,
+        column_info=None,
```
gauravpurohit06, line 170: Breaking Change, same as above.

gauravpurohit06 (changes requested):

google/cloud/spanner_v1/_helpers.py
gauravpurohit06: ```ParseFromString clears the object first and then merges the data into the object. It does not create a new object/instance.```
Then you should use the shallow copy, if you are clearing the values then I don't see any value in deep copying the object.

google/cloud/spanner_v1/session.py
harshachinta: Thanks for pointing this out. Changes done

google/cloud/spanner_v1/_helpers.py
harshachinta: Shallow copies of proto messages are not possible. A shallow copy means constructing a new object and then populating it with references to the child objects found in the original. Let's say you have a message with in a message and now if you shallow copy you get the new object of the outer message, but for the inner message you share the reference to that in the original. This has so many complications and is not used for proto messages.

This is why deep copy is preferred and used.

gauravpurohit06 (approved):
LGTM, As discussed offline just remove the tags from the additional samples which we don't want to publish.

google/cloud/spanner_v1/_helpers.py
harshachinta: Shallow copy of Proto messages are not possible. A shallow copy of an object is a copy whose properties share the same references as those of the source object from which the copy was made. This implies, if a proto message is shallow copied from another, then both of it would be pointing to same inner references. Hence going forward with deep copy.

PR #910, opened 2023-03-13
feat: foreign key on delete cascade action testing and samples

asthamohta:


samples/samples/snippets.py
```diff
@@ -2428,6 +2428,105 @@ def enable_fine_grained_access(
...
+        )
+    )
+
+
+# [END spanner_create_table_with_foreign_key_delete_cascade]
+
+
+# [START spanner_alter_table_with_foreign_key_delete_cascade]
+def alter_table_with_foreign_key_delete_cascade(instance_id, database_id):
+    """Creates a table with foreign key delete cascade action"""
```
vishwarajanand, line 2473:
```suggestion
    """Alters a table with foreign key delete cascade action"""
```

samples/samples/snippets_test.py
```diff
@@ -781,3 +781,25 @@ def test_list_database_roles(capsys, instance_id, sample_database):
...
+
+@pytest.mark.dependency(name="create_table_with_foreign_key_delete_cascade")
+def test_create_table_with_foreign_key_delete_cascade(capsys, instance_id, sample_database):
+    snippets.create_table_with_foreign_key_delete_cascade(instance_id, sample_database.database_id)
+    out, _ = capsys.readouterr()
+    assert "Created Customers and ShoppingCarts table with FKShoppingCartsCustomerId" in out
+
+
+@pytest.mark.dependency(name="alter_table_with_foreign_key_delete_cascade",
+                        depends=["create_table_with_foreign_key_delete_cascade"])
```
anuragsrivstv, line 794: Hi, As we are adding `Foreign Key Delete Cascade` constraint , shouldn't is should **Not** depend on sample which has already added the constraints while creating the tables ?

samples/samples/snippets_test.py
```diff
@@ -781,3 +781,25 @@ def test_list_database_roles(capsys, instance_id, sample_database):
     snippets.list_database_roles(instance_id, sample_database.database_id)
     out, _ = capsys.readouterr()
     assert "new_parent" in out
+
+
+@pytest.mark.dependency(name="create_table_with_foreign_key_delete_cascade")
+def test_create_table_with_foreign_key_delete_cascade(capsys, instance_id, sample_database):
+    snippets.create_table_with_foreign_key_delete_cascade(instance_id, sample_database.database_id)
+    out, _ = capsys.readouterr()
+    assert "Created Customers and ShoppingCarts table with FKShoppingCartsCustomerId" in out
```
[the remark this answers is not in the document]
asthamohta: For IT, yes but for samples we just need the to check if an error is not thrown

samples/samples/snippets_test.py
asthamohta: It is common in samples to have this dependency to reduce the time to run these

tests/system/test_database_api.py
```diff
@@ -21,13 +21,16 @@
...
 from google.cloud.spanner_v1.pool import FixedSizePool, PingingPool
+from google.cloud.spanner_admin_database_v1 import DatabaseDialect
 from google.type import expr_pb2
 from . import _helpers
 from . import _sample_data
-from google.cloud.spanner_admin_database_v1 import DatabaseDialect
 
 
 DBAPI_OPERATION_TIMEOUT = 240  # seconds
+FKDCA_CUSTOMERS_COLUMNS = ("CustomerId", "CustomerName")
```
surbhigarg92, line 31: nit: Some places we are using `FKADC` and other places we are using `FKDCA` . It will be good to use same terminology everywhere.

tests/_fixtures.py
```diff
@@ -68,6 +68,19 @@
      email STRING(MAX),
      deleted BOOL NOT NULL )
      PRIMARY KEY(id, commit_ts DESC);
+
+CREATE TABLE Customers (
```
surbhigarg92, line 72: Why are we adding Customers table here when we already have `FKDAC_DDL` to add customers table.

Same for PG

rajatbhatta (approved):

PR #934, opened 2023-04-27
fix: Catch rst stream error for all transactions

asthamohta:
Also fixing names for tests in tests/unit/spanner_dbapi/test_connection.py


tests/unit/test__helpers.py
```diff
@@ -669,3 +670,76 @@ def test(self):
...
+        import functools
+
+        test_api = mock.create_autospec(self.test_class)
+        test_api.test_fxn.side_effect = [
+            InternalServerError("testing"),
+            InternalServerError("testing"),
+        ]
+
+        with self.assertRaises(InternalServerError):
+            _retry(functools.partial(test_api.test_fxn), retry_count=1)
```
surbhigarg92, line 726: How are we verifying retry is only done once , We can assert on `test_api.test_fxn.call_count` to verify this ?

tests/unit/test__helpers.py
```diff
@@ -669,3 +670,76 @@ def test(self):
...
+
+    def test_check_rst_stream_error(self):
```
surbhigarg92, line 728: Can we also add a test when the function first throws allowed_exception and then throws a different one?

We can also check that the thrown exception is the last one.

tests/unit/test_transaction.py
```diff
@@ -188,6 +188,25 @@ def test_begin_ok(self):
             "CloudSpanner.BeginTransaction", attributes=TestTransaction.BASE_ATTRIBUTES
         )
 
+    def test_begin_w_retry(self):
```
surbhigarg92, line 191: Can we also add a test for a query or read method ?

google/cloud/spanner_v1/_helpers.py
```diff
@@ -292,3 +293,57 @@ def _metadata_with_prefix(prefix, **kw):
...
+                    allowed_exceptions is not None
+                    and allowed_exceptions[exc.__class__] is not None
+                ):
+                    allowed_exceptions[exc.__class__](exc)
+                time.sleep(delay)
+                delay = delay * 2
+                retries = retries + 1
+            else:
+                raise exc
+        else:
```
surbhigarg92, line 334: We don't need the else block, we can `return func()` instead of `result = func()` at line 319

rajatbhatta: +1

google/cloud/spanner_v1/_helpers.py
```diff
@@ -292,3 +293,57 @@ def _metadata_with_prefix(prefix, **kw):
...
+    allowed_exceptions=None,
+):
+    """
+    Retry a function with a specified number of retries, delay between retries, and list of allowed exceptions.
+
+    Args:
+        func: The function to be retried.
+        retry_count: The maximum number of times to retry the function.
+        delay: The delay in seconds between retries.
+        allowed_exceptions: A tuple of exceptions that are allowed to occur without triggering a retry.
```
asthamohta, line 311: Add in docs that if this is None, it will retry for all exceptions

google/cloud/spanner_v1/_helpers.py
```diff
@@ -292,3 +293,57 @@ def _metadata_with_prefix(prefix, **kw):
...
+        The result of the function if it is successful, or raises the last exception if all retries fail.
+    """
+    retries = 0
+    while retries <= retry_count:
+        try:
+            result = func()
+        except Exception as exc:
+            if (
+                allowed_exceptions is None or exc.__class__ in allowed_exceptions
+            ) and retries < retry_count:
```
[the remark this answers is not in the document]
asthamohta: As discussed this is there to throw the error in the last cycle

google/cloud/spanner_v1/transaction.py
```diff
@@ -33,6 +33,9 @@
 from google.cloud.spanner_v1._opentelemetry_tracing import trace_call
 from google.cloud.spanner_v1 import RequestOptions
 from google.api_core import gapic_v1
+from google.cloud.spanner_v1._helpers import _retry
```
surbhigarg92, line 36: nit: Can you add these imports from helper class at https://github.com/googleapis/python-spanner/pull/934/files#diff-c0a324ccb53381adcd6c782a1d6e4cc18d543a7de1ae49a766977af3ab9bbbf6L20

google/cloud/spanner_v1/snapshot.py
```diff
@@ -36,6 +36,8 @@
     _metadata_with_leader_aware_routing,
 )
 from google.cloud.spanner_v1._helpers import _SessionWrapper
+from google.cloud.spanner_v1._helpers import _retry
```
surbhigarg92, line 39: nit: Can you add these imports at line https://github.com/googleapis/python-spanner/pull/934/files#diff-0e91978e1652aecf8960c9fe642218a42229c21d62eda7589eb60502f7f9f680L34

surbhigarg92 (approved):

PR #937, opened 2023-05-08
samples: bit reverse sequence

surbhigarg92:


samples/samples/snippets.py
```diff
@@ -2548,6 +2549,122 @@ def drop_foreign_key_constraint_delete_cascade(instance_id, database_id):
...
+    print(
+        "Altered Customers table to drop DEFAULT from CustomerId column and dropped the Seq sequence on database {} on instance {}".format(
+            database_id, instance_id
+        )
+    )
+
+
+# [END spanner_drop_sequence]
+
 if __name__ == "__main__":  # noqa: C901
```
rajatbhatta, line 2668: Shall we do `subparsers.add_parser` for the newly added samples as well?

samples/samples/snippets_test.py
```diff
@@ -816,3 +797,36 @@ def test_drop_foreign_key_contraint_delete_cascade(capsys, instance_id, sample_d
...
+    out, _ = capsys.readouterr()
+    assert (
+        "Altered Seq sequence to skip an inclusive range between 1000 and 5000000 on database"
+        in out
+    )
+    assert "Number of customer records inserted is 3" in out
+    assert "Inserted customer record with Customer Id:" in out
+
+
+@pytest.mark.dependency(depends=["create_sequence"])
```
rajatbhatta, line 825: Shall we have this dependency on `alter_sequence` instead of `create_sequence`? In that way, we'll be sure that `drop_sequence` doesn't get called before `alter_sequence` is executed.

rajatbhatta (approved):

PR #940, opened 2023-05-11
feat: Set LAR True

asthamohta:
BEGIN_COMMIT_OVERRIDE
feat: Enable leader aware routing by default. This update contains performance optimisations that will reduce the latency of read/write transactions that originate from a region other than the default leader region.
END_COMMIT_OVERRIDE


google/cloud/spanner_dbapi/connection.py
```diff
@@ -547,7 +547,7 @@ def connect(
 
     :type route_to_leader_enabled: boolean
     :param route_to_leader_enabled:
-        (Optional) Default False. Set route_to_leader_enabled as True to
+        (Optional) Default True. Set route_to_leader_enabled as True to
```
surbhigarg92, line 550:
```Set route_to_leader_enabled as False to
                   Disable leader aware routing. Disbling leader aware routing
                   would route all requests in RW/PDML transactions to the
                   closest region.```

File: google/cloud/spanner_v1/client.py [hunk 2]
```diff
@@ -116,7 +116,7 @@ class Client(ClientWithProject):
 
     :type route_to_leader_enabled: boolean
     :param route_to_leader_enabled:
-        (Optional) Default False. Set route_to_leader_enabled as True to
+        (Optional) Default True. Set route_to_leader_enabled as True to
```
surbhigarg92, line 119: Same as above

surbhigarg92 (changes requested):

surbhigarg92 (changes requested):

surbhigarg92 (approved):

PR #957, opened 2023-06-07
feat: Session leaks

surbhigarg92:
Adding a couple of option in [Database](https://github.com/googleapis/python-spanner/blob/main/google/cloud/spanner_v1/database.py#L138) and [Instance](https://github.com/googleapis/python-spanner/blob/main/google/cloud/spanner_v1/instance.py#L426) for auto-detecting long running transactions, logging the stack-trace and removing such transactions from consuming resources. 

**close_inactive_transactions** - Default value is False. Setting this to True will close the faulty transactions that are consuming the resources. We will also be generating the warning logs which can be later referred to know the transactions that were closed through this option.
**logging_enabled** -  Default value is True. Setting this to False will stop generating warning logs  which helps in identifying faulty code which is leading to incorrect consumption of sessions.

Batch and Partitioned Transactions are long running and will not be auto cleaned


google/cloud/spanner_v1/pool.py
```diff
@@ -126,6 +138,19 @@ def _new_session(self):
...
         )
 
+    def _format_trace(self, frames: List[traceback.FrameSummary]) -> str:
+        """Helper for formatting traceback frames into more readable trace.
+        :type frames: list[traceback.FrameSummary]
+        :param frames: list of stack frames
+        :rtype: str
+        :returns: fromatted trace.
+        """
+        stack = [
```
asthamohta, line 148: Can we format this in the python format and add a test case for it?

google/cloud/spanner_v1/database.py
```diff
@@ -108,11 +108,20 @@ class Database(object):
                  passed, the database will construct an instance of
                  :class:`~google.cloud.spanner_v1.pool.BurstyPool`.
 
+    :type logging_enabled: boolean
+    :parama logging_enabled: (Optional) Represents whether the database
+                    has logging enabled or not. Default is True
+
```
asthamohta, line 114: In database close() can we add code to close all bg tasks and a test for the same.

google/cloud/spanner_v1/pool.py
```diff
@@ -138,6 +163,113 @@ def session(self, **kwargs):
...
+            # Set the session as None for associated objects (transaction, batch, snapshot)
+            if session._transaction is not None:
+                session._transaction._session = None
+            if session._batch is not None:
+                session._batch._session = None
+            if session._snapshot is not None:
+                session._snapshot._session = None
+
+            # Increment the count of closed transactions and return the session to the pool
+            transactions_closed += 1
```
asthamohta, line 260: integer values are passed by value and not reference so transaction_closed will remain 0 in deleteLongRunningTransactions. Objects are by reference. you can create an object for this

google/cloud/spanner_v1/pool.py
```diff
@@ -316,6 +481,27 @@ def get(self):
             if not session.exists():
                 session = self._new_session()
                 session.create()
+
+        # Set session properties.
```
asthamohta, line 485: Can we add this is another function altogether? Looks repetitive

google/cloud/spanner_v1/pool.py
```diff
@@ -138,6 +163,113 @@ def session(self, **kwargs):
...
+            ]
+
+            for session in sessions_to_delete:
+                self._recycle_session(session, transactions_closed)
+
+            # Calculate and sleep for the time remaining until the next iteration based on the interval.
+            iteration_elapsed = time.time() - iteration_start
+            remaining_time = longRunningTransactionInterval_sec - iteration_elapsed
+            if remaining_time > 0:
+                time.sleep(remaining_time)
```
asthamohta, line 234: Can we check that this is running as a bg task and during sleep a different task is being run?

google/cloud/spanner_v1/pool.py
```diff
@@ -235,6 +369,26 @@ def get(self, timeout=None):
...
+        self._traces[session._session_id] = traceback.extract_stack()
+
+        # Start background task for handling long-running transactions if 95% of sessions are checked out.
+        if (
+            self._database.close_inactive_transactions or self._database.logging_enabled
+        ) and len(self._borrowed_sessions) / self.size >= 0.95:
+            self.startCleaningLongRunningSessions()
+
+        # Log a warning message if Session pool is exhausted.
+        if (
```
asthamohta, line 386: Should be if else for 100% and 95%, otherwise two logs will be printed

google/cloud/spanner_v1/pool.py
```diff
@@ -247,7 +401,16 @@ def put(self, session):
 
         :raises: :exc:`queue.Full` if the queue is full.
         """
+        if self._borrowed_sessions.__contains__(session):
```
asthamohta, line 404: Should we pop the traces as well?

google/cloud/spanner_v1/session.py
```diff
@@ -62,6 +62,8 @@ class Session(object):
 
     _session_id = None
     _transaction = None
+    _batch = None
```
asthamohta, line 65: Club into one

google/cloud/spanner_v1/pool.py
surbhigarg92: Removed the custom method and using the method available in traceback library.

google/cloud/spanner_v1/pool.py
```diff
@@ -235,6 +369,26 @@ def get(self, timeout=None):
             session = self._database.session()
             session.create()
 
+        # Set session properties.
```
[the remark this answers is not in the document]
surbhigarg92: We already have `session.exists()` call after the session is fetched. This function makes sure that the session is alive.

google/cloud/spanner_v1/pool.py
surbhigarg92: Added

google/cloud/spanner_v1/pool.py
```diff
@@ -328,7 +514,17 @@ def put(self, session):
         :param session: the session being returned.
         """
         try:
+            if self._borrowed_sessions.__contains__(session):
+                self._borrowed_sessions.remove(session)
             self._sessions.put_nowait(session)
+            self._traces.pop(session._session_id, None)
```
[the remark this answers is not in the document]
surbhigarg92: Here we can not extract out the common method, because there is a very slight difference in few properties here. Like Burstpool has `self. target_size` while other two pool has `self.size` . Pinging pool does `self._sessions.put_nowait((_NOW() + self._delta, session))` while other two pool does `self._sessions.put_nowait(session)` .

google/cloud/spanner_v1/database.py
surbhigarg92: database does not have a close() method in Python. Every pool has a clear method which deletes all the session, added stopCleaningLongRunningSessions call in the same method.  But that method is not referenced any where .

google/cloud/spanner_v1/pool.py
```diff
@@ -138,6 +157,115 @@ def session(self, **kwargs):
...
+            iteration_elapsed = time.time() - iteration_start
+            remaining_time = longRunningTransactionFrequency_sec - iteration_elapsed
+            if remaining_time > 0:
+                time.sleep(remaining_time)
+
+        AbstractSessionPool._cleanup_task_ongoing = False
+        AbstractSessionPool._cleanup_task_ongoing_event.clear()
+
+    def _recycle_session(self, session):
+        """Helper method to recycle session for background task.
```
arpan14, line 234: 1. We can avoid "background task" in method documentation.
2. Instead of "recycle_session" which is again a new terminology can we keep it consistent? We are already using close_long_running_sessions / close_long_running_transactions.

google/cloud/spanner_v1/pool.py
```diff
@@ -138,6 +157,115 @@ def session(self, **kwargs):
...
+                break
+            iteration_start = time.time()
+
+            # Retrieve a list of sessions to delete.
+            sessions_to_delete = [
+                session
+                for session in self._borrowed_sessions
+                if (datetime.datetime.utcnow() - session.checkout_time)
+                > datetime.timedelta(seconds=longRunningTransactionThreshold_sec)
+                and not session.long_running
```
arpan14, line 218: 1. Should we re-name `session.long_running` to `session.eligible_for_long_running`? What we actually mean to say is if the session is allowed to be long running or not. If not (PDML & partitioned reads) then we clean it. While I fully understand the background behind this naming, a new reader will think `session.long_running` means that a session is long running. And then this check looks confusing that we are cleaning it when session is "not long running"
2.  Suggestion : It will be better to have `not session.long_running` as the first check so that we short circuit for a few times (PDML/partitioned reads)

tests/system/test_session_api.py
```diff
@@ -2517,6 +2517,141 @@ def test_partition_query(sessions_database, not_emulator):
     batch_txn.close()
 
 
+def test_should_close_inactive_transactions_with_bursty_pool_and_transaction(
```
arpan14, line 2520: Can we add tests that for PDML/partitioned reads we shouldn't be closing inactive sessions?

tests/unit/test_snapshot.py
```diff
@@ -1155,6 +1168,22 @@ def test_partition_read_other_error(self):
             ),
         )
 
+    def test_partition_read_should_throw_error_for_recycled_session(self):
```
arpan14, line 1171: Theoretically for paritioned reads, sessions should never be cleaned up right? It will be better to change this test. Someone reading the tests can get a different impression and think there to be bug.

tests/unit/test_snapshot.py
```diff
@@ -1314,6 +1343,19 @@ def test_partition_query_other_error(self):
             attributes=dict(BASE_ATTRIBUTES, **{"db.statement": SQL_QUERY}),
         )
 
+    def test_partition_query_should_throw_error_for_recycled_session(self):
```
arpan14, line 1346: same comment as partioned read. we would never have this situation practically right?

google/cloud/spanner_v1/database.py
```diff
@@ -1038,7 +1068,9 @@ def __exit__(self, exc_type, exc_val, exc_tb):
                     "CommitStats: {}".format(self._batch.commit_stats),
                     extra={"commit_stats": self._batch.commit_stats},
                 )
-            self._database._pool.put(self._session)
+            if self._batch._session is not None:
+                self._database._pool.put(self._session)
+            self._session._transaction = None
```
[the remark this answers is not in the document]
surbhigarg92: We are also deleting this in commit and rollback methods, apart from all these methods.  https://github.com/googleapis/python-spanner/blob/main/google/cloud/spanner_v1/transaction.py#L181
https://github.com/googleapis/python-spanner/blob/main/google/cloud/spanner_v1/transaction.py#L246

Also In case of retry due to abort errors. https://github.com/googleapis/python-spanner/blob/main/google/cloud/spanner_v1/session.py#L388

google/cloud/spanner_v1/database.py
```diff
@@ -1062,22 +1094,27 @@ class SnapshotCheckout(object):
...
     def __exit__(self, exc_type, exc_val, exc_tb):
         """End ``with`` block."""
-        if isinstance(exc_val, NotFound):
-            # If NotFound exception occurs inside the with block
-            # then we validate if the session still exists.
-            if not self._session.exists():
-                self._session = self._database._pool._new_session()
-                self._session.create()
-        self._database._pool.put(self._session)
+        if self._snapshot._session is not None:
```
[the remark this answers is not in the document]
surbhigarg92: To make sure the session is not already returned back to pool by the closelongrunningtransaction tasks.

google/cloud/spanner_v1/pool.py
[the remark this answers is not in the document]
asthamohta: Okay got it. Thanks

google/cloud/spanner_v1/_helpers.py
```diff
@@ -38,6 +38,10 @@
     + "numeric has a whole component with precision {}"
 )
 
+# Constants
+DELETE_LONG_RUNNING_TRANSACTION_FREQUENCY_SEC = 120
+DELETE_LONG_RUNNING_TRANSACTION_THRESHOLD_SEC = 3600
```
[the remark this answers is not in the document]
surbhigarg92: Python sleep library expects seconds to be passed and does not support this. Also its helpful in overriding these properties to a smaller value for tests. 

Seconds looks file to me.

google/cloud/spanner_v1/pool.py
```diff
@@ -138,6 +157,115 @@ def session(self, **kwargs):
...
+            # Increment the count of closed transactions and return the session to the pool
+            session_recycled = True
+            self.put(session)
+        elif self._database.logging_enabled:
+            # Log a warning for a potentially leaking long-running transaction.
+            # Only log the warning if it hasn't been logged already.
+            if not session.transaction_logged:
+                self._database.logger.warning(
+                    "Transaction has been running for longer than 60 minutes and might be causing a leak. "
+                    + "Enable closeInactiveTransactions in Session Pool Options to automatically clean such transactions. "
```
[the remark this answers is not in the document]
surbhigarg92: @asthamohta What do you think, we should have a constant message across libraries. And will adding session id help here ?

google/cloud/spanner_v1/database.py
```diff
@@ -108,11 +108,20 @@ class Database(object):
...
+
     :type logger: :class:`logging.Logger`
     :param logger: (Optional) a custom logger that is used if `log_commit_stats`
                    is `True` to log commit statistics. If not passed, a logger
                    will be created when needed that will log the commit statistics
                    to stdout.
+
+    :type close_inactive_transactions: boolean
+    :param close_inactive_transactions: (Optional) Represents whether the database
+                    has close inactive transactions enabled or not. Default is False
```
asthamohta, line 123: Maybe we can add a better description on what close inactive transaction means?

google/cloud/spanner_v1/database.py
```diff
@@ -647,7 +676,7 @@ def execute_partitioned_dml(
             )
 
         def execute_pdml():
-            with SessionCheckout(self._pool) as session:
+            with SessionCheckout(self._pool, isLongRunning=True) as session:
```
asthamohta, line 679: run_in_transaction also has a SessionCheckout call, should we not pass false over there?

asthamohta: Also the session.transaction will be unset for run_in_transaction, so that needs to be added

surbhigarg92: Default is False, so we don't need to pass. Also `run_in_transaction` is calling  `session.run_in_transaction` which is already setting the transaction . https://github.com/googleapis/python-spanner/blob/main/google/cloud/spanner_v1/session.py#L344

google/cloud/spanner_dbapi/connection.py
```diff
@@ -583,7 +583,15 @@ def connect(
 
     instance = client.instance(instance_id)
     conn = Connection(
-        instance, instance.database(database_id, pool=pool) if database_id else None
+        instance,
+        instance.database(
+            database_id,
+            pool=pool,
+            logging_enabled=False,
+            close_inactive_transactions=False,
```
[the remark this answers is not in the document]
surbhigarg92: Moved logging_enabled to pool class as this needs to be enabled at session pool level.

google/cloud/spanner_v1/instance.py
```diff
@@ -480,7 +490,9 @@ def database(
             self,
             ddl_statements=ddl_statements,
             pool=pool,
+            logging_enabled=logging_enabled,
```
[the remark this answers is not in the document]
surbhigarg92: These are not instance level properties. Instance class has database method which ultimately calls database class and create an object. So we need to give an option to pass the database properties in this method as well.

google/cloud/spanner_v1/pool.py
```diff
@@ -235,6 +365,19 @@ def get(self, timeout=None):
             session = self._database.session()
             session.create()
 
+        # Set session properties.
+        self._set_session_properties(session, isLongRunning)
+
+        # Start background task for handling long-running transactions if 95% of sessions are checked out.
+        if (
+            self._database.close_inactive_transactions or self._database.logging_enabled
```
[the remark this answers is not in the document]
surbhigarg92: Moved `logging_enabled` to pool class as discussed.

google/cloud/spanner_v1/database.py
surbhigarg92: Updated. Please check if it looks fine ?

google/cloud/spanner_v1/pool.py
surbhigarg92: Done

google/cloud/spanner_v1/session.py
surbhigarg92: Done

google/cloud/spanner_v1/pool.py
```diff
@@ -138,6 +157,115 @@ def session(self, **kwargs):
...
+            DELETE_LONG_RUNNING_TRANSACTION_THRESHOLD_SEC,
+        )
+
+        if (
+            not AbstractSessionPool._cleanup_task_ongoing_event.is_set()
+            and not AbstractSessionPool._cleanup_task_ongoing
+        ):
+            AbstractSessionPool._cleanup_task_ongoing_event.set()
+            if self._database.logging_enabled:
+                self._database.logger.info("95% of the session pool is exhausted")
```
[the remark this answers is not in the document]
surbhigarg92: Done

google/cloud/spanner_v1/pool.py
```diff
@@ -316,6 +471,23 @@ def get(self):
...
                 session = self._new_session()
                 session.create()
+
+        # Set session properties.
+        self._set_session_properties(session, isLongRunning)
+
+        # Start background task for handling long-running transactions if 95% of sessions are checked out.
+        if (
+            self._database.close_inactive_transactions or self._database.logging_enabled
+        ) and len(self._borrowed_sessions) / self.target_size >= 0.95:
```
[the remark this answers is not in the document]
surbhigarg92: Done

google/cloud/spanner_v1/pool.py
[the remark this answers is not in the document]
surbhigarg92: Updated the message . And as discussed we will not be adding session id in the log.

google/cloud/spanner_v1/pool.py
surbhigarg92: Done

asthamohta (dismissed):

chalmerlowe:

PR #961, opened 2023-06-22
feat(default_retention_period): Changes to backup.create API to suppo…

satyaprakashg:
…rt null expire time

With the default retention period feature, null value for expiration time is a valid value. If the expiration time is not specified (null) at the time of create backup then the backup will be retained for the duration of maximum retention period.

In this commit, the precondition check on expiration time in the create backup API path is removed. The condition remains intact for copy backup API and that would be addressed in a follow-up commit.

Fixes #288359014


satyaprakashg:
@asthamohta Could you please review this diff. Also, can you add a "do not merge" tag on this diff since the backend changes aren't merged yet.

asthamohta (approved):

PR #987, opened 2023-07-31
feat: Add maxBatchingDelayMs Option.

nginsberg-google:
Adds support for max batching delay option in Commit requests.

This feature allows users to express how much latency they are willing for spanner to add to attempt to batch their writes. Generally speaking, the more latency they are willing to tolerate, the more efficiently Spanner can serve the writes.


google/cloud/spanner_v1/batch.py
```diff
@@ -145,9 +146,14 @@ def _check_state(self):
         if self.committed is not None:
             raise ValueError("Batch already committed")
 
-    def commit(self, return_commit_stats=False, request_options=None):
+    def commit(self, max_batching_delay_ms=None, return_commit_stats=False, request_options=None):
```
harshachinta, line 149: Adding new argument `max_batching_delay_ms` as second parameter might break any existing applications that have not used key explicitly while calling `commit`.

Can we move this to end?
```suggestion
    def commit(self, return_commit_stats=False, request_options=None, max_batching_delay_ms=None):
```
@surbhigarg92 Can you please correct me this is incorrect?

google/cloud/spanner_v1/batch.py
```diff
@@ -182,11 +188,16 @@ def commit(self, return_commit_stats=False, request_options=None):
         # Request tags are not supported for commit requests.
         request_options.request_tag = None
 
+        Duration max_batching_delay = None
+        if max_batching_delay_ms is not None:
+            max_batching_delay.nanos = 1000000 * [max_batching_delay_ms]
```
harshachinta, line 193: How is this default decided?

google/cloud/spanner_v1/transaction.py
```diff
@@ -180,9 +181,14 @@ def rollback(self):
         self.rolled_back = True
         del self._session._transaction
 
-    def commit(self, return_commit_stats=False, request_options=None):
+    def commit(self, max_batching_delay_ms=None, return_commit_stats=False, request_options=None):
```
harshachinta, line 184: Same as above comment

harshachinta:
@nginsberg-google Do we still need this PR?

PR #996, opened 2023-08-09
feat: LAR for dbapi

asthamohta:


google/cloud/spanner_dbapi/connection.py
```diff
@@ -508,7 +508,7 @@ def connect(
     pool=None,
     user_agent=None,
     client=None,
-    route_to_leader_enabled=True,
+    route_to_leader_enabled=False,
```
surbhigarg92, line 511: Can we put a comment on top of it mentioning , that ILB should be enabled to enable LAR

surbhigarg92 (approved):

PR #1003, opened 2023-08-17
feat: return list of dictionaries for execute streaming sql

### LINKED ISSUE #958: Is there an option to return execute_sql results as a list of dictionaries?
right now the results returned from `snapshot.execute_sql()` gets returned as a list of arrays.  Is there an existing configuration to change the return format of the StreamedResultSet object?

I would like to use keys instead of row indices when pulling values out for further processing. If i ever had to update the running query I would have to make sure my indices are correct. I can easily convert the row to dictionary but want to see if there's a configuration i'm missing.

asthamohta:


google/cloud/spanner_v1/streamed.py
```diff
@@ -190,6 +190,19 @@ def one_or_none(self):
         except StopIteration:
             return answer
 
+    def fetch_as_dictionary_list(self):
```
surbhigarg92, line 193: Name can be better . Probably "to_dictionary" ?

google/cloud/spanner_v1/streamed.py
```diff
@@ -190,6 +190,19 @@ def one_or_none(self):
...
+        rows = []
```
surbhigarg92, line 194: Add documentation

google/cloud/spanner_v1/streamed.py
```diff
@@ -190,6 +190,19 @@ def one_or_none(self):
...
+        for row in self:
```
surbhigarg92, line 195: What happens if rows are not fully processed when this method is called ?

google/cloud/spanner_v1/streamed.py
rajatbhatta: to_dictionary would mean it's a dict but the return type here is a list of dict.

rajatbhatta: Simply `dict_list`? Similar to method name `one`?

rajatbhatta: Does to_dict_list look good?

google/cloud/spanner_v1/streamed.py
asthamohta: There will be a delay in returning the result in those cases but always the full result will be returned

rajatbhatta (approved):

PR #1027, opened 2023-10-19
feat: Batch Write API implementation and samples

ghost:
This PR adds code changes and samples for Batch Write feature.
https://cloud.google.com/spanner/docs/batch-write


sunnsing-google:
@harshachinta - Can you please review this PR?

sunnsing-google:
> @harshachinta - Can you please review this PR?

@harshachinta / @surbhigarg92 - Can you please take a look?

google/cloud/spanner_v1/batch.py
```diff
@@ -215,6 +216,73 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+    committed = None
+
+    def __init__(self, session):
+        super(MutationGroups, self).__init__(session)
+        self._mutation_groups = []
+
+    def group(self):
+        """Returns a new mutation_group to which mutations can be added."""
+        mutation_group = BatchWriteRequest.MutationGroup()
+        self._mutation_groups.append(mutation_group)
```
harshachinta, line 235: Instead of modifying the `_BatchBase` class can we do this?
```suggestion
        batchbase = _BatchBase(self._session)
        self._mutation_groups.append(
        BatchWriteRequest.MutationGroup(mutations=batchbase._mutations
        )
        return batchbase
```
The problem here is the `_mutations` is a protected variable. Can we see if there is any way to do this without modifying the base class? I will also try to see if there is any better way to do this.

google/cloud/spanner_v1/batch.py
```diff
@@ -215,6 +216,73 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+            :class:`google.cloud.spanner_v1.types.RequestOptions`
+        :param request_options:
+                (Optional) Common options for this request.
+                If a dict is provided, it must be of the same form as the protobuf
+                message :class:`~google.cloud.spanner_v1.types.RequestOptions`.
+
+        :rtype: :class:`Iterable[google.cloud.spanner_v1.types.BatchWriteResponse]`
+        :returns: a sequence of responses for each batch.
+        """
+        if self.committed is not None:
```
harshachinta, line 251: nit: can we move this logic to `_check_state` so that it is clear that we check the state before applying the request.

google/cloud/spanner_v1/database.py
```diff
@@ -734,6 +735,17 @@ def batch(self, request_options=None):
         """
         return BatchCheckout(self, request_options)
 
+    def mutation_groups(self):
```
harshachinta, line 738: Can we pass the `request_options` here similar to existing `batch()` method?

google/cloud/spanner_v1/database.py
```diff
@@ -1040,6 +1052,39 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+        self._database = database
+        self._session = self._mutation_groups = None
+
+    def __enter__(self):
+        """Begin ``with`` block."""
+        session = self._session = self._database._pool.get()
+        return MutationGroups(session)
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        """End ``with`` block."""
```
harshachinta, line 1078: Can we call the `batch_write()` method when context manager `__exit__` so that the mutation groups get committed even when customer does not explicitly call it.
I see similar behaviour for existing mutations code. https://github.com/googleapis/python-spanner/blob/e5acb568c276063d45a6db58d0744e8e59a6afce/google/cloud/spanner_v1/database.py#L1026-L1030

google/cloud/spanner_v1/database.py
```diff
@@ -1040,6 +1052,39 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+        if isinstance(exc_val, NotFound):
+            # If NotFound exception occurs inside the with block
+            # then we validate if the session still exists.
```
harshachinta, line 1081: Is there any reason behind creating a new session here? since this is already handled by the pool when a session is checked out from the pool.
https://github.com/googleapis/python-spanner/blob/e5acb568c276063d45a6db58d0744e8e59a6afce/google/cloud/spanner_v1/pool.py#L235-L236

harshachinta (commented):
Took a look the implementation changes. Will take a look at tests tomorrow.

google/cloud/spanner_v1/batch.py
sunnsing-google: The suggested change doesn't work. I think this change needn't be blocking since it is an internal implementation detail and we can change it anytime later.

google/cloud/spanner_v1/batch.py
sunnsing-google: Done.

google/cloud/spanner_v1/batch.py
```diff
@@ -215,6 +216,73 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+                request=request,
+                metadata=metadata,
+            )
+            response = _retry(
+                method,
+                allowed_exceptions={InternalServerError: _check_rst_stream_error},
+            )
+        self.committed = True
+        return response
+
```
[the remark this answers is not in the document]
sunnsing-google: Context managers are typically expected to perform clean up tasks in `__exit__`. Here, we'd expect the client to call batch_write and consume the responses -- otherwise they won't know the outcome. I'd suggest to keep the explicit call only.

google/cloud/spanner_v1/database.py
sunnsing-google: As discussed offline, that would only be needed if we called `batch_write()` on `__exit__`. Since we aren't doing it, the current way is fine.

google/cloud/spanner_v1/database.py
sunnsing-google: Same comment as before about this being an explicit call the client should make and consume results.

google/cloud/spanner_v1/database.py
sunnsing-google: Yes, please see https://github.com/googleapis/python-spanner/pull/718.

google/cloud/spanner_v1/batch.py
```diff
@@ -215,6 +216,81 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+                 API requests.
+        """
+        if self.committed is not None:
+            raise ValueError("MutationGroups already committed")
+
+    def group(self):
+        """Returns a new mutation_group to which mutations can be added."""
+        mutation_group = BatchWriteRequest.MutationGroup()
+        self._mutation_groups.append(mutation_group)
+        return _BatchBase(self._session, mutation_group.mutations)
```
harshachinta, line 245: We are returning a class that is marked as private. We should not be doing this.

sunnsing-google: We could introduce a new public class -
```python
class MutationGroup(_BatchBase):
```

Caveat is that the client can create a standalone instance of this new class, but that should be okay.

What do you think?

sunnsing-google: Returned a public class in https://github.com/googleapis/python-spanner/pull/1027/commits/38d9b8344e67ddab60e1a3018004a10bf77104a5

google/cloud/spanner_v1/batch.py
```diff
@@ -36,15 +37,18 @@ class _BatchBase(_SessionWrapper):
...
     """
 
     transaction_tag = None
     _read_only = False
 
-    def __init__(self, session):
+    def __init__(self, session, mutations=[]):
         super(_BatchBase, self).__init__(session)
-        self._mutations = []
+        self._mutations = mutations
```
harshachinta, line 51: Revert the changes for `_BatchBase`. There is no need to pass mutations to this and set it.
```suggestion
        self._mutations = []
```

google/cloud/spanner_v1/batch.py
```diff
@@ -36,13 +37,16 @@ class _BatchBase(_SessionWrapper):
     """Accumulate mutations for transmission during :meth:`commit`.
 
     :type session: :class:`~google.cloud.spanner_v1.session.Session`
-    :param session: the session used to perform the commit
+    :param session: The session used to perform the commit.
```
harshachinta, line 40: This change should be removed

google/cloud/spanner_v1/batch.py
```diff
@@ -36,13 +37,16 @@ class _BatchBase(_SessionWrapper):
...
+
+    :type mutations: list
+    :param mutations: The list into which mutations are to be accumulated.
     """
 
     transaction_tag = None
     _read_only = False
 
-    def __init__(self, session):
+    def __init__(self, session, mutations=[]):
```
harshachinta, line 49: This change should be removed

google/cloud/spanner_v1/database.py
```diff
@@ -1040,6 +1052,39 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+    Caller must *not* use the object to perform API requests outside the scope
+    of the context manager.
+
+    :type database: :class:`~google.cloud.spanner_v1.database.Database`
+    :param database: database to use
+    """
+
+    def __init__(self, database):
+        self._database = database
+        self._session = self._mutation_groups = None
```
harshachinta, line 1070: Where are we using this  `self._mutation_groups`?

tests/system/test_session_api.py
```diff
@@ -2521,6 +2521,44 @@ def test_partition_query(sessions_database, not_emulator):
     batch_txn.close()
 
 
+def test_mutation_groups_insert_or_update_then_query(sessions_database):
```
harshachinta, line 2524: Emulator tests are failing as this feature is not implemented there.
Can you add `not_emulator` to skip emulatior.

google/cloud/spanner_v1/batch.py
sunnsing-google: Done

google/cloud/spanner_v1/batch.py
sunnsing-google: Done

tests/system/test_session_api.py
sunnsing-google: Done.

tests/unit/test_database.py
```diff
@@ -2679,6 +2693,163 @@ def test_process_w_query_batch(self):
...
+        from google.cloud._helpers import UTC
+        from google.cloud._helpers import _datetime_to_pb_timestamp
+        from google.cloud.spanner_v1.batch import MutationGroups
+        from google.rpc.status_pb2 import Status
+
+        now = datetime.datetime.utcnow().replace(tzinfo=UTC)
+        now_pb = _datetime_to_pb_timestamp(now)
+        status_pb = Status(code=200)
+        response = BatchWriteResponse(
+            commit_timestamp=now_pb, indexes=[0], status=status_pb
```
harshachinta, line 2738: I don't understand this case.
In the response we have indexes=[0], whereas request passes empty `mutation_groups`.

harshachinta: Agree that the test checks only the expected request parameters. But mocking correct response would help someone looking into this code later and not get confused.

harshachinta:
@sunnsing-google 
LGTM overall, except for one comment on correct mock response. Presubmits are failing and I am trying to figure out the issue.
Will approve once the presubmits get fixed.

tests/unit/test_database.py
sunnsing-google: Updated.

sunnsing-google:
> @sunnsing-google LGTM overall, except for one comment on correct mock response. Presubmits are failing and I am trying to figure out the issue. Will approve once the presubmits get fixed.

Thanks. PR is updated with the test change. PTAL.

harshachinta (approved):

PR #1028, opened 2023-10-25
Chore: Benchmarking

alkatrivedi:
Benchmarking long running session removals

### DIFF
--- a/samples/samples/benchmarking.py
+++ b/samples/samples/benchmarking.py
@@ -1,0 +1,126 @@
+import threading
+from google.cloud import spanner_v1
+import time
+import math
+
+from google.cloud.spanner_v1 import transaction, pool, _helpers
+from google.cloud import spanner
+import concurrent.futures
+
+# Define your Spanner instance and database information
+project_id = "your_project_id"
+instance_id = "your_instance_id"
+database_id = "your_database_id"
+spanner_client = spanner_v1.Client(project=project_id)
+
+# Create a Spanner database instance
+instance = spanner_client.instance(instance_id)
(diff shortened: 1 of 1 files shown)


samples/samples/benchmarking.py
```diff
@@ -0,0 +1,128 @@
...
+
+    # for row in results:
+    #     print("SingerId: {}, FirstName: {}, LastName: {}".format(*row))
+
+    end_time = time.time()
+    transaction_time.append(end_time-start_time)
+# [END spanner_query_data]
+
+# [START spanner_batch_transaction]
+def batch_transaction(thread_id):
```
arpan14, line 59: Where is this function `batch_transaction` invoked? Similarly where is `insert_with_dml` invoked?

samples/samples/benchmarking.py
```diff
@@ -0,0 +1,128 @@
...
+        )
+        print("{} record(s) inserted.".format(row_ct))
+
+    database.run_in_transaction(insert_singers)
+    end_time = time.time()
+    transaction_time.append(end_time-start_time)
+# [END insert_with_dml]
+
+# Define the number of threads
+num_threads = 20
```
arpan14, line 100: * At max only 10 threads will be used up since there are only 10 sessions. Post that 10 threads will always wait.
* Currently you are running 1 query per thread, it will be good to increase the number of transactions per thread. For example, 2000 transactions in total. 20 threads will make it 200 transactions per thread. This will ensure the run executes for more time and does not finish in < 1 second. This will also allow you to remove sleep() of 10s.

PR #1032, opened 2023-11-08
fix: Executing existing DDL statements on executemany statement execution

ankiaga:
More details here https://b.corp.google.com/issues/299533680#comment3


tests/system/test_dbapi.py
```diff
@@ -344,6 +344,50 @@ def test_DDL_autocommit(shared_instance, dbapi_database):
             op.result()
 
 
+def test_ddl_executemany(shared_instance, dbapi_database):
+    """Check that DDLs in autocommit mode are immediately executed."""
```
olavloite, line 348: Isn't this exactly what we know is not happening? What I mean is:
1. If you call `cur.execute(...)` with DDL statement, the DDL statement is NOT executed directly.
2. It is only executed once you call `execute` or `execute_many` with a non-DDL statement.

Can we also test the behavior when the connection is in `autocommit=false` mode? I'm not saying that the behavior should necessarily be the same, but we should have a test that verifies that it does what we want it to do.

EDIT: Looking further down the test it seems that the connection is in `autocommit=false` mode, as we are calling `commit()`. Would you mind verifying whether this comment is correct, and also make sure that we test the behavior for both `autocommit=true|false`.

tests/system/test_dbapi.py
```diff
@@ -344,6 +344,50 @@ def test_DDL_autocommit(shared_instance, dbapi_database):
...
+
+    try:
+        conn = Connection(shared_instance, dbapi_database)
+        want_row = (
```
olavloite, line 352: Can we verify that the connection is in `autocommit=true` mode here? (The comment at the start of the test indicates that the connection should be in autocommit mode)

tests/system/test_dbapi.py
```diff
@@ -344,6 +344,50 @@ def test_DDL_autocommit(shared_instance, dbapi_database):
...
+        assert got_rows == [want_row]
+
+        cur.close()
+        conn.close()
+
+    finally:
+        # Delete table
+        table = dbapi_database.table("Singers")
+        if table.exists():
+            op = dbapi_database.update_ddl(["DROP TABLE Singers"])
```
olavloite, line 387: From a test execution speed perspective, it would be better not to drop the table here. The reason is that each DDL statement that we execute will generate a new schema version, and Cloud Spanner will start throttling DDL statements if we create too many schema versions.

One alternative could be to generate a unique table name during the test execution instead of a static one. Then we can leave the table around, and rely on the test executor to drop the entire database when all tests have been executed.

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -317,6 +317,10 @@ def executemany(self, operation, seq_of_params):
 
         many_result_set = StreamedManyResultSets()
 
+        # For every operation, we've got to ensure that any prior DDL
+        # statements were run.
+        self.connection.run_prior_DDL_statements()
```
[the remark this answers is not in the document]
ankiaga: Yes what you said is true for `autocommit=false`. Here is the flow
- DDL statements as they come are batched
- Whenever an `execute` statement is called for DML/Query, first we run `run_prior_DDL_statements()` which execute all the batched DDL statements in the connection till now
- Then the actual DML/Query statement is executed
- In case of commit also we try to execute all the batched DDL statements

In case of `autocommit=true` DDL statements are executed as they come

tests/system/test_dbapi.py
ankiaga: Sorry for the confusion as I copied the code and forgot updating the comment. 
Ideally there should be 4 tests for execute/executemany and autocommit being True/False combinations
There are no such tests. I added one in this PR so as to verify the issue. Should we add all 4 tests?

tests/system/test_dbapi.py
ankiaga: Same as last comment about wrong comment

tests/system/test_dbapi.py
olavloite: Yes, I think it would be good to add all 4 tests to this PR. That way, we at least know what the current behavior is. Then we can decide whether we want to change any of them, if we see anything strange.

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -317,6 +317,10 @@ def executemany(self, operation, seq_of_params):
 
         many_result_set = StreamedManyResultSets()
 
+        # For every operation, we've got to ensure that any prior DDL
```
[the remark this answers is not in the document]
ankiaga: Done

tests/system/test_dbapi.py
ankiaga: Added tests

tests/system/test_dbapi.py
ankiaga: Done

google/cloud/spanner_dbapi/cursor.py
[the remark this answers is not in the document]
olavloite: Thanks for clarifying

tests/system/test_dbapi.py
```diff
@@ -344,6 +343,133 @@ def test_DDL_autocommit(shared_instance, dbapi_database):
...
+    conn = Connection(shared_instance, dbapi_database)
+    conn.autocommit = True
+    cur = conn.cursor()
+    cur.execute(
+        """
+        CREATE TABLE DdlExecuteAutocommit (
+            SingerId     INT64 NOT NULL,
+            Name    STRING(1024),
+        ) PRIMARY KEY (SingerId)
+    """
```
olavloite, line 359: super nit: indent this the same as the start quotes

tests/system/test_dbapi.py
```diff
@@ -344,6 +343,133 @@ def test_DDL_autocommit(shared_instance, dbapi_database):
...
+    conn.autocommit = True
+    cur = conn.cursor()
+    with pytest.raises(ProgrammingError):
+        cur.executemany(
+            """
+            CREATE TABLE DdlExecuteManyAutocommit (
+                SingerId     INT64 NOT NULL,
+                Name    STRING(1024),
+            ) PRIMARY KEY (SingerId)
+        """,
```
olavloite, line 382: here also: indentation

tests/system/test_dbapi.py
```diff
@@ -344,6 +343,133 @@ def test_DDL_autocommit(shared_instance, dbapi_database):
...
+
+
+def test_ddl_executemany_autocommit_true(shared_instance, dbapi_database):
+    """Check that DDL statement in autocommit mode results in exception for
+    executemany method ."""
+
+    conn = Connection(shared_instance, dbapi_database)
+    conn.autocommit = True
+    cur = conn.cursor()
+    with pytest.raises(ProgrammingError):
```
olavloite, line 375: If I understand it correctly, this will fail in both autocommit=true and autocommit=false modes, right? If so, we should test both here.

tests/system/test_dbapi.py
```diff
@@ -344,6 +343,133 @@ def test_DDL_autocommit(shared_instance, dbapi_database):
...
+        "first-name",
+    )
+    cur = conn.cursor()
+    cur.execute(
+        """
+        CREATE TABLE DdlExecute (
+            SingerId     INT64 NOT NULL,
+            Name    STRING(1024),
+        ) PRIMARY KEY (SingerId)
+    """
```
olavloite, line 408: nit: indentation

tests/system/test_dbapi.py
```diff
@@ -344,6 +343,133 @@ def test_DDL_autocommit(shared_instance, dbapi_database):
...
+    )
+    table = dbapi_database.table("DdlExecute")
+    assert table.exists() is False
+
+    cur.execute(
+        """
+INSERT INTO DdlExecute (SingerId, Name)
```
olavloite, line 415: nit: indentation

tests/system/test_dbapi.py
```diff
@@ -344,6 +343,133 @@ def test_DDL_autocommit(shared_instance, dbapi_database):
...
+        CREATE TABLE DdlExecute (
+            SingerId     INT64 NOT NULL,
+            Name    STRING(1024),
+        ) PRIMARY KEY (SingerId)
+    """
+    )
+    table = dbapi_database.table("DdlExecute")
+    assert table.exists() is False
+
+    cur.execute(
```
olavloite, line 413: This behavior is what has been implemented, but it is really weird. For an outside observer (and especially one that is not familiar with Cloud Spanner and/or this dbapi driver), it will seem like:
1. Cloud Spanner is super fast in executing DDL statements. They finish in like 10 microseconds!
2. Cloud Spanner is super slow in executing DML statements. They take like 10 seconds!

Would you mind writing a small one-pager on this behavior and share it with the rest of the Connectivity team + Adam + .... (maybe more)? I think that we need to discuss whether we want to keep this behavior, or otherwise document it very well. I'm surprised that we haven't had any questions regarding this yet.

tests/system/test_dbapi.py
```diff
@@ -344,6 +343,133 @@ def test_DDL_autocommit(shared_instance, dbapi_database):
...
+            Name    STRING(1024),
+        ) PRIMARY KEY (SingerId)
+    """
+    )
+    table = dbapi_database.table("DdlExecuteMany")
+    assert table.exists() is False
+
+    cur.executemany(
+        """
+INSERT INTO DdlExecuteMany (SingerId, Name)
```
olavloite, line 455: indentation

olavloite (approved):
lgtm, with some nits on formatting

tests/system/test_dbapi.py
ankiaga: Done

tests/system/test_dbapi.py
ankiaga: Done

tests/system/test_dbapi.py
ankiaga: Added a test

tests/system/test_dbapi.py
ankiaga: Done

tests/system/test_dbapi.py
ankiaga: Done

tests/system/test_dbapi.py
ankiaga: Done

tests/system/test_dbapi.py
ankiaga: Sure, will start a doc

PR #1037, opened 2023-11-17
feat: Implementing client side statements in dbapi (starting with commit)

ankiaga:
Implementation to Commit the ongoing transaction whenever execute() method of Cursor is called with a sql query "COMMIT" (or "COMMIT TRANSACTION")
The plan is to later add more client side statements


google/cloud/spanner_dbapi/client_side_statement_executor.py
```diff
@@ -0,0 +1,34 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from google.cloud.spanner_dbapi.parsed_statement import (
+    ParsedStatement,
+    ClientSideStatementType,
+)
+
+
+class StatementExecutor(object):
```
aseering, line 20: Does this benefit from being a class?  (Is `execute()` going to be called more than once?  Will it be called in a context where the connection is not plumbed through but the StatementExecutor object is available?)

There's nothing necessarily wrong with its being a class.  But in Python, if a collection of functions don't have shared state, it's also totally fine to leave them as bare functions, and use the file/module itself as a way of grouping them.  At a quick glance, this class as currently implemented looks to me like it could just be an `execute()` function that takes a `connection` object as an additional argument.  If you have bigger future plans for this class, though, then keeping it as a class makes total sense.

google/cloud/spanner_dbapi/client_side_statement_executor.py
```diff
@@ -0,0 +1,34 @@
...
+    def __init__(self, connection):
+        self.connection = connection
```
aseering, line 22: Is `self.connection` intended to be a public property on this class?

I would naively assume that it'd be a private member.  In Python:

* Members that are intended to be private are, by convention, prefixed with an underscore (`self._connection`).  This leading underscore has no functional impact -- callers can still refer to the field with no problem, the interpreter doesn't treat it as special though some linters do -- but it's very obvious during a code review if someone is accessing a field with a leading underscore.
* Members that really need to be private are prefixed with two underscores (`self.__connection`).  In this case, Python will play namespace games to make it hard (though not impossible if someone tries hard enough) to access the field.  In practice, I can't recall having ever seen this used in production code; people just use the single underscore and accept that breaking the abstraction is an occasionally-necessary antipattern; but it's an option if you need it.

google/cloud/spanner_dbapi/client_side_statement_executor.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def execute(self, parsed_statement: ParsedStatement):
+        """Executes the client side statements by calling the relevant method
+
+        :type parsed_statement: ParsedStatement
+        :param parsed_statement: parsed_statement based on the sql query
+        """
+        if (
```
aseering, line 30: Random Python-ism:  This looks kindof like a switch/case statement.  If you find yourself implementing a switch/case statement with a large number of cases, one way to improve performance (whether it's actually more readable is subjective) is to construct a dict() mapping the possible values to functions implementing the handler logic for each value.  Then do something like `statement_type_actions[parsed_statement.client_side_statement_type](self.connection)`.

But that's certainly overkill for a single type of statement.  Personally, I wouldn't bother unless you have at least a couple dozen statement types to check; maybe more.  For anything smaller, just chain a bunch of `elif`'s.  Just mentioning it for future reference.

The modern way to do switch/case in Python would be [Structural Pattern Matching](https://peps.python.org/pep-0636/).  But this project currently supports Python 3.7+ (released in mid 2018) and Structural Pattern Matching was I believe released in Python 3.10 (released in late 2021).  So taking a dependency on it would be a big update for us.  I believe Python has a 5-year support lifecycle, so this might make more sense to discuss in 2026 or so.

(Tangent, on that note:  I believe Python 3.7 has just recently run past its support cycle?  Google generally frowns upon supporting, testing, or otherwise running software that no longer receives security updates.  So it might be worth starting to think about what the process looks like to bump this package up to a minimum dependency of Python 3.8.  Mechanically, it involves updating `setup.py` in the top-level directory to list a new minimum Python version -- though I'll note that the whole concept of `setup.py` is now deprecated and replaced with `setup.toml`; also I imagine some customers may care about the deprecation of support for a version of Python that they might still be using.)

google/cloud/spanner_dbapi/parsed_statement.py
```diff
@@ -0,0 +1,36 @@
...
+class ClientSideStatementType(Enum):
+    COMMIT = 1
+    BEGIN = 2
+
+
+@dataclass
+class ParsedStatement:
+    statement_type: StatementType
+    query: str
+    client_side_statement_type: ClientSideStatementType = None
```
aseering, line 36: Nice switch to enums and dataclasses!

Maybe of note, `dataclass` was introduced in Python 3.7 (the minimum requirement for this project).  So it should be fine to use, just be wary of using anything newer.  (Often, features that are new to a given Python version are available via third-party packages in prior versions, in which case you can have installations against older versions of Python take on the third-party package as a dependency.)  `Enum` was introduced in Python 3.4.

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -309,19 +315,26 @@ def executemany(self, operation, seq_of_params):
...
             raise ProgrammingError(
                 "Executing DDL statements with executemany() method is not allowed."
             )
 
         # For every operation, we've got to ensure that any prior DDL
         # statements were run.
         self.connection.run_prior_DDL_statements()
+        if parsed_statement.statement_type == StatementType.CLIENT_SIDE:
+            raise ProgrammingError(
+                "Executing ClientSide statements with executemany() method is not allowed."
```
aseering, line 329: Do we explain somewhere to customers what a "ClientSide" statement is?

Is this necessarily a requirement?  Could we just split the batch on the client-side statement, then run everything before as a batch, then do the client-side statement, then run everything after as a batch?  That might be out of scope for this PR, though; just asking if it makes sense in general.

tests/system/test_dbapi.py
```diff
@@ -810,3 +805,33 @@ def test_dml_returning_delete(shared_instance, dbapi_database, autocommit):
     assert cur.fetchone() == (1, "first-name")
     assert cur.rowcount == 1
     conn.commit()
+
+
+def _execute_common_precommit_statements(cursor: Cursor):
```
aseering, line 810: It looks like this code uses pytest?  If so, I would recommend making this a [pytest fixture](https://docs.pytest.org/en/6.2.x/fixture.html#back-to-fixtures).

It looks like this file already has a couple fixtures in it.  I would add another fixture that constructs a connection and cursor; then add this logic as a fixture for initial database data.  The tests above can then depend on both new fixtures; this gets them both a cursor and an initial dataset.

One potential benefit of fixtures is that they support a clean-up phase.  (Just write `yield` in place of `return`, below; then any logic after the `yield` statement is executed as part of fixture tear-down.)  This might, for example, allow you to have a single database that doesn't need to be re-created with every test run (because database creation is expensive), where initial data is populated at the start of the test and any state in the database is cleaned up at the end of the test.

tests/system/test_dbapi.py
```diff
@@ -810,3 +805,33 @@ def test_dml_returning_delete(shared_instance, dbapi_database, autocommit):
...
+    # execute several DML statements within one transaction
+    cursor.execute(
+        """
+        INSERT INTO contacts (contact_id, first_name, last_name, email)
```
aseering, line 814: I forget, does this client support Spanner's generic batch DML API yet?  If so, should these three statements be batched?  It'd be a significant performance win, aside from testing the batch-DML API.

(I'm thinking a bit about test performance:  Python is not a fast language.  In my experience, Python test suites have a tendency to grow and become slower over time to the point where it becomes annoying to wait on the test suite.  So I personally like to put at least a little thought into test performance up front, in an attempt to mitigate this.)

google/cloud/spanner_dbapi/parse_utils.py
```diff
@@ -180,27 +177,29 @@ def classify_stmt(query):
     :type query: str
     :param query: A SQL query.
 
-    :rtype: str
-    :returns: The query type name.
+    :rtype: ParsedStatement
+    :returns: parsed statement attributes.
```
aseering, line 181: Is this function considered to be a public/external interface?  Is it possible that someone externally is calling it and might be depending on this return value?  In which case changing the function's signature would be a breaking change, and should be treated as such (with a major-version bump and other discussion/communication).

(Tangent, although breaking changes should only be made with care in clients, I enjoy that it's actually possible to do so.  Over here in backend-code-land, we can basically never change any API, only add new features, because our service is hosted/multitenant and customers can't pin to an old backend version if they're not ready for an update yet.)

google/cloud/spanner_dbapi/client_side_statement_parser.py
```diff
@@ -0,0 +1,46 @@
...
+
+import re
+
+from google.cloud.spanner_dbapi.parsed_statement import (
+    ParsedStatement,
+    StatementType,
+    ClientSideStatementType,
+)
+
+RE_COMMIT = re.compile(r"^\s*(COMMIT)(TRANSACTION)?", re.IGNORECASE)
```
aseering, line 23: +1 for pre-compiling regexes!  This helps with performance and is general a good thing to do, but people so-often don't remember to do so.

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -85,6 +87,7 @@ def __init__(self, connection):
         self._row_count = _UNSET_COUNT
         self.lastrowid = None
         self.connection = connection
+        self.client_side_statement_executor = StatementExecutor(connection)
```
aseering, line 90: I'd be curious to get a sense for what architecture you have in mind for `StatementExecutor` in the long term.  It seems like it's currently only trying to execute a few specific statement types.  Is the goal to eventually move more of the logic from this file into `StatementExecutor`, and have it be responsible for executing most types of statements?

aseering (approved):
Thanks for putting this together!  No major concerns but a bunch of little comments.  Let me know what you think.

google/cloud/spanner_dbapi/client_side_statement_parser.py
```diff
@@ -0,0 +1,46 @@
...
+
+RE_BEGIN = re.compile(r"^\s*(BEGIN|START)(TRANSACTION)?", re.IGNORECASE)
```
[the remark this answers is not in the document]
ankiaga: Same as above to have it in next PR. I just wanted to setup the base for client side statement in this PR using Commit statement

google/cloud/spanner_dbapi/client_side_statement_executor.py
ankiaga: Yes I was also confused about making it as a class as I dont see any state that it needs to maintain. My thought was that as I will implement different statements I might need attributes other than connection but I am not sure if they would be needed.
I can remove the class and later add if needed

google/cloud/spanner_dbapi/client_side_statement_executor.py
ankiaga: Thanks, I also looked into how to use switch/case in python for current python versions and found there is no native support for it. We might need to refactor this class to have config file similar to java and use reflection so didn't choose to create a dict(). Let me know if this sounds good

google/cloud/spanner_dbapi/cursor.py
ankiaga: No, my plan is not to move any logic from this class to StatementExecutor (part of client_side_statement_executor module). StatementExecutor will have different functions to execute different client side statements only similar to https://github.com/googleapis/java-spanner/blob/main/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ConnectionStatementExecutorImpl.java

tests/system/test_dbapi.py
ankiaga: Cursor has executemany() method which supports batch DML in a limited way where one can only pass single sql query and a list of parameters to run the query with.
I had discussed this with Knut and we have a plan to implement a method in Cursor where multiple sql queries can be passed and run as a batch DML

google/cloud/spanner_dbapi/cursor.py
ankiaga: Let me give a better error message.

Sorry might be I have not understood the comment completely. Do you mean if multiple statements are passed to executemany() (please remember its not possible) and one of them is client side statement?

google/cloud/spanner_dbapi/parse_utils.py
ankiaga: I dont think so but let me try to confirm it

google/cloud/spanner_dbapi/client_side_statement_executor.py
```diff
@@ -0,0 +1,34 @@
...
+            parsed_statement.client_side_statement_type
+            == ClientSideStatementType.COMMIT
+        ):
```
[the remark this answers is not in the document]
aseering: Makes sense.  Yeah, that sounds fine to me.

google/cloud/spanner_dbapi/client_side_statement_executor.py
aseering: Sounds good to me.  I think the current code is fine for now, and any of the proposed changes here could make sense depending on how the code evolves over time.  Reflection is easy and common in Python (IMO even more so than in Java) so a reflection-based approach seems to me like it could also make sense if the preference is to follow the precedent that's already set in Java.

tests/system/test_dbapi.py
aseering: Makes sense.

Sounds like that method doesn't exist yet?  Maybe it would make sense to update this code someday in the future; no need to change it now, though.

google/cloud/spanner_dbapi/cursor.py
aseering: Hm, in that case I wonder if the class should be called `LocalStatementExecutor` or `ClientSideStatementExecutor` or something, to make that more clear?

Python does generally prefer shorter names, though.  So maybe that's overkill.  Anyway it sounds like you're thinking of dropping the class and just using the module.

google/cloud/spanner_dbapi/client_side_statement_executor.py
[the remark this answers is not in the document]
ankiaga: Created b/312624765 for tracking

google/cloud/spanner_dbapi/client_side_statement_executor.py
ankiaga: Created b/312624765 for tracking

google/cloud/spanner_dbapi/client_side_statement_parser.py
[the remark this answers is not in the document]
ankiaga: Created b/312624765 for tracking

google/cloud/spanner_dbapi/client_side_statement_executor.py
ankiaga: Removed the class so not needed

tests/system/test_dbapi.py
ankiaga: This will involve a lot of refactoring so wanted to have a separate PR for it. I have create https://github.com/ankiaga/python-spanner/pull/1 and will raise post merging of this PR. Let me know if this sounds good

tests/system/test_dbapi.py
ankiaga: Yes it doesn't exist yet. Created b/312625672 for tracking

google/cloud/spanner_dbapi/client_side_statement_parser.py
ankiaga: Thanks

google/cloud/spanner_dbapi/cursor.py
ankiaga: Yes dropped the class

google/cloud/spanner_dbapi/client_side_statement_executor.py
ankiaga: Sure, thanks

google/cloud/spanner_dbapi/parsed_statement.py
ankiaga: Sure, thanks

google/cloud/spanner_dbapi/client_side_statement_executor.py
ankiaga: Done

google/cloud/spanner_dbapi/cursor.py
ankiaga: Changed the error message

google/cloud/spanner_dbapi/parse_utils.py
ankiaga: I have left this method as it is and mark it as deprecated and created a new function _classify_stmt to be used

tests/system/test_dbapi.py
aseering: Sure, sounds good to me.

google/cloud/spanner_dbapi/parse_utils.py
aseering: If this function is deprecated, you might consider updating its code to issue a `DeprecationWarning`:

https://docs.python.org/3/library/warnings.html#warnings.warn
https://docs.python.org/3/library/exceptions.html#DeprecationWarning

For future reference (not really applicable to this PR):  If a project commonly deprecates functions (which I don't think this project does yet?, but it might as of some future PR), that project may want to write for itself a little utility decorator `@deprecated` that issues the above warning and then calls the base function.  For both convenience and readability.  Projects that prefer third-party dependencies can get such a decorator via the [`Deprecated` package](https://pypi.org/project/Deprecated/).  Python has a [PEP to make this a builtin](https://peps.python.org/pep-0702/) as of Python 3.13, in probably about a year.

ankiaga: Please see line 182, I have already done that or am I missing anything

aseering: Oh oops, I was looking at the commit view, not the overall PR view.  Sorry!  LGTM.

ankiaga:
Thanks @aseering for taking out time and providing valuable comments. Got to learn a lot from your comments

google/cloud/spanner_dbapi/client_side_statement_executor.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 20203 Google LLC All rights reserved.
```
olavloite, line 1: super-nit: (here and elsewhere) I think it is a bit optimistic to think that any of us will be alive at this copyright year :-)

google/cloud/spanner_dbapi/client_side_statement_parser.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def parse_stmt(query):
```
olavloite, line 28: This also: should it be internal?

google/cloud/spanner_dbapi/client_side_statement_parser.py
```diff
@@ -0,0 +1,46 @@
...
+    :returns: ParsedStatement object.
+    """
+    if RE_COMMIT.match(query):
+        return ParsedStatement(
+            StatementType.CLIENT_SIDE, query, ClientSideStatementType.COMMIT
+        )
+    if RE_BEGIN.match(query):
+        return ParsedStatement(
+            StatementType.CLIENT_SIDE, query, ClientSideStatementType.BEGIN
+        )
```
olavloite, line 45: Would it make more sense to leave this out until we actually do something with it? AFAICT, this PR only implements behavior for COMMIT, and not for BEGIN.

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -309,19 +314,28 @@ def executemany(self, operation, seq_of_params):
...
+        parsed_statement = parse_utils._classify_stmt(operation)
+        if parsed_statement.statement_type == StatementType.DDL:
             raise ProgrammingError(
                 "Executing DDL statements with executemany() method is not allowed."
             )
 
         # For every operation, we've got to ensure that any prior DDL
         # statements were run.
         self.connection.run_prior_DDL_statements()
+        if parsed_statement.statement_type == StatementType.CLIENT_SIDE:
```
olavloite, line 326: I think this code block should be moved to before the `run_prior_DDL_statements`. Now it is (theoretically?) possible that the DDL statements are executed, and then we return an error.

google/cloud/spanner_dbapi/client_side_statement_executor.py
```diff
@@ -0,0 +1,27 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from google.cloud.spanner_dbapi.parsed_statement import (
+    ParsedStatement,
+    ClientSideStatementType,
+)
+
+
+def execute(connection, parsed_statement: ParsedStatement):
```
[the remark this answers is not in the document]
ankiaga: IMO `single leading underscore` means that the function should not be accessed outside the module, which we need to access in this case. Also Intellij is giving the warning if I add an underscore to the function.

As per https://peps.python.org/pep-0008/#public-and-internal-interfaces we dont have this function listed here https://github.com/googleapis/python-spanner/blob/07fbc45156a1b42a5e61c9c4b09923f239729aa8/google/cloud/spanner_dbapi/__init__.py#L61 so it should not be considered a public interface but its just a convention and like everything else in python the function can be accessed as a public interface. Should we have something like an internalapi annotation from Java?

@aseering if you have any suggestions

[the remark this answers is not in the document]
olavloite: What do you mean with 'which we need to access in this case'? Do you mean that we will be accessing this from our Django/SQLAlchemy implementations? Or do you mean from some other module in the dbapi driver?

I would consider our Django and SQLAlchemy implementations 'internal' in this case as well, although maybe technically they are not. The difference here is that:
1. If we publish this function as public to outside users, then we cannot change the signature or behavior of it later.
2. If we mark it as internal, then we are free to change it if necessary, and as we are in control of for example our Django and SQLAlchemy implementations, so we could easily coordinate any changes to this function with the usage there.

From https://peps.python.org/pep-0008/#public-and-internal-interfaces:

> Even with \_\_all\_\_ set appropriately, internal interfaces (packages, modules, classes, functions, attributes or other names) should still be prefixed with a single leading underscore.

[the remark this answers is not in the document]
ankiaga: At discussed offline added a comment stating that its is an internal method that can make backwards-incompatible changes and not using underscores.

google/cloud/spanner_dbapi/client_side_statement_executor.py
ankiaga: Haha. Reduced it by 18180 years :)

google/cloud/spanner_dbapi/client_side_statement_parser.py
ankiaga: Same as last comment

google/cloud/spanner_dbapi/client_side_statement_parser.py
ankiaga: Removed from this PR

google/cloud/spanner_dbapi/cursor.py
ankiaga: Done

olavloite (approved):

google/cloud/spanner_dbapi/client_side_statement_executor.py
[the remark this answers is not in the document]
aseering: > IMO `single leading underscore` means that the function should not be accessed outside the module, which we need to access in this case.

In my experience, leading underscore simply means "if you're not the one who wrote this function, it's not for you to use."  Scoping to any particular namespace is not expected.  So I would not expect this to need to be "public" for that reason.

Surfacing a standard interface between components is a good practice to strive for, though.  Is this the interface that we want to standardize on?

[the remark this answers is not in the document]
ankiaga: Thanks Adam

> Surfacing a standard interface between components is a good practice to strive for, though. Is this the interface that we want to standardize on?

I didn't understand this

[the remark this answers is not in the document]
aseering: I guess what I'm trying to say is, if we want `def execute(...)` to be public, we should spend a bit of time thinking about whether it's the right interface for the job.  Are we ok with supporting this function with this name and this signature in this file, ~forever?  If a bunch of customers find it and decide to use it, are we ok with that?  Or would we wish that we'd spent a bit more time thinking and designing because there's clearly a different signature that would be better here?

(If the function is marked with an underscore, to some extent that communicates that we haven't spent that time reviewing and committing to this as a public interface, so we might discover a mistake and change it in the future.)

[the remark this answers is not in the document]
ankiaga: Thanks for the explanation. I had already commented on the function  that `its is an internal method that can make backwards-incompatible changes` as we were confused about using underscore. But now that you have explained it, I will refactor this function (and few others) to use underscore in a separate PR

google/cloud/spanner_dbapi/parse_utils.py
```diff
@@ -174,12 +177,11 @@
 RE_PYFORMAT = re.compile(r"(%s|%\([^\(\)]+\)s)+", re.DOTALL)
 
 
+@deprecated(reason="This method is deprecated. Use _classify_stmt method")
```
asottile-sentry, line 180: this pulls in two new dependencies (deprecated, wrapt) into this library -- this can be replaced entirely with this inside the function body:

```python
warnings.warn("This method is deprecated.  Use classify_stmt method instead", stacklevel=2)
```

this uses `warnings` directly avoiding the two new dependencies (including one which has a C extension!)

asottile-sentry: would one be open to a PR to do this instead?

olavloite: @ankiaga Would you mind taking a look what would be the best solution here?

asottile-sentry: threw one together: https://github.com/googleapis/python-spanner/pull/1120

PR #1050, opened 2023-12-05
feat: Add support for max commit delay

nginsberg-google:
This feature adds support for max_commit_delays on commits to spanner.


google/cloud/spanner_v1/batch.py
```diff
@@ -146,7 +147,9 @@ def _check_state(self):
         if self.committed is not None:
             raise ValueError("Batch already committed")
 
-    def commit(self, return_commit_stats=False, request_options=None):
+    def commit(
+        self, return_commit_stats=False, request_options=None, max_commit_delay_ms=None
```
harshachinta, line 151: What is the expected input for `max_commit_delay_ms` here? Is it a number?

Instead, I would prefer getting input as `datatime.timedelta`. What do you think?

google/cloud/spanner_v1/batch.py
```diff
@@ -183,11 +189,17 @@ def commit(self, return_commit_stats=False, request_options=None):
         # Request tags are not supported for commit requests.
         request_options.request_tag = None
 
+        max_commit_delay = None
+        if max_commit_delay_ms:
+            max_commit_delay = Duration()
+            max_commit_delay.FromMilliseconds(max_commit_delay_ms)
```
harshachinta, line 195: We can avoid this handwritten logic by getting input `max_commit_delay_ms` as [datetime.timedelta](https://docs.python.org/3/library/datetime.html#timedelta-objects). This way we can directly pass the input to the `CommitRequest` without any additional logic.

Let me know if you have better approach?

google/cloud/spanner_v1/batch.py
```diff
@@ -183,11 +189,17 @@ def commit(self, return_commit_stats=False, request_options=None):
...
+
         request = CommitRequest(
             session=self._session.name,
             mutations=self._mutations,
             single_use_transaction=txn_options,
             return_commit_stats=return_commit_stats,
+            max_commit_delay=max_commit_delay,
```
harshachinta, line 202: What are the validations that backend does on this `max_commit_delay` field?
For ex, does backend accept only data in `ms` etc.

google/cloud/spanner_v1/batch.py
harshachinta: We do something similar for `max_staleness` field. Reference: https://github.com/googleapis/python-spanner/blob/07a02023d588b780ca760280be311e9d221a4a41/google/cloud/spanner_v1/snapshot.py#L691

harshachinta: I feel this way client library changes will be very minimal

nginsberg-google: Good idea - done.

google/cloud/spanner_v1/batch.py
nginsberg-google: done.

google/cloud/spanner_v1/batch.py
nginsberg-google: Any Duration is fine to pass in here. If the duration is outside of the allowed range [0, 500] the backend will complain and pass an error back out. 

I would suggest we don't implement any bounds checks here - I think it's fine to allow the backend to return the error, and then we can have all the bounds checking in one place. If we ever increase the range, I'd like to not have to make client changes.

google/cloud/spanner_v1/database.py
```diff
@@ -733,11 +733,15 @@ def batch(self, request_options=None):
                 (Optional) Common options for the commit request.
                 If a dict is provided, it must be of the same form as the protobuf
                 message :class:`~google.cloud.spanner_v1.types.RequestOptions`.
+        :type max_commit_delay: :class:`datetime.timedelta`
```
harshachinta, line 736: Is there any range of values that are valid for this option? If yes can we mention that range here for customers to understand the right values to pass.
When I look at autogenerated code it mentions the range as `0 and 500 ms`.
https://github.com/googleapis/python-spanner/blob/5b94dac507cebde2025d412da0a82373afdbdaf5/google/cloud/spanner_v1/types/spanner.py#L1445

google/cloud/spanner_v1/database.py
```diff
@@ -796,9 +801,11 @@ def run_in_transaction(self, func, *args, **kw):
 
         :type kw: dict
         :param kw: (Optional) keyword arguments to be passed to ``func``.
-                   If passed, "timeout_secs" will be removed and used to
+                   If passed,
+                   "timeout_secs" will be removed and used to
                    override the default retry timeout which defines maximum timestamp
                    to continue retrying the transaction.
+                   "max_commit_delay" will be removed and used to set the max_commit_delay for the request.
```
harshachinta, line 808: Same as above comment. Can we mention the allowed range of max_commit_delay as this is public interface through which customers can pass this option.

harshachinta (commented):
LGTM. Just one minor comment on adding range in the docs comment.

google/cloud/spanner_v1/database.py
nginsberg-google: Done.

google/cloud/spanner_v1/database.py
nginsberg-google: Done.

nginsberg-google:
I think this should be ready to merge. What are the next steps here? Thanks!

harshachinta (approved):

PR #1055, opened 2023-12-12
feat: Implementation for batch dml in dbapi

ankiaga:


google/cloud/spanner_dbapi/connection.py
```diff
@@ -196,6 +203,24 @@ def read_only(self, value):
             )
         self._read_only = value
 
+    @property
+    def batch_mode(self):
```
olavloite, line 207: If I understand it correctly, giving these names that do not start with an underscore will make them part of the public API. In that case, we should document them and also add validations to verify that they are only called with valid arguments. (But probably we should make them internal instead)

google/cloud/spanner_dbapi/connection.py
```diff
@@ -196,6 +203,24 @@ def read_only(self, value):
...
+        return self._batch_mode
+
+    @batch_mode.setter
+    def batch_mode(self, value):
+        """`batch_mode` flag setter.
+
+        Args:
+            value (BatchMode)
+        """
+        self._batch_mode = value
```
olavloite, line 222: What happens if an external uses calls this function when the connection is already in the middle of a different type of batch (e.g. it is now in a DML batch, and it is called to set it to DDL batch)?

tests/system/test_dbapi.py
```diff
@@ -425,6 +425,67 @@ def test_read_timestamp_client_side_autocommit(self):
...
+                    (6, "first-name-6", "last-name-6", "test.email@domen.ru"),
+                ]
+            ).sort()
+        )
+
+        # Test starting another dml batch in same connection post commit works
+        self._cursor.execute("start batch dml")
+        self._insert_row(7)
+        self._insert_row(8)
+        self._cursor.execute("run batch")
```
olavloite, line 471: It's probably difficult, but: Do we have any way that we could verify that the `run batch` call really sends an `ExecuteBatchDml` request, and does not execute multiple `ExecuteSql` requests sequentially?

tests/unit/spanner_dbapi/test_batch_dml_executor.py
```diff
@@ -0,0 +1,40 @@
+import unittest
```
olavloite, line 1: nit: Add copyright header

tests/system/test_dbapi.py
```diff
@@ -425,6 +425,67 @@ def test_read_timestamp_client_side_autocommit(self):
...
+        assert len(self._cursor.fetchall()) == 9
+
+    def _insert_row(self, i):
+        self._cursor.execute(
+            f"""
+            INSERT INTO contacts (contact_id, first_name, last_name, email)
+            VALUES ({i}, 'first-name-{i}', 'last-name-{i}', 'test.email@domen.ru')
+            """
+        )
+
```
olavloite, line 488: We should also have tests for the unhappy path of using batch DML:
1. What happens if the batch contains an invalid statement as the first statement?
2. What happens if the batch contains an invalid statement in the middle?
3. What happens if the batch contains an invalid statement at the end?
4. Can we test and verify that the retry logic for Batch DML works as expected?

google/cloud/spanner_dbapi/batch_dml_executor.py
```diff
@@ -0,0 +1,119 @@
...
+
+    :type cursor: Cursor
+    :param cursor: Database Cursor object
+
+    :type statements: List[Statement]
+    :param statements: list of statements to execute in batch
+    """
+    from google.cloud.spanner_dbapi import OperationalError
+
+    connection = cursor.connection
```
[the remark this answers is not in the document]
ankiaga: This is the method outside of the class so we cant

google/cloud/spanner_dbapi/batch_dml_executor.py
```diff
@@ -0,0 +1,119 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
ankiaga: Done

google/cloud/spanner_dbapi/batch_dml_executor.py
```diff
@@ -0,0 +1,119 @@
...
+         params
+        """
+        from google.cloud.spanner_dbapi import ProgrammingError
+
+        if (
+            parsed_statement.statement_type != StatementType.UPDATE
+            and parsed_statement.statement_type != StatementType.INSERT
+        ):
+            raise ProgrammingError(
+                "Only DML statements are allowed in batch " "DML mode."
```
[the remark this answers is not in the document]
ankiaga: Fixed

google/cloud/spanner_dbapi/client_side_statement_parser.py
```diff
@@ -29,6 +30,8 @@
 RE_SHOW_READ_TIMESTAMP = re.compile(
     r"^\s*(SHOW)\s+(VARIABLE)\s+(READ_TIMESTAMP)", re.IGNORECASE
 )
+RE_START_BATCH_DML = re.compile(r"^\s*(START)\s+(BATCH)\s+(DML)", re.IGNORECASE)
+RE_RUN_BATCH = re.compile(r"^\s*(RUN)\s+(BATCH)", re.IGNORECASE)
```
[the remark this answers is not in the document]
ankiaga: Added

google/cloud/spanner_dbapi/connection.py
ankiaga: Removed the property

google/cloud/spanner_dbapi/connection.py
ankiaga: Removed this property

tests/unit/spanner_dbapi/test_batch_dml_executor.py
ankiaga: Done

tests/system/test_dbapi.py
ankiaga: Added test for invalid statements.

There are issues with the retry logic which I am fixing in the next PR

google/cloud/spanner_dbapi/batch_dml_executor.py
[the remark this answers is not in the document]
olavloite: Ah, see my excellent Python skills shine in practice 🤣

google/cloud/spanner_dbapi/client_side_statement_parser.py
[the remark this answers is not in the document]
olavloite: Thanks

tests/system/test_dbapi.py
olavloite: Yes, I meant automated. That would then also guard us against any regressions, for example if something changes in the underlying client library, which cause this to use multiple `ExecuteSql` RPCs instead of one `ExecuteBatchDml` RPC.

tests/system/test_dbapi.py
```diff
@@ -425,6 +425,125 @@ def test_read_timestamp_client_side_autocommit(self):
...
+        self._insert_row(1)
+        self._insert_row(2)
+        self._cursor.execute(
+            """
+            INSERT INTO unknown_table (contact_id, first_name, last_name, email)
+            VALUES (2, 'first-name', 'last-name', 'test.email@domen.ru')
+            """
+        )
+        with pytest.raises(OperationalError):
+            self._cursor.execute("run batch")
```
olavloite, line 537: I would have expected this to return a specific batch error that:
1. Contains the error code, message etc.
2. And contains the update counts of the statements that did succeed.

See for example this for the JDBC driver: https://github.com/googleapis/java-spanner-jdbc/blob/1f89f78c37b9e118e2c0cbc7f56d3eb1d5745863/src/test/java/com/google/cloud/spanner/jdbc/it/ITJdbcPreparedStatementTest.java#L797

tests/unit/spanner_dbapi/test_cursor.py
```diff
@@ -248,13 +252,13 @@ def test_execute_statement(self):
...
         with mock.patch(
             "google.cloud.spanner_dbapi.parse_utils.classify_statement",
-            return_value=ParsedStatement(StatementType.UPDATE, sql),
+            return_value=ParsedStatement(StatementType.UPDATE, Statement(sql)),
         ):
             cursor.connection._database = mock_db = mock.MagicMock()
             mock_db.run_in_transaction = mock_run_in = mock.MagicMock()
             cursor.execute(sql="sql")
             mock_run_in.assert_called_once_with(
-                cursor._do_execute_update_in_autocommit, "sql WHERE 1=1", None
```
olavloite, line 257: Why did this change in this PR?

tests/system/test_dbapi.py
olavloite: Ack

olavloite (approved):
Looks generally good to me, but with one remaining question/nit on the exception that is raised when a batch fails. The exception should also contain the update counts of the statements that did succeed. Are we sure that it does?

tests/unit/spanner_dbapi/test_cursor.py
ankiaga: The logic to add WHERE clause has moved to `parse_utils.classify_statement` now which we are mocking here so its same as what is returned as ler line 255

tests/system/test_dbapi.py
ankiaga: As discussed we need to create a custom exception class, so will take it in follow up PR

google/cloud/spanner_dbapi/batch_dml_executor.py
```diff
@@ -0,0 +1,119 @@
...
+                res_checksum.consume_result(status.code)
+                if not retried:
+                    connection._statements.append((statements, res_checksum))
+                cursor._row_count = sum([max(val, 0) for val in res])
+
+                if status.code == ABORTED:
+                    connection._transaction = None
+                    raise Aborted(status.message)
+                elif status.code != OK:
+                    raise OperationalError(status.message)
```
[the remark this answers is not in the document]
ankiaga: Will take it in a follow up PR

tests/system/test_dbapi.py
ankiaga: This needs some investigation, so will take it in a follow up PR

PR #1056, opened 2023-12-15
feat: Fixing and refactoring transaction retry logic in dbapi. Also adding interceptors support for testing

ankiaga:
We found couple of bugs in the existing retry logic, the major one being [b/315807641](https://buganizer.corp.google.com/issues/315807641)
Other one being assuming tuple to be a list (line 316 in old connection.py) and comparing which would never work. 
These issues were not caught as there are no proper tests for these. 

So this PR is fixing those issues. Also the current code was not very easy to understand so this PR also refactor the retry logic from connection.py to transaction_helper.py along with the fixes. Tests from test_connection.py have also moved to test_transaction_helper.py along with many new tests (mainly on batch statements) added 

Also added interceptors support so we can test retry logic in system tests. Added system tests for the same


tests/system/test_dbapi.py
```diff
@@ -691,32 +694,6 @@ def test_rollback_on_connection_closing(self, shared_instance, dbapi_database):
         cursor.close()
         conn.close()
 
-    def test_results_checksum(self):
```
olavloite, line 694: I assume that this test is being removed because it was wrong, but would you mind explaining what was wrong with it?

tests/unit/spanner_dbapi/test_connection.py
```diff
@@ -595,244 +518,30 @@ def test_rollback_clears_statements(self, mock_transaction):
...
         mock_transaction.commit.side_effect = [Aborted("Aborted"), None]
-        run_mock = connection.run_statement = mock.Mock()
-        run_mock.return_value = ([row], ResultsChecksum())
+        run_mock = connection._transaction_helper = mock.Mock()
 
         connection.commit()
 
-        run_mock.assert_called_with(statement, retried=True)
-
-    @mock.patch("google.cloud.spanner_v1.Client")
```
olavloite, line 685: Are all these tests moved to a different file? It makes it very hard to compare what changed and what remained the same. Could you give a TLDR of that?

google/cloud/spanner_dbapi/transaction_helper.py
```diff
@@ -0,0 +1,105 @@
...
+        attempt = 0
+        while True:
+            self._connection._spanner_transaction_started = False
+            attempt += 1
+            if attempt > MAX_INTERNAL_RETRIES:
+                raise
+
+            try:
+                self._rerun_previous_statements()
+                self._rerun_previous_batch_statements()
```
olavloite, line 62: This means that the retry will always first run all 'single' statements and then all batched statements. That can be a completely different order than in the original transaction, which makes the retry do something different than the original attempt.

Consider for example if the original attempt executed this:

```sql
start batch dml;
insert into foo (id) values (1);
insert into foo (id) values (2);
run batch;
select * from foo;
```

The retry will execute:

```
select * from foo;
start batch dml;
insert into foo (id) values (1);
insert into foo (id) values (2);
run batch;
```

That would cause the checksum of `select * from foo` to be different. The fact that no test cases seem to fail indicates that we do not have enough test coverage, as this should then certainly have failed.

google/cloud/spanner_dbapi/transaction_helper.py
```diff
@@ -0,0 +1,105 @@
...
+            status, res = transaction.batch_update(statements_tuple)
+            if status.code == ABORTED:
+                raise Aborted(status.details)
+
+            retried_checksum = _get_batch_statements_result_checksum(res, status.code)
+            _compare_checksums(original_checksum, retried_checksum)
+
+    def _rerun_previous_statements(self):
+        for single_statement in self._single_statements:
+            res_iter = self._connection.run_statement(single_statement)
```
olavloite, line 87: If this statement fails (for example if the statement returns `syntax error`), then the entire retry will fail with syntax error. That is not correct, and would also be very confusing to the user, as they would get a `syntax error` message for a random statement that triggered the retry.

It is important to consider that the list of statements in the transaction can contain statements that will never work, for example because they contain a syntax error (or any other error). If the user caught those exceptions during the initial transaction attempt, then they will (should!) be in the list of statements, and should also be retried. During the retry, it should be verified that the statement returned the same error as during the initial attempt.

In Java we for example have the class `FailedUpdate` that is recorded as the executed statement if a DML statement fails during the initial attempt. During the retry, the connection checks that the same error code is returned as during the initial attempt: https://github.com/googleapis/java-spanner/blob/c56c56981b00ee059ec4dbad2154f24ebc12f8b9/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/FailedUpdate.java#L32

olavloite (commented):
Would you mind adding a description to the PR what the faulty behavior was that you found, and how you fixed it? The PR is relatively big and complex, so it would be good to have some context.

Also: After an initial pass, it is very clear that this feature does not have a good enough test coverage. There are some big problems with the current implementation (probably much of it was already there), and the fact that none of the tests have caught that is a clear sign that it is not tested well enough.

tests/unit/spanner_dbapi/test_connection.py
ankiaga: Yes they have been moved to test_transaction_helper.py file along with the source code refactored. I have added more unit tests in test_transaction_helper.py mainly for batch statements which were not present earlier

google/cloud/spanner_dbapi/transaction_helper.py
ankiaga: Ohh didn't thought about the order, thanks for catching this. Let me try to fix this

google/cloud/spanner_dbapi/transaction_helper.py
ankiaga: > It is important to consider that the list of statements in the transaction can contain statements that will never work, for example because they contain a syntax error (or any other error). If the user caught those exceptions during the initial transaction attempt, then they will (should!) be in the list of statements, and should also be retried. During the retry, it should be verified that the statement returned the same error as during the initial attempt.

I am not adding the failed statement with exception being caught by user to be retried. Shouldn't that be ok as the user already caught the exception for those statements?

tests/system/test_dbapi.py
ankiaga: cursor doesn't have a _checksum field now. Also this is more of a unit test which is already added

google/cloud/spanner_dbapi/transaction_helper.py
olavloite: Statements that return an error are part of the original attempt of the transaction, and must also be part of the retry attempt. The reason for this is that an error during a transaction can be an indication of what is and is not present in the database.

Consider for example a very naive implementation of insert-or-update (in pseudo-code):

```sql
try:
  insert into foo (id, value) values (1, 'One');
  return;
rescue AlreadyExistsError:
  update foo set value='One' where id=1;
```

During a retry we should verify that `insert into foo` returns the same result as during the initial attempt, as that result can be used to control the further program flow. A 'result' in this case can also be a specific error.

google/cloud/spanner_dbapi/batch_dml_executor.py
```diff
@@ -90,28 +91,24 @@ def run_batch_dml(cursor: "Cursor", statements: List[Statement]):
...
         cursor._row_count = sum([max(val, 0) for val in res])
     else:
-        retried = False
         while True:
             try:
                 transaction = connection.transaction_checkout()
                 status, res = transaction.batch_update(statements_tuple)
-                many_result_set.add_iter(res)
-                res_checksum = ResultsChecksum()
-                res_checksum.consume_result(res)
```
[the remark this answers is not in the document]
ankiaga: Please see `add_execute_statement_for_retry` and `add_fetch_statement_for_retry` in transaction_helper.py

google/cloud/spanner_dbapi/connection.py
```diff
@@ -443,11 +365,11 @@ def commit(self):
...
         except Aborted:
-            self.retry_transaction()
+            self._transaction_helper.retry_transaction()
             self.commit()
         finally:
             self._release_session()
-            self._statements = []
+            self._transaction_helper.reset()
             self._transaction_begin_marked = False
             self._spanner_transaction_started = False
```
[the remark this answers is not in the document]
ankiaga: Done

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -219,7 +220,7 @@ def _batch_DDLs(self, sql):
         self.connection._ddl_statements.extend(statements)
 
     @check_not_closed
-    def execute(self, sql, args=None):
+    def execute(self, sql, args=None, call_from_execute_many=False):
```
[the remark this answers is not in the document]
ankiaga: Done

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -233,68 +234,74 @@ def execute(self, sql, args=None):
...
                 not self.connection._client_transaction_started
-                and parsed_statement.statement_type == StatementType.QUERY
+                and self._parsed_statement.statement_type == StatementType.QUERY
             ):
                 self._handle_DQL(sql, args or None)
-            elif parsed_statement.statement_type == StatementType.DDL:
+            elif self._parsed_statement.statement_type == StatementType.DDL:
                 self._batch_DDLs(sql)
                 if not self.connection._client_transaction_started:
                     self.connection.run_prior_DDL_statements()
```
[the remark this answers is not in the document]
ankiaga: Thats correct

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -405,25 +405,49 @@ def fetchmany(self, size=None):
...
+                        if self._in_retry_mode:
+                            raise
+                        else:
+                            self.transaction_helper.retry_transaction()
+        except Exception as e:
+            exception = e
+            raise
+        finally:
+            if not self._in_retry_mode:
+                self.transaction_helper.add_fetch_statement_for_retry(
```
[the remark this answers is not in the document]
ankiaga: No we would just record 1 statement. `_last_statement_result_details` in transaction_helper.py would store the last statement and while adding a new statement using `add_fetch_statement_for_retry` we would check if the last statement is also of fetch type then we only modify and dont add a new statement

google/cloud/spanner_dbapi/transaction_helper.py
```diff
@@ -0,0 +1,230 @@
...
+        result_details = None
+        if exception is not None:
+            result_details = exception
+        elif (
+            parsed_statement.statement_type == StatementType.INSERT
+            or parsed_statement.statement_type == StatementType.UPDATE
+            or parsed_statement.client_side_statement_type
+            == ClientSideStatementType.RUN_BATCH
+        ):
+            result_details = rowcount
```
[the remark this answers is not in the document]
ankiaga: yes they would be considered equal as we are just maintaining total row count in cursor. Let me see how to fix this

google/cloud/spanner_dbapi/transaction_helper.py
```diff
@@ -0,0 +1,230 @@
...
+        self._connection._transaction_begin_marked = False
+        self._connection._batch_mode = BatchMode.NONE
+
+    def reset(self):
+        """
+        Resets the state of the class when the ongoing transaction is committed
+        or aborted
+        """
+        self._statement_result_details_list = []
+        self._last_statement_result_details = None
```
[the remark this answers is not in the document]
ankiaga: Yes it wont work. Thanks for catching this. Let me try to fix

google/cloud/spanner_dbapi/transaction_helper.py
```diff
@@ -0,0 +1,230 @@
...
+        else:
+            result_details = _get_statement_result_checksum(rows)
+            if is_fetch_all:
+                self._last_statement_result_details = FetchStatement(
+                    statement_type=CursorStatementType.FETCH_ALL,
+                    result_details=result_details,
+                )
+            else:
+                self._last_statement_result_details = FetchStatement(
+                    statement_type=CursorStatementType.FETCH_MANY,
```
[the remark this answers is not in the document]
ankiaga: No, we are just storing `CursorStatementType.FETCH_MANY` in the statements to retry. If there are 100 fetchOne statements in the transaction, we would just store one 1 statement with statement_type as CursorStatementType.FETCH_MANY and size as 100.

I know its a bit hacky and less comprehensible but couldn't find a better way. I can add comments in code to make it clear

google/cloud/spanner_dbapi/transaction_helper.py
```diff
@@ -0,0 +1,230 @@
...
+        self._statement_result_details_list = []
+        self._last_statement_result_details = None
+
+    def add_fetch_statement_for_retry(self, rows, exception, is_fetch_all):
+        if not self._connection._client_transaction_started:
+            return
+        if (
+            self._last_statement_result_details is not None
+            and self._last_statement_result_details.statement_type
+            == CursorStatementType.FETCH_MANY
```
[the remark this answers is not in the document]
ankiaga: Please see the above comment where we are storing fetchone statement as CursorStatementType.FETCH_MANY. So in case of a new fetch statement we are checking whether the last statement is of type fetchOne or fetchMany and if so then we are just appending the modifying the last statement rather than adding  a new statement in the list to retry

google/cloud/spanner_dbapi/transaction_helper.py
```diff
@@ -0,0 +1,230 @@
...
+                    except Exception as ex:
+                        if (
+                            type(statement_result_details.result_details)
+                            is not type(ex)
+                            or ex.args != statement_result_details.result_details.args
+                        ):
+                            raise RetryAborted(RETRY_ABORTED_ERROR, ex)
+                return
+            except Aborted:
+                delay = 2**attempt + random.random()
```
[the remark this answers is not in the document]
ankiaga: Done

google/cloud/spanner_dbapi/transaction_helper.py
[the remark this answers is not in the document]
ankiaga: Made changes to fix this by also storing cursor object. Also add a system test `test_abort_retry_multiple_cursors` to verify this.
As discussed also trying to understand how its implemented in Java and if we can reuse the logic to make it easier

tests/unit/spanner_dbapi/test_connection.py
```diff
@@ -138,10 +138,6 @@ def test_read_only_not_retried(self):
             )
         )
 
-        cursor.fetchone()
```
olavloite, line 141: Why are these removed?

google/cloud/spanner_v1/testing/database_test.py
```diff
@@ -0,0 +1,112 @@
...
+        self._method_count_interceptor = MethodCountInterceptor()
+        self._method_abort_interceptor = MethodAbortInterceptor()
+        self._interceptors = [
+            self._method_count_interceptor,
+            self._method_abort_interceptor,
+        ]
+
+    @property
+    def spanner_api(self):
+        """Helper for session-related API calls."""
```
olavloite, line 70: nit: Maybe update this comment, as this client is not necessarily only for session-related API calls. It could for example also be used to manually execute a `Rollback` request.

tests/system/test_dbapi.py
```diff
@@ -544,6 +570,245 @@ def _insert_row(self, i):
...
+            row_data,
+        )
+        # called 2 times and as this would make 3 execute streaming sql calls
+        # so total 6 calls
+        self._cursor.executemany(
+            """SELECT * FROM contacts WHERE contact_id = %s""",
+            ((1,), (2,), (3,)),
+        )
+        self._cursor.fetchone()
+        self._cursor.fetchmany(2)
```
olavloite, line 611: Do we have a test that shows that this works as expected if the following happens:
1. `executemany` is called with a query and 3 parameter values, and each of the queries returns 1 row (so the same as in this test).
2. `fetchone` is called. That returns the first row.
3. `fetchmany(2)` is called. The first `ExecuteStreamingSql` RPC succeeds. The second `ExecuteStreamingSql` RPC fails with `Aborted`

tests/system/test_dbapi.py
```diff
@@ -544,6 +570,245 @@ def _insert_row(self, i):
...
+    @pytest.mark.noautofixt
+    def test_abort_retry_multiple_cursors(self, shared_instance, dbapi_database):
+        """Test that retry works when multiple cursors are involved in the transaction."""
+
+        try:
+            conn = Connection(shared_instance, dbapi_database)
+            cur = conn.cursor()
+            cur.execute(
+                """
+                CREATE TABLE Singers (
```
olavloite, line 754: nit: Creating an extra table just for this test is relatively expensive. Could we instead just use one table, and instead separate the cursors by letting cursor1 select the rows with id (1, 2), and cursor2 select the rows with id (3 4)?

tests/system/test_dbapi.py
```diff
@@ -544,6 +570,245 @@ def _insert_row(self, i):
...
+            VALUES (2, 'first-name')
+            """
+            )
+            conn.commit()
+
+            cur1 = conn.cursor()
+            cur1.execute("SELECT * FROM contacts")
+            cur2 = conn.cursor()
+            cur2.execute("SELECT * FROM Singers")
+            row1 = cur1.fetchone()
```
olavloite, line 790: Note (not for this pull request, but something that we should look into in a future pull request): `cursor.fetchone`, `cursor.fetchmany` and `cursor.fetchall` can all also return an `Aborted` error, even when we are halfway through the results. We should verify that that situation is also handled gracefully. That means that either of the following should happen:
1. The method just propagates the `Aborted` error to the application and the transaction fails.
2. OR: The cursor continues from where it was.

So assume the following query

```
select *
from singers
```

which returns:

```
(1, 'FirstName1', 'LastName1')
(2, 'FirstName2', 'LastName2')
(3, 'FirstName3', 'LastName3')
```

The application then calls:

```
// Returns row 1
cursor.fetchone()

// Raises an Aborted error.
This should either be propagated to the application, OR internally retried and then return row 2.
cursor.fetchone()

// Returns row 3 (or is never reached if the previous `fetchone` just propagated the Aborted error.
cursor.fetchone() 
```

The reason that `fetchone` (and other fetch methods) can return `Aborted` is that `ExecuteStreamingSql` returns a stream of `PartialResultSet`. This stream can break __at any moment__ with an `Aborted` error if the transaction has been aborted by Cloud Spanner.

google/cloud/spanner_dbapi/transaction_helper.py
```diff
@@ -61,67 +61,106 @@ def reset(self):
...
+            result_type = ResultType.EXCEPTION
             result_details = exception
-        elif (
-            parsed_statement.statement_type == StatementType.INSERT
-            or parsed_statement.statement_type == StatementType.UPDATE
-            or parsed_statement.client_side_statement_type
-            == ClientSideStatementType.RUN_BATCH
-        ):
-            result_details = rowcount
+        elif cursor._batch_dml_rows_count is not None:
```
olavloite, line 147: One very special thing about batch DML is that it can contain __both__ update counts __and__ an exception. If you for example execute a batch DML statement like this:

```sql
start batch dml;
insert into test (id, value) values (1, 'One');
insert into test (id, value) values (2, 'Two');
insert into test (id, value) values (1, 'Three');
```

Then the result will be:
1. Update counts: [1, 1]
2. Exception: `AlreadyExists`

In other words: The first two statements succeeded, the third failed. The current retry logic does not appear to take this into account.

Note that the following could in theory also happen:

```sql
insert into test (id, value) values (1, 'One');
insert into test (id, value) values (2, 'Two');
insert into test (id, value) values (3, 'Three');
```

Result:
1. Update counts: [1, 1]
2. Exception: `Aborted`

This means that the transaction was aborted after executing the first two DML statements.

olavloite:
My general impression is that we need a more extensive test suite for transaction retries. Tests that I would like to see are:

1. Tests that show that a retry fails if the underlying data has changed.
2. Tests that show that a retry succeeds if the underlying data has remained unchanged. The test should include a query that selects data from a table that uses all data types that are supported by Cloud Spanner. This will show that the checksum calculation works correctly for all data types.
3. Tests that verify that the same retry as in test 2 fails if one of the columns in the result has changed. This also verifies that the checksum calculation is 'correct' for all data types in the sense that it produces a different outcome if a value has changed.
4. (Potentially in a separate PR) A test that shows that a Batch DML that contains N elements and that successfully executes M=N-X of the DML statements and returns an error for statement number M+1, is correctly handled. That is: During a retry, both the M update counts + the error must be the same for the retry to be successful, and otherwise it should fail.
5. (Potentially in a separate PR) A test that shows that a `fetchxxx` call that returns an `Aborted` error is correctly handled. 'Correctly' in this case means either:
5.1. Propagates the `Aborted` error to the user without retrying
5.2. OR: Internally retries the transaction and continues the cursor from the same point.

google/cloud/spanner_dbapi/transaction_helper.py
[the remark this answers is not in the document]
olavloite: I like the solution. It makes the actual retry logic simpler, as we can combine `fetchone` and `fetchmany` into one block, and we don't have to maintain a list of `fetchxxx` statements in the retry helper.

google/cloud/spanner_dbapi/transaction_helper.py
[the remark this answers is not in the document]
ankiaga: Fixed by maintaining state of the result (using variable `_batch_dml_rows_count`)

tests/system/test_dbapi.py
ankiaga: We dont have a test when fetch calls results in Aborted error as I am not able to simulate that like in this case how can we make sure that there are multiple `ExecuteStreamingSql` calls?

tests/system/test_dbapi.py
ankiaga: Thanks, done

tests/system/test_dbapi.py
ankiaga: The code takes care of retry when Aborted exception happens during any of the fetch call but I have not been able to add tests for it

google/cloud/spanner_v1/testing/database_test.py
ankiaga: As discussed not needed

google/cloud/spanner_dbapi/transaction_helper.py
ankiaga: As discussed created a bug https://buganizer.corp.google.com/issues/319404848 for this and will take it later as part of separate PR

ankiaga:
> My general impression is that we need a more extensive test suite for transaction retries. Tests that I would like to see are:
> 
> 1. Tests that show that a retry fails if the underlying data has changed.
> 2. Tests that show that a retry succeeds if the underlying data has remained unchanged. The test should include a query that selects data from a table that uses all data types that are supported by Cloud Spanner. This will show that the checksum calculation works correctly for all data types.
> 3. Tests that verify that the same retry as in test 2 fails if one of the columns in the result has changed. This also verifies that the checksum calculation is 'correct' for all data types in the sense that it produces a different outcome if a value has changed.
> 4. (Potentially in a separate PR) A test that shows that a Batch DML that contains N elements and that successfully executes M=N-X of the DML statements and returns an error for statement number M+1, is correctly handled. That is: During a retry, both the M update counts + the error must be the same for the retry to be successful, and otherwise it should fail.
> 5. (Potentially in a separate PR) A test that shows that a `fetchxxx` call that returns an `Aborted` error is correctly handled. 'Correctly' in this case means either:
>    5.1. Propagates the `Aborted` error to the user without retrying
>    5.2. OR: Internally retries the transaction and continues the cursor from the same point.

Thanks for spending time and figuring out all tests missing here. As discussed created a bug https://buganizer.corp.google.com/issues/319404848 for this

tests/system/test_dbapi.py
ankiaga: Will add the tests later. Create a bug  for this https://buganizer.corp.google.com/issues/319404848

tests/system/test_dbapi.py
ankiaga: As discussed created a bug https://buganizer.corp.google.com/issues/319404848 for this and will take it later as part of separate PR

olavloite: ACK

tests/system/test_dbapi.py
olavloite: ACK

google/cloud/spanner_dbapi/transaction_helper.py
olavloite: ACK

olavloite (approved):

olavloite:
Thanks for your patience and perseverance on this 👍

PR #1058, opened 2023-12-15
 feat: Adding support for grpc client interceptors

ankiaga:
We want to intercept that ExecuteBatchDml api is being called in the system test for batch dml


google/cloud/spanner_v1/services/spanner/client.py
```diff
@@ -28,6 +28,7 @@
     Type,
```
parthea, line 28: Please see the [history](https://github.com/googleapis/python-spanner/commits/main/google/cloud/spanner_v1/services/spanner) of directory `google/cloud/spanner_v1/services/spanner` which shows that this file is autogenerated. Please propose changes to [GAPIC generator python](https://github.com/googleapis/gapic-generator-python) instead. 

Any handwritten changes to files in directory `google/cloud/spanner_v1/services/spanner` will be reverted in the next automated owlbot PR.

google/cloud/spanner_v1/testing/interceptors.py
```diff
@@ -0,0 +1,41 @@
+from collections import defaultdict
```
parthea, line 1: Is this file intended to be public? If it is used for testing purposes, should it be under the `tests` directory

setup.py
```diff
@@ -44,6 +44,8 @@
     "proto-plus >= 1.22.2, <2.0.0dev; python_version>='3.11'",
     "protobuf>=3.19.5,<5.0.0dev,!=3.20.0,!=3.20.1,!=4.21.0,!=4.21.1,!=4.21.2,!=4.21.3,!=4.21.4,!=4.21.5",
     "deprecated >= 1.2.14",
+    "grpc-interceptor",
+    "grpcio>=1.60.0rc1",
```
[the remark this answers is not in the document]
parthea: Please can you add a comment to clarify the reason for requiring a newer version of `grpc`? 

When new dependencies are added, this file also needs to be updated to ensure that we run tests against the minimum supported version.
https://github.com/googleapis/python-spanner/blob/main/testing/constraints-3.7.txt

PR #1065, opened 2023-12-26
docs: samples and tests for auto-generated createDatabase and createInstance APIs.

rahul2393:
### Notes

* Adds samples and tests for auto-generated createDatabase and createInstance APIs.
* The sample code style is mostly consistent with our existing samples.
* A new file path has been intentionally introduced so that the samples don't get updated immediately on public documentation  - https://cloud.google.com/spanner/docs/create-manage-databases.


samples/generated/snippets.py
```diff
@@ -0,0 +1,104 @@
...
+from google.cloud import spanner
+from google.cloud.spanner_admin_instance_v1.types import spanner_instance_admin
+
+OPERATION_TIMEOUT_SECONDS = 240
+
+
+# [START spanner_create_instance]
+def create_instance(instance_id):
+    """Creates an instance."""
+    spanner_client = spanner.Client()
```
harshachinta, line 33: Why are we still using handwritten Client? I think this should use the autogenerated instance client directly like below,
```
from google.cloud import spanner_admin_instance_v1

instance_client = spanner_admin_instance_v1.InstanceAdminClient()
```

samples/generated/snippets.py
```diff
@@ -0,0 +1,104 @@
...
+
+    config_name = "{}/instanceConfigs/regional-us-central1".format(
+        spanner_client.project_name
+    )
+
+    operation = spanner_client.instance_admin_api.create_instance(
```
harshachinta, line 39: We should not rely on `instance_admin_api` because we should remove this once we ask customers to start using autogenerated admin clients.

samples/generated/snippets_test.py
```diff
@@ -0,0 +1,207 @@
+# Copyright 2024 Google, Inc.
```
harshachinta, line 1: Can you correct this formatting? Why is there an empty line alternatively?

samples/generated/snippets.py
rahul2393: > instance_admin_api
No, this is not true, we will never remove spanner.Client support.

samples/generated/snippets.py
rahul2393: >    spanner_client = spanner.Client()

spanner.Client will never be removed since the object will be used for data queries also, only the admin level methods from client will be removed.

samples/generated/snippets.py
rahul2393: We will just ask custmer to not use spanner.Client() exposed methods for admin operations and eventually the methods will be only removed we will never remove `spanner_client.instance_admin_api`.

samples/generated/conftest.py
```diff
@@ -0,0 +1,293 @@
+# Copyright 2024 Google LLC All rights reserved.
```
parthea, line 1: Should this be moved to https://github.com/googleapis/python-spanner/tree/main/samples instead of https://github.com/googleapis/python-spanner/tree/main/samples/generated since the actual samples are handwritten?

samples/generated/noxfile.py
```diff
@@ -0,0 +1,292 @@
...
+# modify the test configurations. The values here should be in sync
+# with `noxfile_config.py`. Users will copy `noxfile_config.py` into
+# their directory and modify it.
+
+TEST_CONFIG = {
+    # You can opt out from the test for specific Python versions.
+    "ignored_versions": [],
+    # Old samples are opted out of enforcing Python type hints
+    # All new samples should feature them
+    "enforce_type_hints": False,
```
parthea, line 47: Please add a [noxfile_config.py](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/noxfile_config.py) file alongside `noxfile.py` and set `enforce_type_hints` to True for new samples

samples/generated/snippets.py
```diff
@@ -0,0 +1,104 @@
...
+
+    print("Created instance {}".format(instance_id))
+
+
+# [END spanner_create_instance]
+
+
+# [START spanner_create_database_with_default_leader]
+def create_database_with_default_leader(instance_id, database_id, default_leader):
+    """Creates a database with tables with a default leader."""
```
parthea, line 65: Do we want to include the function name in the snippet, where folks have to call the function or do we want to have it as a runnable snippet? 

For example, see the snippet below which can be run directly
https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/automl/snippets/batch_predict.py#L21

Instead of 
```
# [START spanner_create_database_with_default_leader]
def create_database_with_default_leader(instance_id, database_id, default_leader):
    """Creates a database with tables with a default leader."""
```
we could have
```
def create_database_with_default_leader(instance_id, database_id, default_leader):
    """Creates a database with tables with a default leader."""
    # [START spanner_create_database_with_default_leader]

    # TODO(developer): Uncomment and set the following variables
    # instance_id = "<replace with useful text for developer>"
    # database_id = "<replace with useful text for developer>"
    # default_leader =  "<replace with useful text for developer>"
    spanner_client = spanner.Client()
```

samples/generated/snippets.py
harshachinta: I understand that `spanner.Client` will not be removed.
But I am not clear on why do we still need to create admin client for customers? I mean why do we need `instance_admin_api` property in our client? Can we remove it and ask customers to create that using `spanner_admin_instance_v1.InstanceAdminClient() ` https://github.com/googleapis/python-spanner/blob/ec87c082570259d6e16834326859a73f6ee8286a/google/cloud/spanner_v1/client.py#L221

We do the same for Golang where customers create admin client from autogenerated code. Why are we doing this differently here by creating the admin client?
https://github.com/GoogleCloudPlatform/golang-samples/blob/f3a558cc0cf0212414e30235c7f69ee92357e3d5/spanner/spanner_snippets/spanner/spanner_create_instance.go#L31C33-L31C55

samples/generated/snippets.py
```diff
@@ -0,0 +1,104 @@
...
+    spanner_client = spanner.Client()
+    operation = spanner_client.database_admin_api.create_database(
+        request={
+            "parent": "projects/{}/instances/{}".format(
+                spanner_client.project, instance_id
+            ),
+            "create_statement": "CREATE DATABASE {}".format(database_id),
+            "extra_statements": [
+                """CREATE TABLE Singers (
```
harshachinta, line 74: nit: I understand string formatting does not work properly when applying lint.
Would you mind fixing this since this would be public snippet?

samples/generated/conftest.py
harshachinta: +1
My concern is we have duplicated files like conftest, noxfile etc. This would bring in burden of maintaining files at 2 places.
Can we move this to samples/samples/admin/generated? and share the same nox and conftest file that are there in samples/samples?

samples/generated/snippets.py
rahul2393: I don't think we need to be aligned with other languages here, benefit of going with spanner.Client is customers don't have to bother about yet another object, they can keep using the existing object to use any of the auto generated methods. 
@olavloite What are your thoughts on this?

harshachinta: I also looked at Java sample for autogenerated admin and that creates the admin client from autogenerated.
https://github.com/googleapis/java-spanner/blob/f689f742d8754134523ed0394b9c1b8256adcae2/samples/snippets/src/main/java/com/example/spanner/admin/generated/CreateInstanceExample.java#L39

rahul2393: Same I don't think we need to be aligned with Java/Go here.

I think we need to think about existing Python client libraries users and amount of change they need to make to migrate from handwritten methods to autogenerated methods.

olavloite: Hmm.... Looking at the Java example, I remember leaving a note somewhere that we need a strategy and example for how to create an admin client that uses the emulator. With this discussion on top of that, I'm starting to regret not following up more on that comment regarding the emulator.

My thinking here is that:
1. For 'normal' operations, creating an auto-generated client works fine, as you can just call `InstanceAdminClient.create()` (or something similar).
2. For operations that are not on the happy path, for example because you want to use the emulator, or you want to set some additional timeout and retry settings, then creating the auto-generated client directly is harder, as you need to specify those settings for the auto-generated client.

The above problem is especially an issue with the emulator, as we currently tell our customers 'just set `SPANNER_EMULATOR_HOST=localhost:9010`, and everything will use the emulator instead of production'. That is no longer true in this setup. Note that Go is a big exception here: It has a special feature that ensures that the `SPANNER_EMULATOR_HOST` environment variable is used for auto-generated clients. That is not true for (all? most of?) the other languages.

So based on this, I tend to agree with @rahul2393 here. We should follow the strategy of instructing our customers to create a normal hand-written client, and then getting the admin client from that. That will allow us to control what endpoint that admin client should use.
Alternatively, we need to look into the code generator for the various languages to make sure that they all support the `SPANNER_EMULATOR_HOST` environment variable (which would then need to be more generic than just for Spanner, meaning that each service must have some generic way of configuring 'my emulator env var is called XYZ').

Go is probably an exception here, as we already have support for using the emulator in auto-generated admin clients.

olavloite: Also left a comment on Node.js for this: https://github.com/googleapis/nodejs-spanner/pull/1987/files#r1464456457

I think you should follow up with the other team members involved here to pan out a strategy for what to do with the emulator.

harshachinta: Thanks @olavloite for taking a look on this.
I agree with Rahul and your opinion since using autogenerated client will fail for emulator.

samples/generated/noxfile.py
rahul2393: Removed the duplic noxfile and using the existing one.

samples/generated/conftest.py
rahul2393: Moved to samples/samples/admin/generated

samples/samples/admin/samples.py
```diff
@@ -0,0 +1,104 @@
...
+    database = operation.result(OPERATION_TIMEOUT_SECONDS)
+
+    print(
+        "Database {} created with default leader {}".format(
+            database.name, database.default_leader
+        )
+    )
+
+
+# [END spanner_create_database_with_default_leader]
```
harshachinta, line 104: This python file should be run as `main` program since this will get referred in the [Getting started with Python](https://cloud.google.com/spanner/docs/getting-started/python) guide.
Can we add it similar to existing one in snippets.py?
https://github.com/googleapis/python-spanner/blob/f3b23b268766b6ff2704da18945a1b607a6c8909/samples/samples/snippets.py#L2799

harshachinta (approved):

samples/samples/admin/samples.py
```diff
@@ -0,0 +1,104 @@
...
+
+# [END spanner_create_instance]
+
+
+# [START spanner_create_database_with_default_leader]
+def create_database_with_default_leader(instance_id, database_id, default_leader):
+    """Creates a database with tables with a default leader."""
+    spanner_client = spanner.Client()
+    operation = spanner_client.database_admin_api.create_database(
+        request={
```
harshachinta, line 68: nit: The `request` here can be either `dict` or `spanner_database_admin.CreateDatabaseRequest`.
Can we use `spanner_database_admin.CreateDatabaseRequest` instead of dict, since it will help customers understand how to use autogenerated code for building requests.

samples/samples/admin/samples.py
rahul2393: This is planned when we switch the tags later to new samples.

samples/generated/snippets.py
rahul2393: Thanks @parthea, I want to have one change at a time for this work, and follow the syntax above in separate PR for all samples in repository.

PR #1067, opened 2024-01-02
feat: Implementation for partitioned query in dbapi

ankiaga:


google/cloud/spanner_dbapi/parsed_statement.py
```diff
@@ -35,6 +35,13 @@ class ClientSideStatementType(Enum):
     START_BATCH_DML = 6
     RUN_BATCH = 7
     ABORT_BATCH = 8
+    PARTITION_QUERY = 9
+    RUN_PARTITION = 10
+
+
+class ClientSideStatementParamKey(Enum):
```
olavloite, line 42: Do we need this enum? Or is it enough to just know that a client-side statement has N parameters (in this case N=1 in all cases)

google/cloud/spanner_dbapi/partition_helper.py
```diff
@@ -0,0 +1,32 @@
+from dataclasses import dataclass
```
olavloite, line 1: nit: add copyright header

google/cloud/spanner_dbapi/partition_helper.py
```diff
@@ -0,0 +1,32 @@
...
+import base64
+
+
+def decode_from_string(encoded_partition_id):
+    gzip_bytes = base64.b64decode(bytes(encoded_partition_id, "utf-8"))
+    partition_id_bytes = gzip.decompress(gzip_bytes)
+    return pickle.loads(partition_id_bytes)
+
+
+def encode_to_string(batch_transaction_id, batch_result):
```
olavloite, line 15: What is `batch_result` in this case? The partition ID? If so, I think it would be better to name it like that. This sounds more like a query result.

olavloite (commented):
This looks generally good to me. I think we need some more tests that cover the un-happy path (e.g. what happens if you try to use this in a read/write transaction?)

google/cloud/spanner_dbapi/partition_helper.py
ankiaga: Done

google/cloud/spanner_dbapi/partition_helper.py
ankiaga: Renamed to partition_result as it contains the partition details coming from Spanner

google/cloud/spanner_dbapi/parsed_statement.py
ankiaga: We need this enum as its a key in the map which needs to accessed at different places

ankiaga: Removed the enum as discussed

google/cloud/spanner_dbapi/client_side_statement_executor.py
```diff
@@ -63,24 +64,26 @@ def execute(cursor: "Cursor", parsed_statement: ParsedStatement):
...
     if statement_type == ClientSideStatementType.SHOW_COMMIT_TIMESTAMP:
-        if connection._transaction is None:
-            committed_timestamp = None
-        else:
-            committed_timestamp = connection._transaction.committed
+        if (
+            connection._transaction is not None
+            and connection._transaction.committed is not None
+        ):
+            column_values.append(connection._transaction.committed)
```
olavloite, line 71: This seems to change the behavior of `show commit_timestamp`:
1. Previously (before this change), if the transaction had not committed, then `show commit_timestamp` would return one row with value `None`.
2. Now `show commit_timestamp` will return a result set with zero row.

Both return types would be OK, but we should not change the behavior unintentionally, and also not once one of the two behaviors has been included in a public release. Are you sure that 'no rows' is the return value that you would prefer in those cases? (I assume that we have not done a release of the dbapi driver since support for `show commit_timestamp` was added. If we have, then we need to undo this change in this PR.)

google/cloud/spanner_dbapi/client_side_statement_executor.py
```diff
@@ -63,24 +64,26 @@ def execute(cursor: "Cursor", parsed_statement: ParsedStatement):
...
     if statement_type == ClientSideStatementType.SHOW_READ_TIMESTAMP:
-        if connection._snapshot is None:
-            read_timestamp = None
-        else:
-            read_timestamp = connection._snapshot._transaction_read_timestamp
+        if (
+            connection._snapshot is not None
+            and connection._snapshot._transaction_read_timestamp is not None
+        ):
+            column_values.append(connection._snapshot._transaction_read_timestamp)
```
olavloite, line 82: Same comment as for `commit_timestamp`: this changes the behavior when there is no read timestamp from returning a single row with `None` to an empty result set.

tests/system/test_dbapi.py
```diff
@@ -536,6 +536,90 @@ def test_batch_dml_invalid_statements(self):
...
+        self._conn.commit()
+
+    def test_partitioned_query_in_rw_transaction(self):
+        """Test partition query throws exception when connection is not in
+        read-only mode and neither in auto-commit mode."""
+        self._cursor.execute("start batch dml")
+        for i in range(1, 11):
+            self._insert_row(i)
+        self._cursor.execute("run batch")
+        self._conn.commit()
```
olavloite, line 566: nit: You can skip this part in this test, as it is not needed

tests/system/test_dbapi.py
```diff
@@ -536,6 +536,90 @@ def test_batch_dml_invalid_statements(self):
...
+        with pytest.raises(ProgrammingError):
+            self._cursor.execute("PARTITION SELECT * FROM contacts")
+
+    def test_partitioned_query_with_dml_query(self):
+        """Test partition query throws exception when sql query is a DML query."""
+        self._cursor.execute("start batch dml")
+        for i in range(1, 11):
+            self._insert_row(i)
+        self._cursor.execute("run batch")
+        self._conn.commit()
```
olavloite, line 577: nit: same as above, this can be skipped

tests/system/test_dbapi.py
```diff
@@ -536,6 +536,90 @@ def test_batch_dml_invalid_statements(self):
...
+        self._cursor.execute("PARTITION SELECT * FROM contacts")
+        partition_id_rows = self._cursor.fetchall()
+        assert len(partition_id_rows) > 0
+
+        rows = []
+        for partition_id_row in partition_id_rows:
+            self._cursor.execute("RUN PARTITION " + partition_id_row[0])
+            rows = rows + self._cursor.fetchall()
+        assert len(rows) == 10
+        self._conn.commit()
```
olavloite, line 607: I would have expected this to fail, as `autocommit == True`

tests/system/test_dbapi.py
```diff
@@ -536,6 +536,90 @@ def test_batch_dml_invalid_statements(self):
...
+
+    def test_partitioned_query_with_client_transaction_started(self):
+        """Test partition query throws exception when connection is in
+        auto-commit mode but transaction started using client side statement."""
+        self._cursor.execute("start batch dml")
+        for i in range(1, 11):
+            self._insert_row(i)
+        self._cursor.execute("run batch")
+        self._conn.commit()
```
olavloite, line 616: nit: skip

olavloite (approved):

google/cloud/spanner_dbapi/client_side_statement_executor.py
ankiaga: So the release has not happened in last couple of months.

Also I think there is no change in the resultset from before as we would have returned a result set with zero row previously as well as we were checking that `column_value is not None` on line 100

google/cloud/spanner_dbapi/client_side_statement_executor.py
ankiaga: Please check the reply on previous comment

google/cloud/spanner_dbapi/client_side_statement_executor.py
```diff
@@ -89,14 +92,28 @@ def execute(cursor: "Cursor", parsed_statement: ParsedStatement):
...
     if statement_type == ClientSideStatementType.ABORT_BATCH:
         return connection.abort_batch()
+    if statement_type == ClientSideStatementType.PARTITION_QUERY:
+        partition_ids = connection.partition_query(parsed_statement)
+        return _get_streamed_result_set(
+            "PARTITION",
+            TypeCode.STRING,
+            partition_ids,
+        )
+    if statement_type == ClientSideStatementType.RUN_PARTITION:
```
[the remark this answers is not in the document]
ankiaga: IIUC what you are asking is already added in this PR like application developer can execute the client statement like `RUN PARTITION sddssqdwd` where `sddssqdwd` is the partition token. System test has also been added for it. Please see `test_partitioned_query` and other tests added in this PR which verifies this functionality

tests/system/test_dbapi.py
ankiaga: Done

tests/system/test_dbapi.py
ankiaga: Done

tests/system/test_dbapi.py
ankiaga: Done

tests/system/test_dbapi.py
ankiaga: The code just gives a warning if you try to commit in auto commit mode and returns.

Removed the commit call

google/cloud/spanner_dbapi/client_side_statement_executor.py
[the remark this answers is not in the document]
ankiaga: As discussed offline would take it as a separate PR

PR #1073, opened 2024-01-12
chore: support named schemas

olavloite:
Adds support for named schemas to the Python client library and the DB-API driver.


parthea:
The issue with the docs should be fixed via owlbot once https://github.com/googleapis/synthtool/pull/1916 is merged.

olavloite:
> The issue with the docs should be fixed via owlbot once [googleapis/synthtool#1916](https://github.com/googleapis/synthtool/pull/1916) is merged.

Ah, thanks for letting me know! 👍

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -510,13 +510,17 @@ def __iter__(self):
...
+    def list_tables(self, schema_name=""):
         """List the tables of the linked Database.
 
         :rtype: list
         :returns: The list of tables within the Database.
         """
-        return self.run_sql_in_snapshot(_helpers.SQL_LIST_TABLES)
+        return self.run_sql_in_snapshot(
+            sql=_helpers.SQL_LIST_TABLES,
+            params={"schema_name": schema_name},
```
ankiaga, line 521: We are naming param key as `schema_name` here but `@table_schema` in _helpers.py. Isn't that incorrect?

google/cloud/spanner_dbapi/_helpers.py
```diff
@@ -18,13 +18,13 @@
...
 FROM information_schema.tables
-WHERE table_catalog = '' AND table_schema = ''
+WHERE table_catalog = '' AND table_schema = @table_schema
 """
 
 SQL_GET_TABLE_COLUMN_SCHEMA = """
 SELECT COLUMN_NAME, IS_NULLABLE, SPANNER_TYPE
 FROM INFORMATION_SCHEMA.COLUMNS
-WHERE TABLE_SCHEMA = '' AND TABLE_NAME = @table_name
+WHERE TABLE_SCHEMA = @table_schema AND TABLE_NAME = @table_name
```
ankiaga, line 27: Shouldn't we have `table_catalog` check here also like in SQL_LIST_TABLES?

google/cloud/spanner_v1/database.py
```diff
@@ -951,20 +966,40 @@ def table(self, table_id):
...
+                )
             else:
-                where_clause = "WHERE SPANNER_STATE = 'COMMITTED'"
-            results = snapshot.execute_sql(_LIST_TABLES_QUERY.format(where_clause))
+                if self._database_dialect == DatabaseDialect.POSTGRESQL:
+                    where_clause = "WHERE TABLE_SCHEMA = $1"
+                    param_name = "p1"
+                else:
+                    where_clause = (
+                        "WHERE TABLE_SCHEMA = @schema AND SPANNER_STATE = 'COMMITTED'"
```
ankiaga, line 995: We are not adding `SPANNER_STATE = 'COMMITTED'` check in dbapi? Any reason why?

ankiaga, line 995: Dont we need table_catalog check also?

google/cloud/spanner_dbapi/cursor.py
olavloite: Oh, good catch. Yes, that is incorrect, and should also be `table_schema` here.

google/cloud/spanner_v1/database.py
olavloite: In this case, I was just keeping what was already there. My assumption is that the original author used the `SPANNER_STATE = 'COMMITTED'` clause to differentiate between system tables (which are views), and actual user tables. So it could probably be removed here as well, but I didn't want to do that as part of this PR, as it is an unrelated change.

olavloite: We could add that as well, but there's no real need at the moment. We're not adding support for multiple catalogs at the moment, and are also unlikely to do so. It would also complicate things further, because the default catalog for GoogleSQL is an empty string (like for the schema), while for PostgreSQL it is the database name.

google/cloud/spanner_v1/table.py
```diff
@@ -43,13 +43,26 @@ class Table(object):
     :param database: The database that owns the table.
     """
 
-    def __init__(self, table_id, database):
+    def __init__(self, table_id, database, schema_name=None):
```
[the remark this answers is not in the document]
olavloite: Yes:
1. In `list_tables` the current behavior is to list all tables in the default schema (so for GoogleSQL, in the schema named 'empty string'). We want to retain that behavior, so we don't make a breaking change for anyone.
2. `list_tables` should also be able to return all tables across all schemas. To do this, you can pass in `None`. That means that we cannot use `None` to specify the default behavior that we want.
3. The name of the default schema is different in PostgreSQL and GoogleSQL. So for `list_tables` we need an additional special value to indicate 'use the default schema of this database', which cannot be `None` (because point 2), and can also not be for example the empty string (because for PostgreSQL it would be something else).
4. Here, we don't need two different special values, so we can use `None` as the default, meaning that 'the user did not specify any schema'. That's then also what we should use in queries that use this table, and then Cloud Spanner will decide what it means when it receives an unqualified table name (a table name without a schema name).

google/cloud/spanner_dbapi/_helpers.py
ankiaga: As discussed offline not needed

google/cloud/spanner_v1/table.py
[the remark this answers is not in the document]
ankiaga: Got it

google/cloud/spanner_dbapi/cursor.py
ankiaga: Fixed

ankiaga (approved):

PR #1084, opened 2024-01-29
feat(spanner): add support for Proto Columns

harshachinta:
This PR adds support for proto columns feature.

Pending,
1. The integrations tests added for proto columns in this PR does not run on Emulator. The tests will run on emulator once emulator support for proto columns is complete.
2. Improve unit test coverage.
3. Add `column_info` comments in samples.
4. Check if there are any additional integration tests needed.


noxfile.py
```diff
@@ -313,7 +313,7 @@ def cover(session):
     test runs (not system test runs), and then erases coverage data.
     """
     session.install("coverage", "pytest-cov")
-    session.run("coverage", "report", "--show-missing", "--fail-under=99")
+    session.run("coverage", "report", "--show-missing", "--fail-under=98")
```
surbhigarg92, line 316: Why the coverage is reducing ?

tests/_fixtures.py
```diff
@@ -83,6 +91,8 @@
     ) PRIMARY KEY (CartId);
 """
 
+# TODO: Add Proto Bundle DDL statement in EMULATOR_DDL once b/338557401
```
surbhigarg92, line 94: Remove this TODO

surbhigarg92 (approved):

noxfile.py
harshachinta: There have been lot of changes in the spanner DB API that has reduced the unit test coverage drastically. With proto columns code the coverage fell below 99. I have created a seperate issue to improve the proto columns coverage. Doing it in the same PR will make the PR big and hard to review.
https://github.com/googleapis/python-spanner/issues/1139

tests/_fixtures.py
harshachinta: I would prefer to have this TODO as this is in the test suite and not in the client code. This will help any developer understand the changes needed when emulator tests are un-skipped thus saving time.

harshachinta: Removed TODO

PR #1085, opened 2024-01-29
fix: Few fixes in DBAPI

ankiaga:
The fixes done in this PR are

1) When database has not been initialized inside DBapi Connection then it should not throw exception when trying to release a session incase no session has been checked out

2) Adding different None checks while setting `description` property on `Cursor` to make sure its not set if metadata from the resultset doesn't have any row_type.fields present


olavloite:
nit: could you modify both the title of the PR and add a description of the actual fixes that are in this PR. The title is automatically included in the release notes, and users should preferably be able to pick up an understanding from that what bug has been fixed.

olavloite (approved):

PR #1092, opened 2024-02-02
feat!: Implementation of batch ddl in dbapi

ankiaga:
We are also adding a boolean property `buffer_ddl_statements` (disabled by default) which is changing the behavior of dbapi as currently it behave as per the property being enabled. We need to override the flag to enable it in SqlAlchemy and Django.  More details about this flag below:

If the connection is in auto commit mode then this flag doesn't have any significance as ddl statements would be executed as they come.
For connection not in auto commit mode:
-         If enabled ddl statements would be buffered at client and not executed at cloud spanner. When a non ddl statement comes or a transaction is committed then all the existing buffered ddl statements would be executed.
-         If disabled then its an error to execute ddl statement in autocommit mode.

More details here https://docs.google.com/document/d/1mSVC_AhPpdSA0xUoj4LUt2o2tVM_LDv6IdssiwWnbmw

### DIFF
--- a/google/cloud/spanner_dbapi/batch_executor.py
+++ b/google/cloud/spanner_dbapi/batch_executor.py
@@ -50,7 +50,7 @@ class BatchDdlExecutor:
         """Executes the statement when ddl batch is active by buffering the
         statement in-memory.
 
-        This method is internal and not for public use as it can change anytime.
+        This method is internal and not for public use
 
         :type parsed_statement: ParsedStatement
         :param parsed_statement: parsed statement containing sql query
@@ -67,14 +67,8 @@ class BatchDdlExecutor:
         """Executes all the buffered statements on the active ddl batch by
         making a call to Spanner.
 
-        This method is internal and not for public use as it can change anytime.
+        This method is internal and not for public use
         """
-        from google.cloud.spanner_dbapi import ProgrammingError
-
-        if self._connection._client_transaction_started:
-            raise ProgrammingError(
-                "Cannot execute DDL statement when transaction is already active."
-            )
         return self._connection.database.update_ddl(self._statements).result()
 
 
@@ -96,7 +90,7 @@ class BatchDmlExecutor:
         """Executes the statement when dml batch is active by buffering the
         statement in-memory.
 
-        This method is internal and not for public use as it can change anytime.
+        This method is internal and not for public use
 
         :type parsed_statement: ParsedStatement
         :param parsed_statement: parsed statement containing sql query and query
@@ -116,7 +110,7 @@ class BatchDmlExecutor:
         """Executes all the buffered statements on the active dml batch by
         making a call to Spanner.
 
-        This method is internal and not for public use as it can change anytime.
+        This method is internal and not for public use
         """
         return run_batch_dml(self._cursor, self._statements)
 
@@ -124,7 +118,7 @@ class BatchDmlExecutor:
 def run_batch_dml(cursor: "Cursor", statements: List[Statement]):
     """Executes all the dml statements by making a batch call to Spanner.
 
-    This method is internal and not for public use as it can change anytime.
+    This method is internal and not for public use
 
     :type cursor: Cursor
     :param cursor: Database Cursor object
--- a/google/cloud/spanner_dbapi/batch_executor.py
+++ b/google/cloud/spanner_dbapi/batch_executor.py
@@ -16,6 +16,8 @@ from __future__ import annotations
 
 from enum import Enum
 from typing import TYPE_CHECKING, List
+
+from google.cloud.spanner_dbapi import parse_utils
 from google.cloud.spanner_dbapi.parsed_statement import (
     ParsedStatement,
     StatementType,
@@ -28,6 +30,48 @@ from google.cloud.spanner_dbapi.utils import StreamedManyResultSets
 
 if TYPE_CHECKING:
     from google.cloud.spanner_dbapi.cursor import Cursor
+    from google.cloud.spanner_dbapi.connection import Connection
+
+
+class BatchDdlExecutor:
+    """Executor that is used when a DDL batch is started. These batches only
+    accept DDL statements. All DDL statements are buffered locally and sent to
+    Spanner when runBatch() is called.
+
+    :type "Connection": :class:`~google.cloud.spanner_dbapi.connection.Connection`
+    :param connection:
+    """
+
+    def __init__(self, connection: "Connection"):
+        self._connection = connection
+        self._statements: List[str] = []
+
+    def execute_statement(self, parsed_statement: ParsedStatement):
+        """Executes the statement when ddl batch is active by buffering the
+        statement in-memory.
+
+        :type parsed_statement: ParsedStatement
+        :param parsed_statement: parsed statement containing sql query
+        """
+        from google.cloud.spanner_dbapi import ProgrammingError
+
+        if parsed_statement.statement_type != StatementType.DDL:
+            raise ProgrammingError("Only DDL statements are allowed in batch DDL mode.")
+        self._statements.extend(
+            parse_utils.parse_and_get_ddl_statements(parsed_statement.statement.sql)
+        )
+
+    def run_batch(self):
+        """Executes all the buffered statements on the active ddl batch by
+        making a call to Spanner.
+        """
+        from google.cloud.spanner_dbapi import ProgrammingError
+
+        if self._connection._client_transaction_started:
+            raise ProgrammingError(
+                "Cannot execute DDL statement when transaction is already active."
+            )
+        return self._connection.database.update_ddl(self._statements).result()
 
 
 class BatchDmlExecutor:
@@ -52,6 +96,7 @@ class BatchDmlExecutor:
         :param parsed_statement: parsed statement containing sql query and query
          params
         """
+
         from google.cloud.spanner_dbapi import ProgrammingError
 
         if (
@@ -61,7 +106,7 @@ class BatchDmlExecutor:
             raise ProgrammingError("Only DML statements are allowed in batch DML mode.")
         self._statements.append(parsed_statement.statement)
 
-    def run_batch_dml(self):
+    def run_batch(self):
         """Executes all the buffered statements on the active dml batch by
         making a call to Spanner.
         """
--- a/google/cloud/spanner_dbapi/batch_executor.py
+++ b/google/cloud/spanner_dbapi/batch_executor.py
@@ -50,6 +50,8 @@ class BatchDdlExecutor:
         """Executes the statement when ddl batch is active by buffering the
         statement in-memory.
 
+        This method is internal and not for public use as it can change anytime.
+
         :type parsed_statement: ParsedStatement
         :param parsed_statement: parsed statement containing sql query
         """
@@ -64,6 +66,8 @@ class BatchDdlExecutor:
     def run_batch(self):
         """Executes all the buffered statements on the active ddl batch by
         making a call to Spanner.
+
+        This method is internal and not for public use as it can change anytime.
         """
         from google.cloud.spanner_dbapi import ProgrammingError
 
@@ -92,6 +96,8 @@ class BatchDmlExecutor:
         """Executes the statement when dml batch is active by buffering the
         statement in-memory.
 
+        This method is internal and not for public use as it can change anytime.
+
         :type parsed_statement: ParsedStatement
         :param parsed_statement: parsed statement containing sql query and query
          params
@@ -109,12 +115,16 @@ class BatchDmlExecutor:
     def run_batch(self):
         """Executes all the buffered statements on the active dml batch by
         making a call to Spanner.
+
+        This method is internal and not for public use as it can change anytime.
         """
         return run_batch_dml(self._cursor, self._statements)
 
 
 def run_batch_dml(cursor: "Cursor", statements: List[Statement]):
     """Executes all the dml statements by making a batch call to Spanner.
+
+    This method is internal and not for public use as it can change anytime.
 
     :type cursor: Cursor
     :param cursor: Database Cursor object
--- a/google/cloud/spanner_dbapi/connection.py
+++ b/google/cloud/spanner_dbapi/connection.py
@@ -137,15 +137,20 @@ class Connection:
     @property
     def buffer_ddl_statements(self):
         """Whether to buffer ddl statements for this connection.
-        If the connection is in auto commit mode then this flag doesn't have any
-        significance as ddl statements would be executed as they come.
+        This flag determines how DDL statements are handled when auto_commit=False:
 
-        For connection not in auto commit mode:
-        If enabled ddl statements would be buffered at client and not executed
-        at cloud spanner. When a non ddl statement comes or a transaction is
-        committed then all the existing buffered ddl statements would be executed.
+        1. buffer_ddl_statements=True: DDL statements are buffered in the client until the
+            next non-DDL statement, or until the transaction is committed. Executing a
+            non-DDL statement causes the connection to send all buffered DDL statements
+            to Spanner, and then to execute the non-DDL statement. Note that although the
+            DDL statements are sent as one batch to Spanner, they are not guaranteed to be
+            atomic. See https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.admin.database.v1#google.spanner.admin.database.v1.UpdateDatabaseDdlRequest
+            for more details on DDL batches.
+        2. buffer_ddl_statements=False: Executing DDL statements is not allowed and an
+            error will be raised when the connection is in auto_commit=False mode, a
+            transaction is active, and DDL statements cannot be executed in a transaction.
 
-        If disabled then its an error to execute ddl statement in autocommit mode.
+        This flag is ignored when auto_commit=True.
 
 
         :rtype: bool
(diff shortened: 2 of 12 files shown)


google/cloud/spanner_dbapi/connection.py
```diff
@@ -126,6 +134,15 @@ def autocommit(self):
         """
         return self._autocommit
 
+    @property
+    def buffer_ddl_statements(self):
+        """Whether to buffer ddl statements for this connection.
```
olavloite, line 139: This needs more documentation. I (think I) understand what it means. To someone who is not familiar with Cloud Spanner, it is not clear what this means. It is also not clear when they should enable/disable this.

google/cloud/spanner_dbapi/connection.py
```diff
@@ -91,7 +95,9 @@ class Connection:
         should end a that a new one should be started when the next statement is executed.
     """
 
-    def __init__(self, instance, database=None, read_only=False):
+    def __init__(
+        self, instance, database=None, read_only=False, buffer_ddl_statements=False
```
olavloite, line 99: `buffer_ddl_statements=False` means that we are changing the default behavior in a significant way. That requires a major version bump, and should be clearly called out in the release notes. The description of the PR will automatically be included in the release notes, so please add information about this change there.

Also, in order to force a major version bump, you can add a `!` to the title like this: `feat!: Support explicit DDL batching`

(Also note that the title of the PR will be included in the release notes, and as such is intended for public consumption. We should therefore try to keep that title as descriptive as possible for external users so they understand what the change is.)

google/cloud/spanner_dbapi/connection.py
```diff
@@ -463,6 +481,31 @@ def validate(self):
...
+            )
+        if self.buffer_ddl_statements:
+            raise ProgrammingError(
+                "Cannot start a DDL batch when _buffer_ddl_statements flag is True"
+            )
+        self._batch_mode = BatchMode.DDL
+        self._batch_ddl_executor = BatchDdlExecutor(self)
+
+    @check_not_closed
+    def execute_batch_ddl_statement(self, parsed_statement: ParsedStatement):
```
olavloite, line 502: Do we want this method to be public?

google/cloud/spanner_dbapi/connection.py
```diff
@@ -486,22 +529,28 @@ def execute_batch_dml_statement(self, parsed_statement: ParsedStatement):
 
     @check_not_closed
     def run_batch(self):
+        result_set = None
```
olavloite, line 532: I think that we should consider returning something in case of a DDL batch, for example the number of DDL statements that was executed. In Java, we return an array that has the same length as the number of statements, and that contains a 1 for each statement that we executed. This specific return type has to do with that JDBC expects a driver to return, and in this case I think we have some more freedom. What do other database systems return in case like this? For example if you in PostgreSQL execute something like the following SQL string: `create table table1 (...); create table table2 (...); create index index1 on table1 (...);`

google/cloud/spanner_dbapi/connection.py
```diff
@@ -584,10 +633,14 @@ def connect(
     pool=None,
     user_agent=None,
     client=None,
+    buffer_ddl_statements=False,
     route_to_leader_enabled=True,
 ):
     """Creates a connection to a Google Cloud Spanner database.
 
+    :type buffer_ddl_statements: bool
+    :param buffer_ddl_statements:
```
olavloite, line 642: This needs a description of what it does, and when you should enable/disable it.

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -269,9 +259,18 @@ def _execute(self, sql, args=None, call_from_execute_many=False):
...
-                self._batch_DDLs(sql)
-                if not self.connection._client_transaction_started:
-                    self.connection.run_prior_DDL_statements()
+                if not self.connection.buffer_ddl_statements:
+                    if not self.connection._client_transaction_started:
+                        self._batch_DDLs(sql)
+                        self.connection.run_prior_DDL_statements()
+                    else:
+                        raise ProgrammingError(
+                            "Cannot execute DDL statement when transaction is already active"
```
olavloite, line 268:
```suggestion
                            "Cannot execute DDL statement when a transaction is already active"
```

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -269,9 +259,18 @@ def _execute(self, sql, args=None, call_from_execute_many=False):
...
             elif self._parsed_statement.statement_type == StatementType.DDL:
-                self._batch_DDLs(sql)
-                if not self.connection._client_transaction_started:
-                    self.connection.run_prior_DDL_statements()
+                if not self.connection.buffer_ddl_statements:
+                    if not self.connection._client_transaction_started:
+                        self._batch_DDLs(sql)
+                        self.connection.run_prior_DDL_statements()
+                    else:
+                        raise ProgrammingError(
```
olavloite, line 267: Isn't this also a change in behavior? It's a bit hard to read with all the changes here, but I think that previously we would silently accept this, and now we are returning an error?

google/cloud/spanner_dbapi/connection.py
```diff
@@ -126,6 +134,15 @@ def autocommit(self):
         """
         return self._autocommit
 
+    @property
+    def buffer_ddl_statements(self):
```
olavloite, line 138: Is this a read-only property? As in; can it only be set when a connection is created, and never changed after that? If so, is that something that we can/should document? Or is it clear from the context?

tests/system/test_dbapi.py
```diff
@@ -152,7 +154,7 @@ def test_commit_exception(self):
         try:
             self._conn.commit()
         except Exception:
-            pass
+            self._conn.database._pool._sessions.get()
```
olavloite, line 157: If I remember correctly, this change is needed, because in this case no session has been checked out of the pool. Is this something that could also change for user code? In other words: Could it be that users have code in their applications that trigger this same thing, and that previously worked without an error, and that would now throw an error?

tests/system/test_dbapi.py
```diff
@@ -454,6 +456,49 @@ def test_read_timestamp_client_side_autocommit(self):
         read_timestamp_query_result_2 = self._cursor.fetchall()
         assert read_timestamp_query_result_1 != read_timestamp_query_result_2
 
+    @pytest.mark.parametrize("auto_commit", [False, True])
+    def test_batch_ddl(self, auto_commit, dbapi_database):
+        """Test batch ddl."""
+
+        self._conn._buffer_ddl_statements = False
+        if auto_commit:
```
olavloite, line 464: Can this be simplified to `self._conn.autocommit = auto_commit`

tests/system/test_dbapi.py
```diff
@@ -454,6 +456,49 @@ def test_read_timestamp_client_side_autocommit(self):
...
+            self._conn.autocommit = True
+
+        self._cursor.execute("start batch ddl")
```
olavloite, line 467: I think it would make more sense if this statement already causes an error when `auto_commit=False`, instead of letting the error only happen when `run batch` is executed. Or is there some scenario where this could become a valid DDL batch?

tests/system/test_dbapi.py
```diff
@@ -454,6 +456,49 @@ def test_read_timestamp_client_side_autocommit(self):
...
+        except ProgrammingError:
+            exception_raised = True
+        table_1 = dbapi_database.table("Table_1")
+        table_2 = dbapi_database.table("Table_2")
+        if auto_commit:
+            assert exception_raised is False
+            assert table_1.exists() is True
+            self._cursor.execute("DROP TABLE Table_1")
+            assert table_2.exists() is True
+            self._cursor.execute("DROP TABLE Table_2")
```
olavloite, line 496: nit: combine the two `DROP TABLE` statements into one batch, that is more efficient to execute on Spanner. (So first verify that both tables exist, and then drop them)

tests/system/test_dbapi.py
```diff
@@ -1113,44 +1158,13 @@ def test_execute_many(self):
...
-                op = dbapi_database.update_ddl(["DROP TABLE Singers"])
-                op.result()
-
-    def test_ddl_execute_autocommit_true(self, dbapi_database):
-        """Check that DDL statement in autocommit mode results in successful
-        DDL statement execution for execute method."""
+    @pytest.mark.parametrize("autocommit", [True, False])
+    def test_ddl_execute(self, autocommit, dbapi_database):
+        """Check that DDL statement results in successful execution for execute
+        method in autocommit mode while it's a noop in non-autocommit mode."""
```
olavloite, line 1164: This comment should include the fact that the connection also uses buffered DDL. And that again means that the DDL statement is not a no-op (I think), but it is being buffered. And the test should then verify tha the statement is actually buffered, and that it is not really a no-op.

tests/system/test_dbapi.py
```diff
@@ -1196,7 +1232,7 @@ def test_ddl_executemany_autocommit_false(self, dbapi_database):
         table = dbapi_database.table("DdlExecuteManyAutocommit")
         assert table.exists() is False
 
-    def test_ddl_execute(self, dbapi_database):
+    def test_ddl_then_non_ddl_execute(self, dbapi_database):
         """Check that DDL statement followed by non-DDL execute statement in
         non autocommit mode results in successful DDL statement execution."""
```
olavloite, line 1237: Also update the comments of these test cases, that they require `buffer_ddl=True`.

tests/system/test_dbapi.py
```diff
@@ -1339,31 +1375,23 @@ def test_json_array(self, dbapi_database):
                 op = dbapi_database.update_ddl(["DROP TABLE JsonDetails"])
                 op.result()
 
-    @pytest.mark.noautofixt
-    def test_DDL_commit(self, shared_instance, dbapi_database):
+    def test_DDL_commit(self, dbapi_database):
         """Check that DDLs in commit mode are executed on calling `commit()`."""
```
olavloite, line 1379: I think that this also is only true if `buffer_ddl=True`. In other cases, this would cause an error (?)

google/cloud/spanner_dbapi/batch_executor.py
```diff
@@ -61,7 +106,7 @@ def execute_statement(self, parsed_statement: ParsedStatement):
             raise ProgrammingError("Only DML statements are allowed in batch DML mode.")
         self._statements.append(parsed_statement.statement)
 
-    def run_batch_dml(self):
+    def run_batch(self):
```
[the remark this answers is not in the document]
ankiaga: Ack. Added the following documentation on the method `This method is internal and not for public use as it can change anytime.`

google/cloud/spanner_dbapi/connection.py
ankiaga: Added

google/cloud/spanner_dbapi/cursor.py
ankiaga: No this code path would only be reached when `buffer_ddl_statements` flag is disabled which was not the case earlier

google/cloud/spanner_dbapi/connection.py
ankiaga: Yes its a read only property and there are no setters for it but in python the private variable can also be overridden.

tests/system/test_dbapi.py
ankiaga: Not completely sure but seems like something has changed in the session logic at python client or Span FE as this test was working before but has started failing (irrespective of changes in this PR). 
So in this test we are deleting the session at SpanFE so that commit would fail and we catch the exception. We then try to test that a new transaction works on the same connection but seems like now the new transaction is trying to checkout the same session which was deleted resulting in `Session not found` exception. So to fix this I am removing the session from the session pool at client by getting the first session from the queue

tests/system/test_dbapi.py
ankiaga: Done

tests/system/test_dbapi.py
ankiaga: Done

tests/system/test_dbapi.py
ankiaga: Done

tests/system/test_dbapi.py
ankiaga: Thats correct. Added the comment

google/cloud/spanner_dbapi/connection.py
ankiaga: Added comment that its an internal method to this and other methods as well

google/cloud/spanner_dbapi/batch_executor.py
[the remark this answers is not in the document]
parthea: Instead of renaming the method, we could mark it as deprecated and create a new one with the `_` prefix which indicates that it is internal. Please avoid a major version bump if at all possible as it is disruptive for downstream users.

google/cloud/spanner_dbapi/batch_executor.py
```diff
@@ -61,16 +112,20 @@ def execute_statement(self, parsed_statement: ParsedStatement):
...
         """Executes all the buffered statements on the active dml batch by
         making a call to Spanner.
+
+        This method is internal and not for public use as it can change anytime.
```
olavloite, line 119: nit: here and elsewhere

```suggestion
        This method is internal and not for public use
```

google/cloud/spanner_dbapi/connection.py
```diff
@@ -126,6 +134,25 @@ def autocommit(self):
...
+        """Whether to buffer ddl statements for this connection.
+        If the connection is in auto commit mode then this flag doesn't have any
+        significance as ddl statements would be executed as they come.
+
+        For connection not in auto commit mode:
+        If enabled ddl statements would be buffered at client and not executed
+        at cloud spanner. When a non ddl statement comes or a transaction is
+        committed then all the existing buffered ddl statements would be executed.
+
+        If disabled then its an error to execute ddl statement in autocommit mode.
```
olavloite, line 148:
```suggestion
        This flag determines how DDL statements are handled when auto_commit=False:
        
        1. buffer_ddl_statements=True: DDL statements are buffered in the client until the
            next non-DDL statement, or until the transaction is committed. Executing a
            non-DDL statement causes the connection to send all buffered DDL statements
            to Spanner, and then to execute the non-DDL statement. Note that although the
            DDL statements are sent as one batch to Spanner, they are not guaranteed to be
            atomic. See https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.admin.database.v1#google.spanner.admin.database.v1.UpdateDatabaseDdlRequest
            for more details on DDL batches.
        2. buffer_ddl_statements=False: Executing DDL statements is not allowed and an
            error will be raised. When the connection is in auto_commit=False mode, a
            transaction is active, and DDL statements cannot be executed in a transaction.
            
        This flag is ignored when auto_commit=True.
```

google/cloud/spanner_dbapi/cursor.py
olavloite: ack

tests/system/test_dbapi.py
olavloite: OK, would you mind adding a comment to the code here to explain why we are doing this.

tests/system/test_dbapi.py
olavloite: I think that my main takeaway from that is: It should not be possible to change the value of `auto_commit` while a batch is in progress. That should be the case for both for DDL batches and DML batches.

Take for example DML batches, and assume the following script:

```
set auto_commit=True;
start batch dml;
insert into foo (..);
set auto_commit=False;
insert into bar (...);
run batch;
rollback;
```

What should be the result of the above script?

google/cloud/spanner_dbapi/batch_executor.py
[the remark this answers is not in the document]
ankiaga: This is not the reason because of which we are proposing a major version bump. Please check https://docs.google.com/document/d/1mSVC_AhPpdSA0xUoj4LUt2o2tVM_LDv6IdssiwWnbmw for details

tests/system/test_dbapi.py
ankiaga: Made the change to not allow changing autocommit mode when a batch is in progress. Also raising error when starting a ddl batch when connection is not in autocommit mode

google/cloud/spanner_dbapi/batch_executor.py
ankiaga: Done

google/cloud/spanner_dbapi/connection.py
ankiaga: Tested that postgres also returns None

olavloite: Ack. Let's keep it None here as well then 👍

google/cloud/spanner_dbapi/batch_executor.py
olavloite: Could it be that you have not pushed that change yet?

ankiaga: Yes I did not. Thanks done

google/cloud/spanner_dbapi/connection.py

PR #1101, opened 2024-02-13
chore: Adding schema name property in dbapi connection

ankiaga:


google/cloud/spanner_dbapi/connection.py
```diff
@@ -124,6 +125,24 @@ def spanner_client(self):
...
+        return self._schema_name
+
+    @schema_name.setter
+    def schema_name(self, value):
+        """Updates the database schema name used in the connection.
+
+        :type value: str
+        :param value: database schema name used in the connection
+        """
+        self._schema_name = value
```
olavloite, line 144: We should remove this. Setting a current schema is currently not (fully) supported, so we should also not 'pretend' that it is here.

The thing that is not supported, but that would be implicitly implied by this, is the following:

```sql
-- The following statement should automatically use "current_schema.my_table"
select *
from my_table
```

google/cloud/spanner_dbapi/connection.py
```diff
@@ -124,6 +125,24 @@ def spanner_client(self):
         """
         return self._instance._client
 
+    @property
+    def schema_name(self):
```
olavloite, line 129: This property should be named `current_schema`.

google/cloud/spanner_dbapi/connection.py
```diff
@@ -124,6 +125,24 @@ def spanner_client(self):
...
+        """schema name for this database.
```
olavloite, line 130: This property indicates the default schema that is used by this __connection__, not the database that it is connected to.

```suggestion
        """schema name for this connection.
```

google/cloud/spanner_dbapi/connection.py
```diff
@@ -124,6 +125,24 @@ def spanner_client(self):
...
+
+        :rtype: str
+        :returns: "" for GoogleSQL and "public" for PostgreSQL for default schema
```
olavloite, line 133:
```suggestion
        :returns: the current default schema of this connection. Currently, this is always "" for GoogleSQL and "public" for PostgreSQL databases.
```

google/cloud/spanner_dbapi/connection.py
```diff
@@ -116,6 +116,7 @@ def __init__(self, instance, database=None, read_only=False):
         self._batch_mode = BatchMode.NONE
         self._batch_dml_executor: BatchDmlExecutor = None
         self._transaction_helper = TransactionRetryHelper(self)
+        self._schema_name = database.default_schema_name
```
olavloite, line 119: (Overall comment, not specifically for this line): You should always add tests, even for something as simple as this. Not because it is unclear whether this implementation works as expected, but to prevent accidental regressions in the future. In this case, we should verify that the returned value for `current_schema` is "" for GoogleSQL and "public" for PostgreSQL. This prevents accidental regressions if someone modifies the property definition in the future without realizing that other features might depend on this specific behavior.

google/cloud/spanner_dbapi/connection.py

google/cloud/spanner_dbapi/connection.py

google/cloud/spanner_dbapi/connection.py
ankiaga: Thanks. Done

google/cloud/spanner_dbapi/connection.py
ankiaga: Ack. Will take care from next time.
Done

google/cloud/spanner_dbapi/connection.py
```diff
@@ -591,6 +604,7 @@ def connect(
     pool=None,
     user_agent=None,
     client=None,
+    database_dialect=DatabaseDialect.GOOGLE_STANDARD_SQL,
```
olavloite, line 607: I think that it would be better if we would automatically read this from the database, instead of requiring the user to supply it as an argument. We already read this property from the database in the `database.default_schema` method.

tests/system/test_dbapi.py
```diff
@@ -57,6 +58,7 @@ def raw_database(shared_instance, database_operation_timeout, not_postgres):
         database_id,
         ddl_statements=DDL_STATEMENTS,
         pool=pool,
+        database_dialect=DatabaseDialect.GOOGLE_STANDARD_SQL,
```
olavloite, line 61: Why are we setting this here? Shouldn't the database just read it itself from the server?

olavloite (approved):

tests/system/test_dbapi.py
ankiaga: Yes, removed. Thanks

PR #1103, opened 2024-02-15
feat: support partitioned dml in dbapi

ankiaga:


google/cloud/spanner_dbapi/connection.py
```diff
@@ -540,7 +550,8 @@ def partition_query(
         return partition_ids
 
     @check_not_closed
-    def run_partition(self, encoded_partition_id):
+    def run_partition(self, parsed_statement: ParsedStatement):
```
olavloite, line 553: This is a breaking change of a public method

google/cloud/spanner_dbapi/connection.py
```diff
@@ -565,6 +576,24 @@ def run_partitioned_query(
             partitioned_query, statement.params, statement.param_types
         )
 
+    @check_not_closed
+    def set_autocommit_dml_mode(
```
olavloite, line 580: I think that it would make more sense to have a public method that takes a `AutocommitDmlMode` enum value as an argument, rather than one that takes a parsed statement as an argument. This method looks like a private method that should only be called by our own parser.

Instead, we should:
1. Have a private method that is basically the same as this.
2. Have a public method that takes an `AutocommitDmlMode` enum as an input argument and that actually changes the value of the flag. That method would then also be usable for anyone using the dbapi driver programmatically, as they could just call that method to change the value.

See https://github.com/googleapis/java-spanner-jdbc/blob/910a130a02f72c4d8764f12e347f6c3d1bd51b2b/src/main/java/com/google/cloud/spanner/jdbc/CloudSpannerJdbcConnection.java#L136 for how the API in the JDBC driver looks.

google/cloud/spanner_dbapi/connection.py
ankiaga: Reverted

google/cloud/spanner_dbapi/connection.py

google/cloud/spanner_dbapi/connection.py
```diff
@@ -565,6 +575,31 @@ def run_partitioned_query(
...
+    @check_not_closed
+    def _set_autocommit_dml_mode(
+        self,
+        parsed_statement: ParsedStatement,
+    ):
+        autocommit_dml_mode_str = parsed_statement.client_side_statement_params[0]
+        autocommit_dml_mode = AutocommitDmlMode[autocommit_dml_mode_str.upper()]
+        self.set_autocommit_dml_mode(autocommit_dml_mode)
+
+    def set_autocommit_dml_mode(
```
olavloite, line 587: nit: add a comment to this (public) method that explains what it does.

google/cloud/spanner_dbapi/connection.py
```diff
@@ -155,6 +157,14 @@ def database(self):
         """
         return self._database
 
+    @property
+    def autocommit_dml_mode(self):
+        """AutocommitDmlMode of this connection.
```
olavloite, line 162: nit: Add information what each value means, and what you can do with this property.

olavloite (approved):

PR #1107, opened 2024-02-26
feat: add retry and timeout for batch dml

### LINKED ISSUE #927: Add sample for passing timeout and retry parameters

harshachinta:
Fixes b/326596176

This PR makes the following changes
1. allows setting retry and timeout to batch_update method. This PR https://github.com/googleapis/python-spanner/pull/251 added changes only to few methods.
2. Add a public sample for retry and timeout. https://cloud.google.com/spanner/docs/custom-timeout-and-retry

Fixes #927


samples/samples/snippets.py
```diff
@@ -2796,6 +2796,33 @@ def directed_read_options(
...
+    # [START spanner_set_custom_timeout_and_retry]
+    from google.api_core.retry import Retry
+
+    # instance_id = "your-spanner-instance"
+    # database_id = "your-spanner-db-id"
+    spanner_client = spanner.Client()
+    instance = spanner_client.instance(instance_id)
+    database = instance.database(database_id)
+
+    retry = Retry(initial=5, maximum=100, multiplier=2, timeout=60)
```
surbhigarg92, line 2810: Add comments for each retry option.

harshachinta: PTAL.

surbhigarg92: These values should be inline with the documentation . As per the below documentation , we should use `multiplier` as `1.5` , `initial` as `500ms` , `maxretrydelay` as `16seconds` which i guess is the `timeout` value here. 

Also we should add a retry for `DEADLINE_EXCEEDED` error.

`
In the following sample, a timeout of 60 seconds is set for the given operation. If the operation takes longer than this timeout, the operation fails with a DEADLINE_EXCEEDED error.

If the operation fails with an UNAVAILABLE error code, for example, if there is a transient network problem, the operation is retried. The client waits for 500 ms before starting the first retry attempt. If the first retry fails, the client waits for 1.5 * 500 ms = 750 ms before starting the second retry. This retry delay continues to increase until the operation either succeeds or reaches the maximum retry delay of 16 seconds. The operation fails with a DEADLINE_EXCEEDED error if the total time that is spent on trying the operation exceeds the total timeout value of 60 seconds.
`

harshachinta: @surbhigarg92 
Thanks for pointing this out. I have modified the sample to align with the documentation
https://cloud.google.com/spanner/docs/custom-timeout-and-retry

samples/samples/snippets.py
```diff
@@ -3017,6 +3017,49 @@ def directed_read_options(
...
+    spanner_client = spanner.Client()
+    instance = spanner_client.instance(instance_id)
+    database = instance.database(database_id)
+
+    # Customize retry with an initial wait time of 500 milliseconds.
+    # Customize retry with a maximum wait time of 16 seconds.
+    # Customize retry with a wait time multiplier per iteration of 1.5.
+    # Customize retry with a timeout on
+    # how long a certain RPC may be retried in case the server returns an error.
+    retry = retry.Retry(
```
surbhigarg92, line 3037: nit: Can we add all these comments above individual property. Something like below

```
 retry = retry.Retry(
      # Customize retry with an initial wait time of 500 milliseconds.
        initial=0.5,
     # Customize retry with a maximum wait time of 16 seconds.
        maximum=16,
        multiplier=1.5,
        timeout=60,
        predicate=retry.if_exception_type(
            core_exceptions.ServiceUnavailable,
        ),
    )
```

surbhigarg92 (approved):

samples/samples/snippets.py
harshachinta: Done

samples/samples/snippets.py
```diff
@@ -3017,6 +3017,51 @@ def directed_read_options(
...
+        maximum=16,
+        # Customize retry with a wait time multiplier per iteration of 1.5.
+        multiplier=1.5,
+        # Customize retry with a timeout on
+        # how long a certain RPC may be retried in
+        # case the server returns an error.
+        timeout=60,
+        # Configure which errors should be retried.
+        predicate=retry.if_exception_type(
+            core_exceptions.ServiceUnavailable,
```
harshachinta, line 3045: https://github.com/googleapis/python-api-core/blob/4fed37cbc32122f156e38250b5fa8b2b08a787a1/google/api_core/exceptions.py#L420

ServiceUnavailable is same as gRPC `UNAVAILABLE` error code.

PR #1109, opened 2024-02-29
docs: update all public documents to use auto-generated admin clients.

rahul2393:
* Moved existing samples to - `samples/samples/archived` . The archived samples will be referenced from release notes, public docs, etc. so that older users can navigate to the older docs.
* Moved new samples to - `samples/samples` (on merging the PR, this will ensure public documentation is updated since the name, file-path of sample is now same as what our documents reference).


arpan14 (approved):

harshachinta:
@rahul2393 
Quick check on Snippet bot comment.
image

There is a mismatch in region tags that got updated. I am not sure of the reason but can you check if any tags got missed to be updated or got deleted?

rahul2393:
@harshachinta It was expected 3 samples which got missed from earlier merge was incorporated with this.

PR #1110, opened 2024-03-04
docs: use autogenerated methods to get names from admin samples

rahul2393:
* Updated admin samples to remove usage of handwritten instance and database classes.

### DIFF
--- a/samples/samples/backup_sample.py
+++ b/samples/samples/backup_sample.py
@@ -126,18 +126,17 @@ def restore_database(instance_id, new_database_id, backup_id):
     from google.cloud.spanner_admin_database_v1 import RestoreDatabaseRequest
 
     spanner_client = spanner.Client()
-    instance_admin_api = spanner_client.instance_admin_api
+    database_admin_api = spanner_client.database_admin_api
 
     # Start restoring an existing backup to a new database.
     request = RestoreDatabaseRequest(
-        parent=instance_admin_api.instance_path(spanner_client.project, instance_id),
+        parent=database_admin_api.instance_path(spanner_client.project, instance_id),
         database_id=new_database_id,
-        backup="{}/backups/{}".format(
-            instance_admin_api.instance_path(spanner_client.project, instance_id),
-            backup_id,
+        backup=database_admin_api.backup_path(
+            spanner_client.project, instance_id, backup_id
         ),
     )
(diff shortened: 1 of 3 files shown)


samples/samples/backup_sample.py
```diff
@@ -35,23 +35,25 @@ def create_backup(instance_id, database_id, backup_id, version_time):
         backup as backup_pb
 
     spanner_client = spanner.Client()
-    instance = spanner_client.instance(instance_id)
-    database = instance.database(database_id)
+    instance_admin_api = spanner_client.instance_admin_api
```
arpan14, line 38: Do we really need both `instance_admin_api` and `database_admin_api`?

Its a bit weird that customers need to create both these clients even though all they need is an utility method - `instance_path`.

samples/samples/backup_sample.py
```diff
@@ -465,11 +502,14 @@ def update_backup(instance_id, backup_id):
         backup as backup_pb
 
     spanner_client = spanner.Client()
-    instance = spanner_client.instance(instance_id)
+    instance_admin_api = spanner_client.instance_admin_api
 
     backup = spanner_client.database_admin_api.get_backup(
         backup_pb.GetBackupRequest(
-            name="{}/backups/{}".format(instance.name, backup_id)
+            name="{}/backups/{}".format(
```
arpan14, line 509: use backup_path if available.

samples/samples/pg_snippets.py
```diff
@@ -1681,6 +1690,9 @@ def insert_customers(transaction):
             )
         )
 
+    instance = spanner_client.instance(instance_id)
```
arpan14, line 1693: why do we need this handwritten objects?

samples/samples/snippets.py
```diff
@@ -2862,6 +2903,9 @@ def insert_customers(transaction):
             )
         )
 
+    instance = spanner_client.instance(instance_id)
```
arpan14, line 2906: where are these handwritten objects used? If not needed, can we perform a search in the entire folder and remove the references of handwritten objects?

rahul2393: We need this to do data client inserts.

samples/samples/pg_snippets.py
rahul2393: We need this to do data client inserts.

PR #1113, opened 2024-03-08
feat: add support of float32 type

rahul2393:
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-spanner/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Code coverage does not decrease (if any source code was changed)


tests/unit/test__helpers.py
```diff
@@ -466,6 +466,27 @@ def test_w_float_str(self):
 
         self.assertEqual(self._callFUT(value_pb, field_type), expected_value)
 
+    def test_w_float32(self):
```
harshachinta, line 469: As per FLOAT32 design, the request from client is sent as `number_value` field. What will happen when the input is `float("-inf")` or `float("nan")` and the column in table is of type FLOAT32?

google/cloud/spanner_v1/_helpers.py
```diff
@@ -228,6 +228,11 @@ def _parse_value_pb(value_pb, field_type):
             return float(value_pb.string_value)
         else:
             return value_pb.number_value
+    elif type_code == TypeCode.FLOAT32:
```
harshachinta, line 231: We have integration tests for every type in [test_session_api.py](https://github.com/googleapis/python-spanner/blob/e73c6718b23bf78a8f264419b2ba378f95fa2554/tests/system/test_session_api.py#L2182). Can we have a test for `FLOAT32`?
Reference: `test_execute_sql_w_float64_bindings` and `test_execute_sql_w_float_bindings_transfinite`

The comment mentioned [below](https://github.com/googleapis/python-spanner/pull/1113/files#r1518576993) can be verified through the test `test_execute_sql_w_float_bindings_transfinite`

tests/unit/test__helpers.py
harshachinta: We need to verify both insert and read with this data when column type is FLOAT32.
For read we already have an integration test for FLOAT64 `test_execute_sql_w_float_bindings_transfinite `. We can copy it for FLOAT32.

olavloite:
We should register the same merge method for `FLOAT32` as for `FLOAT64` here: https://github.com/googleapis/python-spanner/blob/e73c6718b23bf78a8f264419b2ba378f95fa2554/google/cloud/spanner_v1/streamed.py#L334

tests/system/_sample_data.py
```diff
@@ -90,5 +90,8 @@ def _check_cell_data(found_cell, expected_cell, recurse_into_lists=True):
         for found_item, expected_item in zip(found_cell, expected_cell):
             _check_cell_data(found_item, expected_item)
 
+    elif isinstance(found_cell, float):
+        assert abs(found_cell - expected_cell) < 0.00001
```
rahul2393, line 94: Similar to JAVA: https://github.com/googleapis/java-spanner/blob/main/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITFloat32Test.java#L357C42-L357C51

tests/unit/test__helpers.py
rahul2393: behaviour for `float("-inf")` or `float("nan")` is same as FLOAT64, added the skipped test case for manual verification.

olavloite (approved):

tests/system/test_session_api.py
```diff
@@ -2216,6 +2216,47 @@ def test_execute_sql_w_float_bindings_transfinite(sessions_database, database_di
...
+    pytest.skip(f"float32 is not yet supported in production.")
+    key = "p1" if database_dialect == DatabaseDialect.POSTGRESQL else "neg_inf"
+    placeholder = "$1" if database_dialect == DatabaseDialect.POSTGRESQL else f"@{key}"
+
+    # Find -inf
+    _check_sql_results(
+        sessions_database,
+        sql=f"SELECT {placeholder}",
+        params={key: NEG_INF},
+        param_types={key: spanner_v1.param_types.FLOAT64},
```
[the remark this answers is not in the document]
olavloite: Nvm, I see that it has already been changed

harshachinta (approved):

PR #1120, opened 2024-03-27
fix: use stdlib warnings module instead of a third party dependency

asottile-sentry:
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-spanner/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea -- https://github.com/googleapis/python-spanner/pull/1037#discussion_r1520126879
- [x] Code coverage does not decrease (if any source code was changed)

see also https://github.com/googleapis/python-spanner/pull/1037#discussion_r1520126879


ankiaga (approved):
Thanks for making the change!

parthea (approved):

PR #1152, opened 2024-06-18
feat(spanner): add support for txn changstream exclusion

harshachinta:
Add support for excluding transactions from being recorded in the change streams by passing a new boolean option ExcludeTxnFromChangeStreams in the various write APIs:

1. run_in_transaction
2. batch
3. batch_write
4. execute_partitioned_dml

### DIFF
--- a/google/cloud/spanner_v1/batch.py
+++ b/google/cloud/spanner_v1/batch.py
@@ -287,6 +287,13 @@ class MutationGroups(_SessionWrapper):
                 If a dict is provided, it must be of the same form as the protobuf
                 message :class:`~google.cloud.spanner_v1.types.RequestOptions`.
 
+        :type exclude_txn_from_change_streams: bool
+        :param exclude_txn_from_change_streams:
+          If true, instructs the transaction to be excluded from being recorded in change streams
+          with the DDL option `allow_txn_exclusion=true`. This does not exclude the transaction from
+          being recorded in the change streams with the DDL option `allow_txn_exclusion` being false or
+          unset.
+
         :rtype: :class:`Iterable[google.cloud.spanner_v1.types.BatchWriteResponse]`
         :returns: a sequence of responses for each batch.
         """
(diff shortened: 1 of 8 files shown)


google/cloud/spanner_v1/batch.py
```diff
@@ -147,7 +147,11 @@ def _check_state(self):
             raise ValueError("Batch already committed")
 
     def commit(
-        self, return_commit_stats=False, request_options=None, max_commit_delay=None
+        self,
+        return_commit_stats=False,
+        request_options=None,
+        max_commit_delay=None,
+        exclude_txn_from_change_streams=None,
```
surbhigarg92, line 154: Default should be False not None , Right ?

google/cloud/spanner_v1/database.py
```diff
@@ -771,10 +785,19 @@ def batch(self, request_options=None, max_commit_delay=None):
                 in order to improve throughput. Value must be between 0ms and
                 500ms.
 
+        :type exclude_txn_from_change_streams: bool
+        :param exclude_txn_from_change_streams:
+          If true, instructs the transaction to be excluded from being recorded in change streams
```
surbhigarg92, line 790: Add `(Optional) in the documentation

tests/unit/test_session.py
```diff
@@ -1696,6 +1696,191 @@ def unit_of_work(txn, *args, **kw):
             ],
         )
 
+    def test_run_in_transaction_w_exclude_txn_from_change_streams(self):
```
surbhigarg92, line 1699: Can you also add a test for ILB. use case, when there is no begin transaction and "ExecuteSql" should have `exclude_txn_from_change_streams` set

google/cloud/spanner_v1/batch.py
harshachinta: Since the type of the field is boolean (primitive) it always gets set to false by default. Eventhough you set None it is like setting false. However, updated all references to False

google/cloud/spanner_v1/database.py
harshachinta: Done

tests/unit/test_session.py
harshachinta: Added

surbhigarg92 (approved):

PR #1156, opened 2024-07-05
chore: update templated files

parthea:
- Use the latest version of the post processor which includes updates to the release script dependencies as well as testing for [all protobuf implementations](https://github.com/protocolbuffers/protobuf/tree/main/python#implementation-backends)

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright 2023 Google LLC
+# Copyright 2024 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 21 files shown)


noxfile.py
```diff
@@ -184,34 +184,48 @@ def default(session):
...
-    session.run("pip", "install", "--upgrade", "pip")
-    session.install("-e", ".[tracing]", "-c", constraints_path)
-    # XXX: Dump installed versions to debug OT issue
-    session.run("pip", "list")
 
-    # Run py.test against the unit tests with OpenTelemetry.
+@nox.session(python=UNIT_TEST_PYTHON_VERSIONS)
+@nox.parametrize(
+    "protobuf_implementation",
+    ["python", "upb", "cpp"],
```
harshachinta, line 191: @parthea Just curious to understand what does this `protobuf_implementation` mean and why do we have 3 options `["python", "upb", "cpp"]`?

harshachinta:
@parthea Can you please add PR description for this change so that it will help understand for someone looking at this change.

harshachinta (approved):

noxfile.py
parthea: This is a feature of `protobuf` where they offer [three separate implementations](https://github.com/protocolbuffers/protobuf/tree/main/python#implementation-backends) of the `Python Protobuf` API. There is an expectation that we should support all 3 implementations in the Cloud Python SDK.

`upb` and `python ` are actively supported in the latest `Protobuf 5.x` while `cpp` is only available in `Protobuf 3.x`

parthea:
> Can you please help add PR description for this change so that it will help understand for someone looking at this change.

Done

> Also what changes will get affected with this?

This only touches the release script dependencies and updates to testing to include all [protobuf implementation backends](https://github.com/protocolbuffers/protobuf/tree/main/python#implementation-backends). There are no user facing changes.

PR #1168, opened 2024-07-24
feat(spanner): add samples for instance partitions

varuncnaik:


samples/samples/snippets_test.py
```diff
@@ -188,6 +195,33 @@ def test_create_instance_with_autoscaling_config(capsys, lci_instance_id):
     retry_429(instance.delete)()
 
 
+def test_create_instance_partition(capsys, instance_partition_instance_id):
+    spanner_client = spanner.Client()
+    operation = spanner_client.instance_admin_api.create_instance(
```
harshachinta, line 200: We can re-use existing instance that is already created in https://github.com/googleapis/python-spanner/blob/ec96d36fca020e17ebb864dab96c5f25e5100a30/samples/samples/snippets_test.py#L149.

This will prevent instance creation time when running tests and also reduce the no of instances that get created.
This test can be marked as dependent to the instance creation test like https://github.com/googleapis/python-spanner/blob/ec96d36fca020e17ebb864dab96c5f25e5100a30/samples/samples/snippets_test.py#L226.

varuncnaik: I implemented Approach 2. The challenge with reusing an existing instance is that there are several restrictions around instance partitions. Notably:
* An instance partition cannot have the same instance config as the parent instance
* Several Spanner features do not work in instances where the user has created one or more instance partitions.

harshachinta (approved):

PR #1172, opened 2024-07-25
fix: use modernized and standardized OpenTelemetry when tracing

### LINKED ISSUE #1170: spanner_v1/trace: trace_call span attributes are hard-coded and don't follow OpenTelemetry semantic conventions; should instead use incorrect standardized OpenTelemetry attributes
If we examine the code in here https://github.com/googleapis/python-spanner/blob/fe20d41e6ee333290fa461f40811c791ca1655fd/google/cloud/spanner_v1/_opentelemetry_tracing.py#L41-L46

## Fixes
We should import the package `opentelemetry.semconv.attributes` and from it we can use the following imported variables
- [ ] "db.type" should be DB_SYSTEM aka "db.system" = "google.cloud.spanner"
- [ ] "db.instance_name" should be DB_NAME aka "db.name"
- [ ] "db.url" should use DB_CONNECTION_STRING aka "db.connection_string"
- [ ] "net.host.name" should use NET_HOST_NAME

The code in question was written in 2020, long before OpenTelemetry standardized semantic conventions hence I am sending over a change pretty shortly to fix this!

### LINKED ISSUE #1171: spanner_v1/trace: allow option for extended tracing which will allow recording the SQL statement if provided
Part of modernized tracing is that it becomes pretty useful for customers to investigate slow queries when observing traces; customers would like to opt-in to ensure that this library records their SQL statements in their traces where applicable e.g. if they've explicitly opted into having their spans annotated with the SQL statement; explicitly because some SQL statement might have personally identifiable information (PII) and leak sensitive data out.

I am going to work on this issue myself.

### LINKED ISSUE #1173: observability: please update to a newer version of opentelemetry because the current dependency opentelemetry-api:1.1.0 is from April 2021 (3+ years ago)
While working on modernizing the observability in this library I encountered the fact that the constrained minimum version of opentelemetry-api being used is 1.1.0 and that was released  in April 2021 (3 years ago) https://pypi.org/project/opentelemetry-api/1.1.0/
per
https://github.com/googleapis/python-spanner/blob/fe20d41e6ee333290fa461f40811c791ca1655fd/setup.py#L49-L53

Let's get to the latest release of opentelemetry. I am going to send a change

odeke-em:
This change modernizes trace span attributes by using OpenTelemetry's
semantic conventions that are standardized and allow for much
better common ground adoption by broader systems, even more as
Google Cloud Tracing & Monitoring pushes towards OpenTelemetry more.

With this change we've made the replacement of these fields, directly
with imports from `opentelemetry.semconv.trace.SpanAttributes`, as:

* "db.type"       => DB_SYSTEM aka "db.system"
* "db.url"        => DB_CONNECTION_STRING aka "db.connection_string"
* "db.instance"   => DB_NAME aka "db.name"
* "net.host.name" => NET_HOST_NAME aka "net.host.name"

While here, also updated opentelemetry-(api, sdk) dependencies to use
versions "1.25.0", then opentelemetry-(instrumentation) to "0.46b0"

Also added in an option toggled by environment variable

    ENABLE_EXTENDED_TRACING=true

which allows spans to be annotated with the SQL statement keyed by

    "db.statement"

Fixes #1170
Fixes #1171
Fixes #1173

## Exhibit
With the change made, we now have this trace with attributes visualized from Google Cloud Tracing per
Screenshot 2024-07-26 at 7 12 42 PM
Screenshot 2024-07-26 at 7 11 10 PM

### DIFF
--- a/.github/workflows/integration-tests-against-emulator.yaml
+++ b/.github/workflows/integration-tests-against-emulator.yaml
@@ -24,9 +24,17 @@ jobs:
           python-version: 3.8
       - name: Install nox
         run: python -m pip install nox
-      - name: Run system tests
+      - name: Run system tests (without extended tracing)
         run: nox -s system
         env:
           SPANNER_EMULATOR_HOST: localhost:9010
           GOOGLE_CLOUD_PROJECT: emulator-test-project
           GOOGLE_CLOUD_TESTS_CREATE_SPANNER_INSTANCE: true
+          EXTENDED_TRACING_ENABLED: false
+      - name: Run system tests (with extended tracing)
+        run: nox -s system
+        env:
+          SPANNER_EMULATOR_HOST: localhost:9010
+          GOOGLE_CLOUD_PROJECT: emulator-test-project
+          GOOGLE_CLOUD_TESTS_CREATE_SPANNER_INSTANCE: true
+          EXTENDED_TRACING_ENABLED: true
--- a/docs/opentelemetry-tracing.rst
+++ b/docs/opentelemetry-tracing.rst
@@ -19,9 +19,9 @@ We also need to tell OpenTelemetry which exporter to use. To export Spanner trac
     from opentelemetry.sdk.trace import TracerProvider
     from opentelemetry.sdk.trace.sampling import TraceIdRatioBased
     from opentelemetry.exporter.cloud_trace import CloudTraceSpanExporter
-    # BatchExportSpanProcessor exports spans to Cloud Trace 
+    # BatchSpanProcessor exports spans to Cloud Trace
     # in a seperate thread to not block on the main thread
-    from opentelemetry.sdk.trace.export import BatchExportSpanProcessor
+    from opentelemetry.sdk.trace.export import BatchSpanProcessor
 
     # Create and export one trace every 1000 requests
     sampler = TraceIdRatioBased(1/1000)
@@ -29,7 +29,7 @@ We also need to tell OpenTelemetry which exporter to use. To export Spanner trac
     trace.set_tracer_provider(TracerProvider(sampler=sampler))
     trace.get_tracer_provider().add_span_processor(
         # Initialize the cloud tracing exporter
-        BatchExportSpanProcessor(CloudTraceSpanExporter())
+        BatchSpanProcessor(CloudTraceSpanExporter())
     )
 
 
@@ -46,7 +46,7 @@ Alternatively you can pass in a tracer provider into the Cloud Spanner initializ
     tracerProvider = TracerProvider(sampler=sampler)
     tracerProvider.add_span_processor(
         # Initialize the cloud tracing exporter
-        BatchExportSpanProcessor(CloudTraceSpanExporter())
+        BatchSpanProcessor(CloudTraceSpanExporter())
     )
 
     o11y = dict(tracer_provider=tracerProvider)
--- a/docs/opentelemetry-tracing.rst
+++ b/docs/opentelemetry-tracing.rst
@@ -9,6 +9,7 @@ To take advantage of these traces, we first need to install OpenTelemetry:
 .. code-block:: sh
 
     pip install opentelemetry-api opentelemetry-sdk opentelemetry-instrumentation
+    pip install opentelemetry-exporter-google-cloud
 
     # [Optional] Installs the cloud monitoring exporter, however you can use any exporter of your choice
     pip install opentelemetry-exporter-google-cloud
@@ -19,14 +20,14 @@ We also need to tell OpenTelemetry which exporter to use. To export Spanner trac
 
     from opentelemetry import trace
     from opentelemetry.sdk.trace import TracerProvider
-    from opentelemetry.trace.sampling import ProbabilitySampler
+    from opentelemetry.sdk.trace.sampling import TraceIdRatioBased
     from opentelemetry.exporter.cloud_trace import CloudTraceSpanExporter
     # BatchExportSpanProcessor exports spans to Cloud Trace 
     # in a seperate thread to not block on the main thread
     from opentelemetry.sdk.trace.export import BatchExportSpanProcessor
 
     # Create and export one trace every 1000 requests
-    sampler = ProbabilitySampler(1/1000)
+    sampler = TraceIdRatioBased(1/1000)
     # Use the default tracer provider
     trace.set_tracer_provider(TracerProvider(sampler=sampler))
     trace.get_tracer_provider().add_span_processor(
@@ -38,3 +39,6 @@ Generated spanner traces should now be available on `Cloud Trace <https://consol
 
 Tracing is most effective when many libraries are instrumented to provide insight over the entire lifespan of a request.
 For a list of libraries that can be instrumented, see the `OpenTelemetry Integrations` section of the `OpenTelemetry Python docs <https://opentelemetry-python.readthedocs.io/en/stable/>`_
+
+To allow for SQL statements to be annotated in your spans, please set
+the environment variable `SPANNER_ENABLE_EXTENDED_TRACING=true`.
--- a/docs/opentelemetry-tracing.rst
+++ b/docs/opentelemetry-tracing.rst
@@ -33,7 +33,8 @@ We also need to tell OpenTelemetry which exporter to use. To export Spanner trac
     )
 
 
-Alternatively you can pass in a tracer provider into the Cloud Spanner initialization
+Alternatively you can pass in a tracer provider into the Cloud Spanner
+initialization, otherwise the global tracer will be used:
 
 .. code:: python
 
@@ -49,17 +50,11 @@ Alternatively you can pass in a tracer provider into the Cloud Spanner initializ
         BatchSpanProcessor(CloudTraceSpanExporter())
     )
 
-    o11y = dict(tracer_provider=tracerProvider)
+    options = dict(tracer_provider=tracerProvider)
     # Pass the tracer provider while creating the Spanner client.
-    spanner_client = spanner.Client(observability_options=o11y)
+    spanner_client = spanner.Client(observability_options=options)
     instance = spanner_client.instance(instance_id)
     database = instance.database(database_id)
-
-please note that the tracer being used is retrieved by invoking:
-
-.. code:: python
-
-   tracer = tracerProvider.get_tracer('cloud.google.com/python/spanner', SPANNER_LIB_VERSION)
 
 
 To get more fine-grained traces from gRPC, you can enable the gRPC instrumentation by the following
--- a/docs/opentelemetry-tracing.rst
+++ b/docs/opentelemetry-tracing.rst
@@ -9,10 +9,7 @@ To take advantage of these traces, we first need to install OpenTelemetry:
 .. code-block:: sh
 
     pip install opentelemetry-api opentelemetry-sdk opentelemetry-instrumentation
-    pip install opentelemetry-exporter-google-cloud
-
-    # [Optional] Installs the cloud monitoring exporter, however you can use any exporter of your choice
-    pip install opentelemetry-exporter-google-cloud
+    pip install opentelemetry-exporter-gcp-trace
 
 We also need to tell OpenTelemetry which exporter to use. To export Spanner traces to `Cloud Tracing <https://cloud.google.com/trace>`_, add the following lines to your application:
 
@@ -57,6 +54,12 @@ Alternatively you can pass in a tracer provider into the Cloud Spanner initializ
     spanner_client = spanner.Client(observability_options=o11y)
     instance = spanner_client.instance(instance_id)
     database = instance.database(database_id)
+
+please not that the tracer being used is retrieved by invoking:
+
+.. code:: python
+
+   tracer = tracerProvider.get_tracer('cloud.google.com/python/spanner', SPANNER_LIB_VERSION)
 
 Generated spanner traces should now be available on `Cloud Trace <https://console.cloud.google.com/traces>`_.
 
--- a/docs/opentelemetry-tracing.rst
+++ b/docs/opentelemetry-tracing.rst
@@ -83,4 +83,10 @@ Tracing is most effective when many libraries are instrumented to provide insigh
 For a list of libraries that can be instrumented, see the `OpenTelemetry Integrations` section of the `OpenTelemetry Python docs <https://opentelemetry-python.readthedocs.io/en/stable/>`_
 
 To allow for SQL statements to be annotated in your spans, please set
-the environment variable `SPANNER_ENABLE_EXTENDED_TRACING=true` or please set the configuration field `enable_extended_tracing` to `True` when configuring the Cloud Spanner client.
+the environment variable `SPANNER_ENABLE_EXTENDED_TRACING=true` or please set the configuration field `enable_extended_tracing` to `True` when configuring the Cloud Spanner client, like this:
+
+.. code:: python
+
+   tracerProvider = TracerProvider(sampler=sampler)
+   opts = dict(tracer_provider=tracerProvider, enable_extended_tracing=true)
+   spanner_client = spanner.Client(project_id, observability_options=opts)
--- a/docs/opentelemetry-tracing.rst
+++ b/docs/opentelemetry-tracing.rst
@@ -35,6 +35,29 @@ We also need to tell OpenTelemetry which exporter to use. To export Spanner trac
         BatchExportSpanProcessor(CloudTraceSpanExporter())
     )
 
+
+Alternatively you can pass in a tracer provider into the Cloud Spanner initialization
+
+.. code:: python
+
+    from opentelemetry.sdk.trace import TracerProvider
+    from opentelemetry.sdk.trace.sampling import TraceIdRatioBased
+    from opentelemetry.exporter.cloud_trace import CloudTraceSpanExporter
+
+    # Create and export one trace every 1000 requests
+    sampler = TraceIdRatioBased(1/1000)
+    tracerProvider = TracerProvider(sampler=sampler)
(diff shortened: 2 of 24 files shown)


odeke-em:
@harshachinta is it possible for you to hep me authorize workflows to have the tests running on every push? Thank you.

odeke-em:
@harshachinta please see the updated and rebased code.

.github/workflows/integration-tests-against-emulator.yaml
```diff
@@ -24,9 +24,17 @@ jobs:
...
         run: python -m pip install nox
-      - name: Run system tests
+      - name: Run system tests (without extended tracing)
         run: nox -s system
         env:
           SPANNER_EMULATOR_HOST: localhost:9010
           GOOGLE_CLOUD_PROJECT: emulator-test-project
           GOOGLE_CLOUD_TESTS_CREATE_SPANNER_INSTANCE: true
+          EXTENDED_TRACING_ENABLED: false
+      - name: Run system tests (with extended tracing)
```
harshachinta, line 34: Why do we need a seperate job for extended tracing? What are we validating with the help of this job?

docs/opentelemetry-tracing.rst
```diff
@@ -9,6 +9,7 @@ To take advantage of these traces, we first need to install OpenTelemetry:
 .. code-block:: sh
 
     pip install opentelemetry-api opentelemetry-sdk opentelemetry-instrumentation
+    pip install opentelemetry-exporter-google-cloud
 
     # [Optional] Installs the cloud monitoring exporter, however you can use any exporter of your choice
     pip install opentelemetry-exporter-google-cloud
```
harshachinta, line 15: This package is deprecated and will not receive any more updates. Also, this line is already available below and is a duplicate here.
https://pypi.org/project/opentelemetry-exporter-google-cloud/

Replace this with `pip install opentelemetry-exporter-gcp-trace`

examples/trace.py
```diff
@@ -0,0 +1,61 @@
...
+from opentelemetry.sdk.trace.sampling import ALWAYS_ON
+from opentelemetry import trace
+
+
+def main():
+    # Setup OpenTelemetry, trace and Cloud Trace exporter.
+    sampler = ALWAYS_ON
+    tracerProvider = TracerProvider(sampler=sampler)
+    tracerProvider.add_span_processor(
+        BatchExportSpanProcessor(CloudTraceSpanExporter()))
```
harshachinta, line 32: Pass `project_id` to `CloudTraceSpanExporter()`?

google/cloud/spanner_v1/_opentelemetry_tracing.py
```diff
@@ -15,19 +15,39 @@
...
+    DB_SYSTEM = SpanAttributes.DB_SYSTEM
+    DB_NAME = SpanAttributes.DB_NAME
+    DB_CONNECTION_STRING = SpanAttributes.DB_CONNECTION_STRING
+    NET_HOST_NAME = SpanAttributes.NET_HOST_NAME
+    DB_STATEMENT = SpanAttributes.DB_STATEMENT
 except ImportError:
     HAS_OPENTELEMETRY_INSTALLED = False
 
 
+EXTENDED_TRACING_ENABLED = os.environ.get('SPANNER_ENABLE_EXTENDED_TRACING', '') == 'true'
```
harshachinta, line 38: In addition to env variable, customer should be given an option to set it via an option during client initialization.

google/cloud/spanner_v1/_opentelemetry_tracing.py
```diff
@@ -15,19 +15,39 @@
...
+
+
+def annotate_with_sql_statement(span, sql):
```
harshachinta, line 41: Where is this getting called?

google/cloud/spanner_v1/_opentelemetry_tracing.py
```diff
@@ -15,19 +15,39 @@
...
+    to the sql statement, only if SPANNER_ENABLE_EXTENDED_TRACING=true
+    is set in the environment.
+    """
+    if EXTENDED_TRACING_ENABLED:
+        span.set_attribute(DB_STATEMENT, sql)
+
+
 @contextmanager
 def trace_call(name, session, extra_attributes=None):
     if not HAS_OPENTELEMETRY_INSTALLED or not session:
```
harshachinta, line 53: 1. We are using global tracer provider here by default. But we should also give an option for customer to pass customer tracer provider for spanner traces and fallback to global if not provided.
2. Set the instrumenting_module_name to `cloud.google.com/language`, instrumenting_library_version when calling get_tracer()

google/cloud/spanner_v1/_opentelemetry_tracing.py
```diff
@@ -39,15 +59,18 @@ def trace_call(name, session, extra_attributes=None):
...
     # Set base attributes that we know for every trace created
     attributes = {
-        "db.type": "spanner",
-        "db.url": SpannerClient.DEFAULT_ENDPOINT,
-        "db.instance": session._database.name,
-        "net.host.name": SpannerClient.DEFAULT_ENDPOINT,
+        DB_SYSTEM: "spanner",
+        DB_CONNECTION_STRING: SpannerClient.DEFAULT_ENDPOINT,
+        DB_NAME: session._database.name,
+        NET_HOST_NAME: SpannerClient.DEFAULT_ENDPOINT,
```
harshachinta, line 65: This can be a breaking change to customers already using these attributes. Lets discuss about this tomorrow.

.github/workflows/integration-tests-against-emulator.yaml
odeke-em: It's to ensure that extended tracing works as expected because there is no other mechanism to test for it.

google/cloud/spanner_v1/_opentelemetry_tracing.py
odeke-em: I removed it, thank you!

google/cloud/spanner_v1/_opentelemetry_tracing.py
odeke-em: Roger that, I am addressing that shortly!

google/cloud/spanner_v1/_opentelemetry_tracing.py
odeke-em: All done! Thank you @harshachinta!

google/cloud/spanner_v1/_opentelemetry_tracing.py
odeke-em: Done!

docs/opentelemetry-tracing.rst
```diff
@@ -9,32 +9,84 @@ To take advantage of these traces, we first need to install OpenTelemetry:
...
+    # Pass the tracer provider while creating the Spanner client.
+    spanner_client = spanner.Client(observability_options=o11y)
+    instance = spanner_client.instance(instance_id)
+    database = instance.database(database_id)
+
+please note that the tracer being used is retrieved by invoking:
+
+.. code:: python
+
+   tracer = tracerProvider.get_tracer('cloud.google.com/python/spanner', SPANNER_LIB_VERSION)
```
harshachinta, line 62: As per opentelemetry, this module name `cloud.google.com/python/spanner ` and version `SPANNER_LIB_VERSION ` are used to identify which library is the trace getting generated from. For ex, traces generated in python spanner client will be identified through these variables.

For customers if they want to add custom trace in their application, they will need to use their package name and application version that will help them understand that the custom trace is getting emited from their application.
There is no usecase for customer to retrieve the tracer with these python spanner package names. So you can remove this example here

docs/opentelemetry-tracing.rst
```diff
@@ -9,32 +9,84 @@ To take advantage of these traces, we first need to install OpenTelemetry:
...
     # Use the default tracer provider
     trace.set_tracer_provider(TracerProvider(sampler=sampler))
     trace.get_tracer_provider().add_span_processor(
         # Initialize the cloud tracing exporter
-        BatchExportSpanProcessor(CloudTraceSpanExporter())
+        BatchSpanProcessor(CloudTraceSpanExporter())
+    )
+
+
+Alternatively you can pass in a tracer provider into the Cloud Spanner initialization
```
harshachinta, line 36: Can we mention here that if the tracer provider is. not provided then the python spanner client will use the configured global tracer provider?

docs/opentelemetry-tracing.rst
```diff
@@ -9,32 +9,84 @@ To take advantage of these traces, we first need to install OpenTelemetry:
...
+
+    # Create and export one trace every 1000 requests
+    sampler = TraceIdRatioBased(1/1000)
+    tracerProvider = TracerProvider(sampler=sampler)
+    tracerProvider.add_span_processor(
+        # Initialize the cloud tracing exporter
+        BatchSpanProcessor(CloudTraceSpanExporter())
     )
 
+    o11y = dict(tracer_provider=tracerProvider)
```
harshachinta, line 52: can we rename the variable to some meaningful name?

examples/grpc_instrumentation_enabled.py
```diff
@@ -0,0 +1,71 @@
...
+
+    # Setup OpenTelemetry, trace and Cloud Trace exporter.
+    sampler = ALWAYS_ON
+    tracerProvider = TracerProvider(sampler=sampler)
+    traceExporter = CloudTraceSpanExporter(project_id)
+    tracerProvider.add_span_processor(
+        BatchSpanProcessor(traceExporter))
+    trace.set_tracer_provider(tracerProvider)
+    # Retrieve the set shared tracer.
+    tracer = tracerProvider.get_tracer('cloud.google.com/python/spanner', spanner.__version__)
```
harshachinta, line 43: I dont think this is needed. In this case `QueryInformationSchema` span will be shown as being generated from python spanner client, but it is generated from customer application.

Can you update the module name and version name to application name and version and see how the output changes?
Can you share me the results of both, with change and without change?

google/cloud/spanner_v1/_opentelemetry_tracing.py
```diff
@@ -15,41 +15,89 @@
...
         attributes.update(extra_attributes)
 
+    extended_tracing = EXTENDED_TRACING_ENABLED or (
+                        observability_options and
+                        observability_options.get('enable_extended_tracing', False))
+
+    if not extended_tracing:
+        attributes.pop(DB_STATEMENT, None)
+
+    span_name = SPAN_NAMESPACE + '/' + name
```
harshachinta, line 98: What is the trace span output with this change?
I assume the span name will change from `CloudSpanner.CreateSession` to `cloud.google.com/python/spanner/CloudSpanner.CreateSession`. If yes this change is not needed, span name can be as it is and need not be prefixed with namespace.

google/cloud/spanner_v1/batch.py
```diff
@@ -205,7 +205,8 @@ def commit(
             max_commit_delay=max_commit_delay,
             request_options=request_options,
         )
-        with trace_call("CloudSpanner.Commit", self._session, trace_attributes):
+        with trace_call("Batch.commit", self._session, trace_attributes,
```
harshachinta, line 208: Can we follow the span naming as per this document?
https://docs.google.com/document/d/1C-2WbnKfAtl-GgOebVNYv0RdYFkb_UfyzagwJOQgxAs/edit?tab=t.0#bookmark=id.ctslkgwt101t

I would suggest to keep the span naming changes in a seperate PR. This will make PR small and easy to review. We will focus on getting this PR out.

google/cloud/spanner_v1/session.py
```diff
@@ -63,12 +63,13 @@ class Session(object):
     _session_id = None
     _transaction = None
 
-    def __init__(self, database, labels=None, database_role=None):
+    def __init__(self, database, labels=None, database_role=None, observability_options=None):
```
harshachinta, line 66: Just to clarify is this method used internally or by customers?
If only used internally, I don't see that we are passing observability_options to this method that is received through spanner client.

google/cloud/spanner_v1/__init__.py
```diff
@@ -88,6 +89,7 @@
     # google.cloud.spanner_v1
     "__version__",
     "param_types",
+    "get_tracer",
```
harshachinta, line 92: Why do we need this here?

.github/workflows/integration-tests-against-emulator.yaml
harshachinta: EXTENDED_TRACING_ENABLED will add sql statements in span attributes right?
I don't think we need to run a job explicitly for that if we don't have any tests specifically to validate this behavior in system tests.

google/cloud/spanner_v1/__init__.py
harshachinta: This is not something that customer will need. We can remove this

google/cloud/spanner_v1/_opentelemetry_tracing.py
```diff
@@ -15,41 +15,89 @@
...
+def get_tracer(tracer_provider=None):
+    """
+    get_tracer is a utility to unify and simplify retrieval of the tracer, without
+    leaking implementation details given that retrieving a tracer requires providing
+    the full qualified library name and version.
+    When the tracer_provider is set, it'll retrieve the tracer from it, otherwise
+    it'll fall back to the global tracer provider and use this library's specific semantics.
+    """
+    if tracer_provider:
+        return tracerProvider.get_tracer(TRACER_NAME, TRACER_VERSION)
```
harshachinta, line 53:
```suggestion
        return tracer_provider.get_tracer(TRACER_NAME, TRACER_VERSION)
```

google/cloud/spanner_v1/_opentelemetry_tracing.py
```diff
@@ -15,41 +15,89 @@
...
+def trace_call(name, session, extra_attributes=None, observability_options=None):
     if not HAS_OPENTELEMETRY_INSTALLED or not session:
         # Empty context manager. Users will have to check if the generated value is None or a span
         yield None
         return
 
-    tracer = trace.get_tracer(__name__)
+    tracer = None
+    if observability_options:
+        tracerProvider = observability_options.get('tracer_provider', None)
```
harshachinta, line 66: variable should be lower case

google/cloud/spanner_v1/_opentelemetry_tracing.py
```diff
@@ -15,41 +15,89 @@
...
+        spanner_endpoint = SpannerClient.DEFAULT_ENDPOINT
 
     # Set base attributes that we know for every trace created
     attributes = {
-        "db.type": "spanner",
-        "db.url": SpannerClient.DEFAULT_ENDPOINT,
-        "db.instance": session._database.name,
-        "net.host.name": SpannerClient.DEFAULT_ENDPOINT,
+        DB_SYSTEM: "spanner",
+        DB_CONNECTION_STRING: spanner_endpoint
```
harshachinta, line 83:
```suggestion
        DB_CONNECTION_STRING: spanner_endpoint,
```

google/cloud/spanner_v1/_opentelemetry_tracing.py
harshachinta, line 53: I tried to run one sample and see that `otel.scope.name` is not getting populated.
image

It should look something like below (this was in java)
image

google/cloud/spanner_v1/_opentelemetry_tracing.py
```diff
@@ -15,41 +15,89 @@
...
 from google.cloud.spanner_v1 import SpannerClient
+from google.cloud.spanner_v1 import gapic_version as TRACER_VERSION
 
 try:
     from opentelemetry import trace
     from opentelemetry.trace.status import Status, StatusCode
+    from opentelemetry.semconv.trace import SpanAttributes
 
     HAS_OPENTELEMETRY_INSTALLED = True
+    DB_SYSTEM = SpanAttributes.DB_SYSTEM
```
harshachinta, line 30: 1. We are changing attribute conventions here. As discussed earlier do we need to make this change under a seperate option to avoid breaking changes?
2. I also see that we might need to change span names to be consistent with this [document](https://docs.google.com/document/d/1C-2WbnKfAtl-GgOebVNYv0RdYFkb_UfyzagwJOQgxAs/edit?tab=t.0#bookmark=id.ctslkgwt101t), which is anyhow a breaking change.

So should we make these changes directly or instead protect them under an option?
My opinion is since we are refactoring observability portion, lets make all changes directly and warn the customers in release notes. Having a seperate option will make code messier.
@surbhigarg92 Can you give your thoughts on this?

google/cloud/spanner_v1/batch.py
harshachinta: @odeke-em 
I am in favour of not changing the span names since that could introduce breaking changes.

harshachinta: I would also suggest if you can take a look at the python spanner code and suggest any additional spans that could help customer for better debugging. 
You can also look at this document to check the existing spans in other languages to get some idea.
https://docs.google.com/document/d/1C-2WbnKfAtl-GgOebVNYv0RdYFkb_UfyzagwJOQgxAs/edit?tab=t.0#bookmark=id.ctslkgwt101t

google/cloud/spanner_v1/_opentelemetry_tracing.py
harshachinta: @odeke-em 
1. For attribute conventions lets enable it under a seperate flag to avoid breaking changes.
2. Lets not change any existing span names as they could bring in a breaking change. Instead can you help add additional spans that could help in debugging.

odeke-em:
Superseded by https://github.com/googleapis/python-spanner/pull/1199

PR #1199, opened 2024-09-11
tracing: update OpenTelemetry dependencies from 2021 to 2024

### LINKED ISSUE #1173: observability: please update to a newer version of opentelemetry because the current dependency opentelemetry-api:1.1.0 is from April 2021 (3+ years ago)

odeke-em:
This change non-invasively introduces dependencies of opentelemetry
bringing in the latest dependencies and modernizing them.

While here also brought in modern span attributes:
* otel.scope.name
* otel.scope.version

Also added a modernized example and updated the docs.

Updates #1170
Fixes #1173
Built from PR #1172

### DIFF
--- a/examples/trace.py
+++ b/examples/trace.py
@@ -1,0 +1,66 @@
+# -*- coding: utf-8 -*-
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License
+
+import os
+import time
+
+import google.cloud.spanner as spanner
+from opentelemetry.exporter.cloud_trace import CloudTraceSpanExporter
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace.export import BatchSpanProcessor
+from opentelemetry.sdk.trace.sampling import ALWAYS_ON
+from opentelemetry import trace
+
+
+def main():
+    # Setup common variables that'll be used between Spanner and traces.
+    project_id = os.environ.get('SPANNER_PROJECT_ID', 'test-project')
+
+    # Setup OpenTelemetry, trace and Cloud Trace exporter.
+    tracer_provider = TracerProvider(sampler=ALWAYS_ON)
+    trace_exporter = CloudTraceSpanExporter(project_id=project_id)
+    tracer_provider.add_span_processor(BatchSpanProcessor(trace_exporter))
+    trace.set_tracer_provider(tracer_provider)
+    # Retrieve a tracer from the global tracer provider.
+    tracer = tracer_provider.get_tracer('MyApp')
+
+    # Setup the Cloud Spanner Client.
+    spanner_client = spanner.Client(project_id)
+    instance = spanner_client.instance('test-instance')
+    database = instance.database('test-db')
+
+    # Now run our queries
+    with tracer.start_as_current_span('QueryInformationSchema'):
+        with database.snapshot() as snapshot:
+            with tracer.start_as_current_span('InformationSchema'):
+                info_schema = snapshot.execute_sql(
+                    'SELECT * FROM INFORMATION_SCHEMA.TABLES')
+                for row in info_schema:
+                    print(row)
+
+        with tracer.start_as_current_span('ServerTimeQuery'):
+            with database.snapshot() as snapshot:
+                # Purposefully issue a bad SQL statement to examine exceptions
+                # that get recorded and a ERROR span status.
+                try:
+                    data = snapshot.execute_sql('SELECT CURRENT_TIMESTAMPx()')
+                    for row in data:
+                        print(row)
+                except Exception as e:
+                    print(e)
+
+
+if __name__ == '__main__':
+    main()
--- a/google/cloud/spanner_v1/_opentelemetry_tracing.py
+++ b/google/cloud/spanner_v1/_opentelemetry_tracing.py
@@ -16,16 +16,38 @@
 
 from contextlib import contextmanager
 
-from google.api_core.exceptions import GoogleAPICallError
 from google.cloud.spanner_v1 import SpannerClient
+from google.cloud.spanner_v1 import gapic_version
 
 try:
     from opentelemetry import trace
     from opentelemetry.trace.status import Status, StatusCode
+    from opentelemetry.semconv.attributes.otel_attributes import (
+        OTEL_SCOPE_NAME,
+        OTEL_SCOPE_VERSION,
+    )
 
     HAS_OPENTELEMETRY_INSTALLED = True
 except ImportError:
     HAS_OPENTELEMETRY_INSTALLED = False
+
+TRACER_NAME = "cloud.google.com/python/spanner"
+TRACER_VERSION = gapic_version.__version__
+
+
+def get_tracer(tracer_provider=None):
+    """
+    get_tracer is a utility to unify and simplify retrieval of the tracer, without
+    leaking implementation details given that retrieving a tracer requires providing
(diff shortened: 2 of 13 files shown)


odeke-em:
Kindly cc-ing you @harshachinta. /cc @willpoint

odeke-em:
## Exhibit
With this change and the example/trace.py being ran, we get back these traces
Screenshot 2024-09-10 at 7 51 08 PM
Screenshot 2024-09-10 at 7 51 19 PM
Screenshot 2024-09-10 at 7 51 43 PM

google/cloud/spanner_v1/_opentelemetry_tracing.py
```diff
@@ -18,15 +18,40 @@
...
+    leaking implementation details given that retrieving a tracer requires providing
+    the full qualified library name and version.
+    When the tracer_provider is set, it'll retrieve the tracer from it, otherwise
+    it'll fall back to the global tracer provider and use this library's specific semantics.
+    """
+    if not tracer_provider:
+        # Acquire the global tracer provider.
+        tracer_provider = trace.get_tracer_provider()
+
+    return tracer_provider.get_tracer(TRACER_NAME, TRACER_VERSION)
```
harshachinta, line 53: Why do we need this condition? I think get_tracer() by default returns global tracer if tracer_provider is empty.
Check this code 
image

tests/_helpers.py
```diff
@@ -9,12 +9,18 @@
         InMemorySpanExporter,
     )
     from opentelemetry.trace.status import StatusCode
+    from opentelemetry.semconv.attributes import (
+       OTEL_SCOPE_NAME,
+       OTEL_SCOPE_VERSION,
```
harshachinta, line 14: I don't think we need this import here. Remove it and directly use this import in test__open... file

tests/unit/test__opentelemetry_tracing.py
```diff
@@ -11,9 +11,16 @@
...
 from google.api_core.exceptions import GoogleAPICallError
 from google.cloud.spanner_v1 import _opentelemetry_tracing
+from google.cloud.spanner_v1 import gapic_version
 
-from tests._helpers import OpenTelemetryBase, HAS_OPENTELEMETRY_INSTALLED
+from tests._helpers import (
+    OpenTelemetryBase,
+    HAS_OPENTELEMETRY_INSTALLED,
+    OTEL_SCOPE_NAME,
+    OTEL_SCOPE_VERSION,
```
harshachinta, line 20: In this file you can directly import from from opentelemetry.semconv.attributes import (
       OTEL_SCOPE_NAME,
       OTEL_SCOPE_VERSION,
) instead of helpers

examples/trace.py
```diff
@@ -0,0 +1,69 @@
...
+    # Setup OpenTelemetry, trace and Cloud Trace exporter.
+    tracer_provider = TracerProvider(sampler=ALWAYS_ON)
+    trace_exporter = CloudTraceSpanExporter(project_id=project_id)
+    tracer_provider.add_span_processor(BatchSpanProcessor(trace_exporter))
+    trace.set_tracer_provider(tracer_provider)
+    # Retrieve a tracer from the global tracer provider.
+    tracer = tracer_provider.get_tracer('MyApp')
+
+    # Retrieve the tracer.
+    tracer = tracer_provider.get_tracer('MySampleApp')
```
harshachinta, line 40: Looks like line duplication

tests/unit/test__opentelemetry_tracing.py
odeke-em: See that's a confusing request because just right above that line there is a try statement to try to import opentelemetry and if it fails there is a conditional to do nothing if the import fails
https://github.com/googleapis/python-spanner/blob/a941adb987ec6e2d1a1583a5aa1254908292e354/tests/unit/test__opentelemetry_tracing.py#L6-L15

and in the helper, there is an explicit polyfilling for those variables if any import failed as we do for other variables
Screenshot 2024-09-13 at 5 03 39 AM

so it is much more nuanced than just seeming like a duplicate import

tests/_helpers.py
odeke-em: Nope, we do need it as a polyfill when imports fail, please see https://github.com/googleapis/python-spanner/blob/a941adb987ec6e2d1a1583a5aa1254908292e354/tests/unit/test__opentelemetry_tracing.py#L6-L15 and then also
Screenshot 2024-09-13 at 5 03 39 AM aka my comment here https://github.com/googleapis/python-spanner/pull/1199#discussion_r1759036705

google/cloud/spanner_v1/_opentelemetry_tracing.py
odeke-em: I was being explicit, Python isn't particularly known for handling None as a fallback. In your suggestion you are asking me to change that code to
```python
return trace.get_tracer(TRACER_NAME, TRACER_VERSION, tracer_provider)
```
but note that this is much more confusing for readers than
```python
if not tracer_provider:
   tracer_provider = tracer.get_provider()
return tracer_provider.get_tracer(TRACER_NAME, TRACER_VERSION)
```
Should I still change it after my explanation above @harshachinta?

examples/trace.py
odeke-em: Thanks, done!

odeke-em:
@harshachinta kindly please see my replies to your review comments. Thank you!

google/cloud/spanner_v1/_opentelemetry_tracing.py
harshachinta: @surbhigarg92 
any thoughts here?

tests/unit/test__opentelemetry_tracing.py
harshachinta: Thinking out loud,
The polyfill that is used in _helpers.py is only for `HAS_OPENTELEMETRY_INSTALLED` to understand whether opentelemetry is installed or not. 
With your approach where you are hardcoding the 2 attributes `OTEL_SCOPE_NAME,
OTEL_SCOPE_VERSION ` in the catch block, these hardcoding will never get used because when 
1. opentelemetry is not installed then all tests skip. 
2. If opentelemetry is present then the attributes are directly used from semconv. 
 
There is no scenario where a hardcoded attribute will get used.

So, my point is the semconv attributes can be directly imported inside the try block of `tests/unit/test__opentelemetry_tracing.py` similar to `opentelemetry.trace.status import StatusCode` and used where needed.

tests/_helpers.py
harshachinta: look at my comment below

tests/unit/test__opentelemetry_tracing.py
odeke-em: Alright, thank you @harshachinta done

tests/_helpers.py
odeke-em: Done, thanks @harshachinta!

harshachinta (approved):

tests/system/test_session_api.py
```diff
@@ -346,6 +355,11 @@ def _make_attributes(db_instance, **kwargs):
         "net.host.name": "spanner.googleapis.com",
         "db.instance": db_instance,
     }
+
+    if ot_helpers.HAS_OPENTELEMETRY_INSTALLED:
```
harshachinta, line 359: @odeke-em This will add a lot of code duplication to add these 2 attributes since we need them in all the test files. Earlier I was assuming we need it only in one file.
May be we should move this code to _helpers.py file and refer these attributes from there similar to your previous approach. This will remove all these duplicates and these 2 attributes can be directly added to base attributes field without additional conditions.

odeke-em: Yeah it was indeed cumbersome to duplicate that code. Sure let me revert it back to what I had.

odeke-em: Refactor done @harshachinta

odeke-em:
@harshachinta all updates made, kindly help me kick off the Kokoro bots, and thank you for the reviews too!

PR #1201, opened 2024-09-13
test: enable emulator tests for POSTGRESQL dialect

larkee:
Enables emulator testing in POSTGRESQL dialect for system tests to improve test coverage.


google/cloud/spanner_v1/database.py
```diff
@@ -525,7 +525,9 @@ def reload(self):
         self._encryption_config = response.encryption_config
         self._encryption_info = response.encryption_info
         self._default_leader = response.default_leader
-        self._database_dialect = response.database_dialect
+        # Only update if the data is specific to avoid losing specificity.
+        if response.database_dialect != DatabaseDialect.DATABASE_DIALECT_UNSPECIFIED:
```
harshachinta, line 529: Will there be such a situation where backend returns UNSPECIFIED dialect?
AFAIK, the default dialect is GOOGLE_SQL. Correct me if I am missing something

harshachinta (approved):

google/cloud/spanner_v1/database.py
larkee: This is an issue when working with the emulator. It is returning `DATABASE_DIALECT_UNSPECIFIED`  which is overriding the dialect that this object already had. This shouldn't be an issue in production.

PR #1202, opened 2024-09-13
test: fix timestamp verification in system test

larkee:
Currently, the `test_batch_insert_then_read_all_datatypes` and `test_execute_sql_w_timestamp_bindings` tests fail when run outside of the UTC timezone. This is frustrating when running tests locally for those who are based in a different timezone.

This PR moves the handling of the timezones to `_assert_timestamp`. All datetime objects created without a timezone are assumed to be in local time. By asserting that it does not have a timezone first, we can safely convert it to UTC which should match the value returned from Spanner.


chalmerlowe:

PR #1217, opened 2024-10-24
test: add mock server tests

olavloite:
Adds a mock server and tests for the Python client.


google/cloud/spanner_v1/testing/spanner_database_admin_pb2_grpc.py
```diff
@@ -0,0 +1,1267 @@
+# Generated by the gRPC Python protocol compiler plugin. DO NOT EDIT!
+
+
+# Generated with the following commands:
```
harshachinta, line 4: It seems that whenever a new admin API is introduced, this file will need to be regenerated. Alternatively, we should explore if there’s an automated way to update or refresh this file.

However, this is currently out of scope from mock server functionality and not a high priority.

harshachinta (approved):

google/cloud/spanner_v1/testing/spanner_database_admin_pb2_grpc.py
olavloite: I agree. Having this auto-generated would be nice. But we only really need to re-generate if there's a new feature that we actually also want to include in the mock server tests. Nothing should break if a new feature that we just don't try to use is introduced.

PR #1223, opened 2024-11-04
fix: pass through route-to-leader option in dbapi

olavloite:
The route-to-leader option given to the dbapi connection was not passed on to the actual Spanner client.


rahul2393 (approved):

harshachinta:
@olavloite 
There seems to be an issue with the kokoro build and it is unrelated to this PR changes.
Looks like some change in kokoro configuration which is causing the issue. Will check with Python SDK team and get back.

harshachinta:
> @olavloite There seems to be an issue with the kokoro build and it is unrelated to this PR changes. Looks like some change in kokoro configuration which is causing the issue. Will check with Python SDK team and get back.

This issue is fixed now. The builds should be green.

PR #1228, opened 2024-11-09
tracing: enable OpenTelemetry as a first class dependency

odeke-em:
This change attempts to make OpenTelemetry a first class dependency used without any guards.

### DIFF
--- a/google/cloud/spanner_v1/_opentelemetry_tracing.py
+++ b/google/cloud/spanner_v1/_opentelemetry_tracing.py
@@ -19,17 +19,12 @@ from contextlib import contextmanager
 from google.cloud.spanner_v1 import SpannerClient
 from google.cloud.spanner_v1 import gapic_version
 
-try:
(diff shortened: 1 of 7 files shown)


odeke-em:
Kindly pinging @harshachinta @sakthivelmanii @surbhigarg92 to kindly help me review and also run the bots.

odeke-em:
Kind ping @rahul2393 to help me run the bots.

odeke-em:
It turns out that we are blocked from this PR being merged because OpenTelemetry is only supported for Python3.8 per https://opentelemetry.io/docs/languages/python/ and above while this package supports Python3.7 which went into end-of-life in 2018 https://www.python.org/downloads/release/python-370/

odeke-em:
@harshachinta kindly help me place a "do not merge" label on this PR so that it'll be sealed and visible for when Python3.8 is finally the minimum version supported (which OpenTelemetry too supports as the minimum)

PR #1236, opened 2024-11-13
fix: json data type for non object values

surbhigarg92:
JSON supports various types of values. At a high level: Objects, Arrays, and various kinds of Scalars. A Spanner column of type JSON can support any of these:

spanner> create table t_json (i int64, j json) primary key (i);
spanner> insert into t_json (i,j) values (1,JSON '[1,2,3]');
spanner> insert into t_json (i,j) values (2,JSON '"foo"');
spanner> insert into t_json (i,j) values (3,JSON '{"a": "b"}');
spanner> select * from t_json;
+---+-----------+
| i | j         |
+---+-----------+
| 1 | [1,2,3]   |
| 2 | "foo"     |
| 3 | {"a":"b"} |
+---+-----------+
Spanner's Python Client assumes that all JSON values are Objects. If you read back a value that is not an Object, you will get an error that looks like:

ValueError: dictionary update sequence element #0 has length 3; 2 is required
The root cause of this issue is this line here: https://github.com/googleapis/python-spanner/blob/main/google/cloud/spanner_v1/data_types.py#L47

            super(JsonObject, self).__init__(*args, **kwargs)


google/cloud/spanner_v1/data_types.py
```diff
@@ -31,17 +31,26 @@ class JsonObject(dict):
     def __init__(self, *args, **kwargs):
         self._is_null = (args, kwargs) == ((), {}) or args == (None,)
         self._is_array = len(args) and isinstance(args[0], (list, tuple))
+        self._is_simple_value = len(args) == 1 and not isinstance(args[0], (list, dict))
```
harshachinta, line 34: nit: is_scalar_value

harshachinta (approved):

sakthivelmanii (approved):

PR #1241, opened 2024-11-18
feat(observability): more descriptive and value adding spans

odeke-em:
This change adds more descriptive and value adding spans to replace the generic CloudSpanner.ReadWriteTransaction. With this change, we add new spans:
* CloudSpanner.Database.run_in_transaction
* CloudSpanner.execute_pdml
* CloudSpanner.execute_sql
* CloudSpanner.execute_update

## Exhibit
Given this code
```python
def tx_dql_select(tx):
    try:
        tx.execute_update("DELETE FROM Singers WHERE 1=1")
    except:
        pass

    tx.execute_update(
        'INSERT INTO Singers(SingerId, FirstName, LastName) VALUES(1000, "Bryan", "Adams"), (2000, "Slash", "GunsNRoses")'
    )
    tx.update(
        "Singers",
        columns=["SingerId", "FirstName"],
        values=[["1", "Bryan"], ["2", "Slash"]],
    )
    rows = tx.execute_sql("SELECT * FROM Singers")
    for row in rows:
        print(row)
```

we now have this exhibit
### Without gRPC instrumentation
Screenshot 2024-11-18 at 2 08 46 AM

### With gRPC instrumentation
Screenshot 2024-11-18 at 2 12 26 AM

### DIFF
--- a/google/cloud/spanner_v1/_opentelemetry_tracing.py
+++ b/google/cloud/spanner_v1/_opentelemetry_tracing.py
@@ -55,15 +55,11 @@ def get_tracer(tracer_provider=None):
     return tracer_provider.get_tracer(TRACER_NAME, TRACER_VERSION)
 
 
-@contextmanager
-def trace_call(name, session=None, extra_attributes=None, observability_options=None):
-    if session:
-        session._last_use_time = datetime.now()
-
-    if not (HAS_OPENTELEMETRY_INSTALLED and name):
-        # Empty context manager. Users will have to check if the generated value is None or a span
-        yield None
-        return
+def _make_tracer_and_span_attributes(
+    session=None, extra_attributes=None, observability_options=None
+):
+    if not HAS_OPENTELEMETRY_INSTALLED:
+        return None, None
 
     tracer_provider = None
 
@@ -103,9 +99,77 @@ def trace_call(name, session=None, extra_attributes=None, observability_options=
 
     if not enable_extended_tracing:
         attributes.pop("db.statement", False)
+        attributes.pop("sql", False)
+    else:
+        # Otherwise there are places where the annotated sql was inserted
+        # directly from the arguments as "sql", and transform those into "db.statement".
+        db_statement = attributes.get("db.statement", None)
+        if not db_statement:
+            sql = attributes.get("sql", None)
+            if sql:
+                attributes = attributes.copy()
+                attributes.pop("sql", False)
+                attributes["db.statement"] = sql
+
+    return tracer, attributes
+
+
+def trace_call_end_lazily(
+    name, session=None, extra_attributes=None, observability_options=None
+):
+    """
+    trace_call_end_lazily is used in situations where you don't want a context managed
+    span in a with statement to end as soon as a block exits. This is useful for example
+    after a Database.batch or Database.snapshot but without a context manager.
+    If you need to directly invoke tracing with a context manager, please invoke
+    `trace_call` with which you can invoke
+    ￼        `with trace_call(...) as span:`
+    It is the caller's responsibility to explicitly invoke the returned ending function.
+    """
+    if not name:
+        return None
+
+    tracer, span_attributes = _make_tracer_and_span_attributes(
+        session, extra_attributes, observability_options
+    )
+    if not tracer:
+        return None
+
+    span = tracer.start_span(
+        name, kind=trace.SpanKind.CLIENT, attributes=span_attributes
+    )
+    ctx_manager = trace.use_span(span, end_on_exit=True, record_exception=True)
+    ctx_manager.__enter__()
+
+    def discard(exc_type=None, exc_value=None, exc_traceback=None):
+        if not exc_type:
+            span.set_status(Status(StatusCode.OK))
+
+        ctx_manager.__exit__(exc_type, exc_value, exc_traceback)
+
+    return discard
+
+
+@contextmanager
+def trace_call(name, session=None, extra_attributes=None, observability_options=None):
+    """
+    ￼   trace_call is used in situations where you need to end a span with a context manager
+    ￼   or after a scope is exited. If you need to keep a span alive and lazily end it, please
+    ￼   invoke `trace_call_end_lazily`.
+    """
+    if not name:
+        yield None
+        return
+
+    tracer, span_attributes = _make_tracer_and_span_attributes(
+        session, extra_attributes, observability_options
+    )
+    if not tracer:
+        yield None
+        return
 
     with tracer.start_as_current_span(
-        name, kind=trace.SpanKind.CLIENT, attributes=attributes
+        name, kind=trace.SpanKind.CLIENT, attributes=span_attributes
     ) as span:
         try:
             yield span
@@ -135,3 +199,16 @@ def get_current_span():
 def add_span_event(span, event_name, event_attributes=None):
     if span:
         span.add_event(event_name, event_attributes)
+
+
+def add_event_on_current_span(event_name, event_attributes=None, span=None):
+    if not span:
+        span = get_current_span()
+
+    add_span_event(span, event_name, event_attributes)
+
+
+def record_span_exception_and_status(span, exc):
+    if span:
+        span.set_status(Status(StatusCode.ERROR, str(exc)))
+        span.record_exception(exc)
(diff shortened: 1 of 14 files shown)


google/cloud/spanner_v1/database.py
```diff
@@ -726,6 +727,15 @@ def execute_pdml():
 
                 return result_set.stats.row_count_lower_bound
 
+            with SessionCheckout(self._pool) as session:
+                observability_options = getattr(self, "observability_options", None)
+                with trace_call(
+                    "CloudSpanner.execute_pdml",
```
harshachinta, line 733:
```suggestion
                    "CloudSpanner.execute_partitioned_dml",
```

google/cloud/spanner_v1/database.py
```diff
@@ -726,6 +727,15 @@ def execute_pdml():
 
                 return result_set.stats.row_count_lower_bound
```
harshachinta, line 728: I am not able to add comment in above line, hence adding comment here
pass `trace_name="CloudSpannerOperation.ExecuteStreamingSql"` at https://github.com/googleapis/python-spanner/blob/054a18658eedc5d4dbecb7508baa3f3d67f5b815/google/cloud/spanner_v1/database.py#L717

google/cloud/spanner_v1/database.py
```diff
@@ -888,7 +898,13 @@ def run_in_transaction(self, func, *args, **kw):
         # done, flip the sanity check bit back.
         try:
             with SessionCheckout(self._pool) as session:
-                return session.run_in_transaction(func, *args, **kw)
+                observability_options = getattr(self, "observability_options", None)
+                with trace_call(
```
harshachinta, line 902: Same as above comment.
This parent span should be above SessionCheckout to capture session related events

google/cloud/spanner_v1/_opentelemetry_tracing.py
```diff
@@ -80,7 +80,7 @@ def trace_call(name, session, extra_attributes=None, observability_options=None)
     attributes = {
         "db.type": "spanner",
         "db.url": SpannerClient.DEFAULT_ENDPOINT,
-        "db.instance": session._database.name,
+        "db.instance": "" if not session._database else session._database.name,
```
harshachinta, line 83: Please see comment below for more details.
Can we get this value from observability_options in case session is. not passed?

google/cloud/spanner_v1/snapshot.py
```diff
@@ -538,7 +538,7 @@ def _get_streamed_result_set(
         iterator = _restart_on_unavailable(
             restart,
             request,
-            "CloudSpanner.ReadWriteTransaction",
+            "CloudSpanner.execute_sql",
```
harshachinta, line 541:
```suggestion
            "CloudSpanner.execute_streaming_sql",
```

google/cloud/spanner_v1/transaction.py
```diff
@@ -423,7 +423,7 @@ def execute_update(
                 response = self._execute_request(
                     method,
                     request,
-                    "CloudSpanner.ReadWriteTransaction",
+                    "CloudSpanner.execute_update",
```
harshachinta, line 426:
```suggestion
                    "CloudSpanner.execute_sql",
```

google/cloud/spanner_v1/transaction.py
```diff
@@ -440,7 +440,7 @@ def execute_update(
             response = self._execute_request(
                 method,
                 request,
-                "CloudSpanner.ReadWriteTransaction",
+                "CloudSpanner.execute_update",
```
harshachinta, line 443:
```suggestion
                "CloudSpanner. execute_sql",
```

google/cloud/spanner_v1/session.py
```diff
@@ -421,21 +421,13 @@ def run_in_transaction(self, func, *args, **kw):
         )
         attempts = 0
 
-        while True:
```
harshachinta, line 424: I don't quite understand the need for refactoring here. Can you help me understand?
Just adding a trace_call with `CloudSpanner.run_in_transaction` before `while True` should work right?

harshachinta (changes requested):

google/cloud/spanner_v1/transaction.py
odeke-em: But the method is in transaction.execute_update, why .execute_sql?

google/cloud/spanner_v1/transaction.py
harshachinta: Don't need this change

google/cloud/spanner_v1/database.py
odeke-em: Done and thank you @harshachinta

google/cloud/spanner_v1/session.py
odeke-em: Gotcha! I really was just being very sensitive with it to ensure that if there were multiple invocations in that while loop of creating a new Transaction that we'd show them all but even better I have added span events for aborting, retrying et al

odeke-em:
Kindly help me take a look @harshachinta and help me run the bots afresh!

odeke-em:
@harshachinta kindly help me run the bots one more time, I sent up a fix!

google/cloud/spanner_v1/_opentelemetry_tracing.py
odeke-em: Yes indeed.

google/cloud/spanner_v1/database.py
```diff
@@ -726,6 +727,15 @@ def execute_pdml():
 
                 return result_set.stats.row_count_lower_bound
 
+            with SessionCheckout(self._pool) as session:
```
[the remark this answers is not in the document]
odeke-em: Roger that, I passed in the name, good ideas and thanks!

google/cloud/spanner_v1/database.py
```diff
@@ -726,6 +734,17 @@ def execute_pdml():
 
                 return result_set.stats.row_count_lower_bound
 
+            observability_options = getattr(self, "observability_options", {})
+            if isinstance(observability_options, dict):
+                observability_options["db_name"] = self.name
+            with trace_call(
+                "CloudSpanner.execute_partitioned_pdml",
```
harshachinta, line 744: The problem i see here is with every retry there will be a new span parent `CloudSpanner.execute_partitioned_pdml` which does not indicate that a retry is happening.
Instead we should capture the retries with in only one parent span `CloudSpanner.execute_partitioned_pdml`.

google/cloud/spanner_v1/database.py
```diff
@@ -887,8 +906,16 @@ def run_in_transaction(self, func, *args, **kw):
         # Check out a session and run the function in a transaction; once
         # done, flip the sanity check bit back.
         try:
-            with SessionCheckout(self._pool) as session:
-                return session.run_in_transaction(func, *args, **kw)
+            observability_options = getattr(self, "observability_options", {})
+            if isinstance(observability_options, dict):
+                observability_options["db_name"] = self.name
```
harshachinta, line 914: same here, we can add this at a single place in observability_options getter property?

google/cloud/spanner_v1/pool.py
```diff
@@ -205,15 +206,16 @@ def bind(self, database):
...
-        while not self._sessions.full():
-            resp = api.batch_create_sessions(
-                request=request,
-                metadata=metadata,
-            )
-            for session_pb in resp.session:
-                session = self._new_session()
-                session._session_id = session_pb.name.split("/")[-1]
-                self._sessions.put(session)
+        while trace_call("Cloudspanner.FixedPool.BatchCreateSessions", self):
```
harshachinta, line 209: 1. We cannot pass self to trace_call since the class here is pool not session.
2. we should also pass observability_options here to consider local tracer provider if any set by customer.

odeke-em: Done, thanks!

google/cloud/spanner_v1/database.py
odeke-em: Done, thanks!

odeke-em:
Was split up but can be revived if needed.

PR #1248, opened 2024-12-03
fix(tracing): only set span.status=OK if UNSET

### LINKED ISSUE #1246: tracing: spans are unconditionally set to a status of OK even after setting to ERROR after invoking `with trace_call` and that fails due to unconditional overriding because OpenTelemetry-Python can change the status after ERROR to OK
Noticed while writing extensive tests for modernized tracing functionality that we've got this code 

https://github.com/googleapis/python-spanner/blob/ccae6e0287ba6cf3c14f15a907b2106b11ef1fdc/google/cloud/spanner_v1/_opentelemetry_tracing.py#L98-L108

which doesn't do exactly what you expect with OpenTelemetry

Once `yield span` is invoked, if no exception was encountered, we unconditionally set the span's value to OK even if it previously were set to ERROR and here is a test

## Reproduction
```python
from tests import _helpers as ot_helpers
        
@pytest.mark.skipif(
    not ot_helpers.HAS_OPENTELEMETRY_INSTALLED,
    reason="Tracing requires OpenTelemetry",
)               
def test_trace_call_status():
    from opentelemetry.sdk.trace.export import SimpleSpanProcessor
    from opentelemetry.sdk.trace.export.in_memory_span_exporter import (
        InMemorySpanExporter,
    )
    from google.cloud.spanner_v1._opentelemetry_tracing import trace_call
    from opentelemetry.trace.status import Status, StatusCode
    from opentelemetry.sdk.trace import TracerProvider
    from opentelemetry.sdk.trace.sampling import ALWAYS_ON
        
    tracer_provider = TracerProvider(sampler=ALWAYS_ON)
    trace_exporter = InMemorySpanExporter()
    tracer_provider.add_span_processor(SimpleSpanProcessor(trace_exporter))
    observability_options = dict(tracer_provider=tracer_provider)
        
    with trace_call(
        "VerifyBehavior", observability_options=observability_options
    ) as span:
        span.set_status(Status(StatusCode.ERROR, "Our error exhibit"))
        
    span_list = trace_exporter.get_finished_spans()
    got_statuses = []
        
    for span in span_list:
        got_statuses.append(
            (span.name, span.status.status_code, span.status.description)
        )
        
    want_statuses = [
        ("VerifyBehavior", StatusCode.ERROR, "Our error exhibit"),
    ]
    assert got_statuses == want_statuses
```

which would fail with

```shell
$ nox -s system-3.8 -- -k test_trace_call_status
>       assert got_statuses == want_statuses
E       AssertionError: assert [('VerifyBeha...OK: 1>, None)] == [('VerifyBeha...ror exhibit')]
E         
E         At index 0 diff: ('VerifyBehavior', <StatusCode.OK: 1>, None) != ('VerifyBehavior', <StatusCode.ERROR: 2>, 'Our error exhibit')
E         Use -v to get more diff

tests/system/test_observability_options.py:180: AssertionError
```

## Fix
```diff
diff --git a/google/cloud/spanner_v1/_opentelemetry_tracing.py b/google/cloud/spanner_v1/_opentelemetry_tracing.py
index 9472251..65b3441 100644
--- a/google/cloud/spanner_v1/_opentelemetry_tracing.py
+++ b/google/cloud/spanner_v1/_opentelemetry_tracing.py
@@ -142,6 +142,9 @@ def trace_call_end_lazily(
     ctx_manager.__enter__()
 
     def discard(exc_type=None, exc_value=None, exc_traceback=None):
+        if not exc_type:
+            span.set_status(Status(StatusCode.OK))
+
         ctx_manager.__exit__(exc_type, exc_value, exc_traceback)
 
     return discard
@@ -175,8 +178,12 @@ def trace_call(name, session=None, extra_attributes=None, observability_options=
             span.record_exception(error)
             raise
         else:
-            span.set_status(Status(StatusCode.OK))
-
+            if span._status.status_code == StatusCode.UNSET:
+                # OpenTelemetry-Python only allows a status change
+                # if the current code is UNSET or ERROR. At the end
+                # of the generator's consumption, only set it to OK
+                # it wasn't previously set otherwise
+                span.set_status(Status(StatusCode.OK))
```

/cc @harshachinta

odeke-em:
In modernized OpenTelemetry-Python, if the SpanStatus was not already set to OK, it can be changed and
the code for trace_call was accidentally unconditionally setting the status to OK if there was no exception. This change fixes that and adds tests to lock this behavior in.

Fixes #1246


google/cloud/spanner_v1/_opentelemetry_tracing.py
```diff
@@ -105,4 +105,10 @@ def trace_call(name, session, extra_attributes=None, observability_options=None)
             span.record_exception(error)
             raise
         else:
-            span.set_status(Status(StatusCode.OK))
```
harshachinta, line 108: I am a bit confused here. Can you help me understand the scenario where this happens.
As per my understanding,
1. When there is `an exception`, the else block will never be invoked and only the error status is set.
2. When there is `no exception`, then the else block will be invoked and `OK` status is set

tests/unit/test__opentelemetry_tracing.py
```diff
@@ -158,3 +158,41 @@ def test_trace_codeless_error(self):
...
+            tracer_provider.add_span_processor(SimpleSpanProcessor(trace_exporter))
+            observability_options = dict(tracer_provider=tracer_provider)
+
+            session = _make_session()
+            with _opentelemetry_tracing.trace_call(
+                "VerifyTerminalSpanStatus",
+                session,
+                observability_options=observability_options,
+            ) as span:
+                span.set_status(Status(StatusCode.ERROR, "Our error exhibit"))
```
harshachinta, line 185: Lets assume if there is a code block here that threw an exception then the status would be recorded by our trace_call right?

google/cloud/spanner_v1/_opentelemetry_tracing.py
odeke-em: So if an exception doesn't occur the else block is executed to set to OK, but the code in the with statement can and will change the status for example to ERROR; when it comes to this block then it gets changed back to OK which is not correct. Please see the test I provided in the issue too.

tests/unit/test__opentelemetry_tracing.py
```diff
@@ -158,3 +158,41 @@ def test_trace_codeless_error(self):
...
+            from opentelemetry.sdk.trace.sampling import ALWAYS_ON
+            from opentelemetry import trace
+
+            tracer_provider = TracerProvider(sampler=ALWAYS_ON)
+            trace_exporter = InMemorySpanExporter()
+            tracer_provider.add_span_processor(SimpleSpanProcessor(trace_exporter))
+            observability_options = dict(tracer_provider=tracer_provider)
+
+            session = _make_session()
+            with _opentelemetry_tracing.trace_call(
```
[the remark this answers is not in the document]
harshachinta: Is there any scenario in our code where you had to explicitly set the span status to ERROR and not throw an exception instead?

google/cloud/spanner_v1/_opentelemetry_tracing.py
```diff
@@ -105,4 +105,10 @@ def trace_call(name, session, extra_attributes=None, observability_options=None)
...
+            if span._status.status_code == StatusCode.UNSET:
```
harshachinta, line 108: is span._status always not `None` or do we need a check for that?

harshachinta: The opentelemetry package has this check
image

google/cloud/spanner_v1/_opentelemetry_tracing.py
harshachinta: Had an offline conversation with @odeke-em .
The scenario here is `When a transaction is aborted and retried, we want to record that transaction status`

google/cloud/spanner_v1/_opentelemetry_tracing.py
odeke-em: It is always defaulted to UNSET but sure, I can add that guard since we are poking the internals. Let me send up a quick change.

tests/unit/test__opentelemetry_tracing.py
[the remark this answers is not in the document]
harshachinta: Scenario here is - When a transaction is aborted and retried, we want to record that transaction status

google/cloud/spanner_v1/_opentelemetry_tracing.py
odeke-em: I've added an extra check for `(not span._status) or `

harshachinta (approved):

PR #1252, opened 2024-12-04
perf: remove repeated GetSession calls for FixedSizePool

olavloite:
Add a _last_use_time to Session and use this to determine whether the FixedSizePool should check whether the session still exists, and whether it should be replaced. This significantly reduces the number of times that GetSession is called when using FixedSizePool.


tests/unit/test_pool.py
```diff
@@ -184,13 +185,30 @@ def test_bind(self):
...
+        for i in (3, 2, 1, 0):
+            session = pool.get()
+            self.assertIs(session, SESSIONS[i])
+            self.assertFalse(session._exists_checked)
+            self.assertFalse(pool._sessions.full())
+
+    def test_get_non_expired(self):
+        pool = self._make_one(size=4)
+        database = _Database("name")
+        last_use_time = datetime.utcnow() - timedelta(minutes=56)
```
harshachinta, line 205: The intention of setting it to 56 minutes (i.e., >55 minutes) is to ensure that the max age has been exceeded, triggering a verification of session existence through the session.exists call. Is my understanding correct?

google/cloud/spanner_v1/session.py
```diff
@@ -273,6 +286,7 @@ def read(self, table, columns, keyset, index="", limit=0, column_info=None):
         :rtype: :class:`~google.cloud.spanner_v1.streamed.StreamedResultSet`
         :returns: a result set instance which can be used to consume rows.
         """
+        self._last_use_time = datetime.now()
```
harshachinta, line 289: Instead of updating the `_last_use_time` attribute here, should we consider updating it in the final layer where the RPC call is actually made? This would likely require changes to the snapshot.py and transaction.py files.
 For reference: https://github.com/googleapis/python-spanner/blob/7df93ca9b11a5cd2fff448587d87d63e29d7a828/google/cloud/spanner_v1/snapshot.py#L180


There are two main reasons for this approach:
1. The session.snapshot method internally calls the Snapshot.read() function. So modifying snapshot file will be a single file change.
2. As far as I understand, it is uncommon for customers to directly call the read method from the session class. Instead, they typically use the snapshot class, as shown in the example below:
```
    results = snapshot.read("SELECT 1")
```

tests/unit/test_pool.py
olavloite: Yes, that is correct. Any value larger than the default max_age would trigger such a check.

google/cloud/spanner_v1/session.py
olavloite: That is good point. This implementation actually also misses a lot of other places where a session is used, for example in transaction.py and batch.py.

I actually think that we can move almost all of the updates of _last_use_time into the trace_call(..) function. That function is used for all RPCs that use a session. I'll update the PR along those lines.

google/cloud/spanner_v1/session.py
```diff
@@ -181,6 +191,7 @@ def exists(self):
         ) as span:
             try:
                 api.get_session(name=self.name, metadata=metadata)
+                self._last_use_time = datetime.now()
```
harshachinta, line 194: nit: we might not need it here as it is getting assigned in trace_call already.

harshachinta (approved):

google/cloud/spanner_v1/session.py
olavloite: Yeah, you're right. I missed the surrounding trace_call.

PR #1259, opened 2024-12-06
feat: add updated span events + trace more methods

odeke-em:
This change carves out parts of PR googleapis/python-spanner#1241 in smaller pieces to ease with smaller reviews.
This change adds more span events, updates important spans to make them more distinct like changing:

"CloudSpanner.ReadWriteTransaction" to more direct and more pointed spans like:
* CloudSpanner.Transaction.execute_streaming_sql

Also added important spans:
* CloudSpanner.Database.run_in_transaction
* CloudSpanner.Session.run_in_transaction

### DIFF
--- a/google/cloud/spanner_v1/_opentelemetry_tracing.py
+++ b/google/cloud/spanner_v1/_opentelemetry_tracing.py
@@ -103,7 +103,6 @@ def trace_call(name, session=None, extra_attributes=None, observability_options=
 
     if not enable_extended_tracing:
         attributes.pop("db.statement", False)
-        attributes.pop("sql", False)
 
     with tracer.start_as_current_span(
         name, kind=trace.SpanKind.CLIENT, attributes=attributes
@@ -136,16 +135,3 @@ def get_current_span():
 def add_span_event(span, event_name, event_attributes=None):
     if span:
         span.add_event(event_name, event_attributes)
-
-
-def add_event_on_current_span(event_name, event_attributes=None, span=None):
-    if not span:
-        span = get_current_span()
-
-    add_span_event(span, event_name, event_attributes)
-
-
-def record_span_exception_and_status(span, exc):
-    if span:
-        span.set_status(Status(StatusCode.ERROR, str(exc)))
-        span.record_exception(exc)
--- a/google/cloud/spanner_v1/_opentelemetry_tracing.py
+++ b/google/cloud/spanner_v1/_opentelemetry_tracing.py
@@ -62,9 +62,6 @@ def trace_call(name, session=None, extra_attributes=None, observability_options=
         yield None
         return
 
-    if session:
-        session._last_use_time = datetime.now()
-
     tracer_provider = None
 
     # By default enable_extended_tracing=True because in a bid to minimize
@@ -104,16 +101,6 @@ def trace_call(name, session=None, extra_attributes=None, observability_options=
     if not enable_extended_tracing:
         attributes.pop("db.statement", False)
         attributes.pop("sql", False)
-    else:
-        # Otherwise there are places where the annotated sql was inserted
-        # directly from the arguments as "sql", and transform those into "db.statement".
-        db_statement = attributes.get("db.statement", None)
-        if not db_statement:
-            sql = attributes.get("sql", None)
-            if sql:
-                attributes = attributes.copy()
-                attributes.pop("sql", False)
-                attributes["db.statement"] = sql
 
     with tracer.start_as_current_span(
         name, kind=trace.SpanKind.CLIENT, attributes=attributes
--- a/google/cloud/spanner_v1/_opentelemetry_tracing.py
+++ b/google/cloud/spanner_v1/_opentelemetry_tracing.py
@@ -56,14 +56,14 @@ def get_tracer(tracer_provider=None):
 
 
 @contextmanager
-def trace_call(name, session, extra_attributes=None, observability_options=None):
-    if session:
-        session._last_use_time = datetime.now()
-
-    if not HAS_OPENTELEMETRY_INSTALLED or not session:
+def trace_call(name, session=None, extra_attributes=None, observability_options=None):
+    if not (HAS_OPENTELEMETRY_INSTALLED and name):
         # Empty context manager. Users will have to check if the generated value is None or a span
         yield None
         return
+
+    if session:
+        session._last_use_time = datetime.now()
 
     tracer_provider = None
 
@@ -72,20 +72,24 @@ def trace_call(name, session, extra_attributes=None, observability_options=None)
     # on by default.
     enable_extended_tracing = True
 
+    db_name = ""
+    if session and getattr(session, "_database", None):
+        db_name = session._database.name
+
     if isinstance(observability_options, dict):  # Avoid false positives with mock.Mock
         tracer_provider = observability_options.get("tracer_provider", None)
         enable_extended_tracing = observability_options.get(
             "enable_extended_tracing", enable_extended_tracing
         )
+        db_name = observability_options.get("db_name", db_name)
 
     tracer = get_tracer(tracer_provider)
 
     # Set base attributes that we know for every trace created
-    db = session._database
     attributes = {
         "db.type": "spanner",
         "db.url": SpannerClient.DEFAULT_ENDPOINT,
-        "db.instance": "" if not db else db.name,
+        "db.instance": db_name,
         "net.host.name": SpannerClient.DEFAULT_ENDPOINT,
         OTEL_SCOPE_NAME: TRACER_NAME,
         OTEL_SCOPE_VERSION: TRACER_VERSION,
@@ -99,6 +103,17 @@ def trace_call(name, session, extra_attributes=None, observability_options=None)
 
     if not enable_extended_tracing:
         attributes.pop("db.statement", False)
+        attributes.pop("sql", False)
+    else:
+        # Otherwise there are places where the annotated sql was inserted
+        # directly from the arguments as "sql", and transform those into "db.statement".
+        db_statement = attributes.get("db.statement", None)
+        if not db_statement:
+            sql = attributes.get("sql", None)
+            if sql:
+                attributes = attributes.copy()
+                attributes.pop("sql", False)
+                attributes["db.statement"] = sql
 
     with tracer.start_as_current_span(
         name, kind=trace.SpanKind.CLIENT, attributes=attributes
@@ -131,3 +146,16 @@ def get_current_span():
 def add_span_event(span, event_name, event_attributes=None):
     if span:
         span.add_event(event_name, event_attributes)
+
+
+def add_event_on_current_span(event_name, event_attributes=None, span=None):
+    if not span:
+        span = get_current_span()
+
+    add_span_event(span, event_name, event_attributes)
+
+
+def record_span_exception_and_status(span, exc):
+    if span:
+        span.set_status(Status(StatusCode.ERROR, str(exc)))
+        span.record_exception(exc)
--- a/google/cloud/spanner_v1/_opentelemetry_tracing.py
+++ b/google/cloud/spanner_v1/_opentelemetry_tracing.py
@@ -57,6 +57,9 @@ def get_tracer(tracer_provider=None):
 
 @contextmanager
 def trace_call(name, session=None, extra_attributes=None, observability_options=None):
+    if session:
+        session._last_use_time = datetime.now()
+
     if not (HAS_OPENTELEMETRY_INSTALLED and name):
         # Empty context manager. Users will have to check if the generated value is None or a span
         yield None
--- a/google/cloud/spanner_v1/batch.py
+++ b/google/cloud/spanner_v1/batch.py
@@ -26,7 +26,10 @@ from google.cloud.spanner_v1._helpers import (
     _metadata_with_prefix,
     _metadata_with_leader_aware_routing,
 )
-from google.cloud.spanner_v1._opentelemetry_tracing import trace_call
+from google.cloud.spanner_v1._opentelemetry_tracing import (
+    add_event_on_current_span,
+    trace_call,
+)
 from google.cloud.spanner_v1 import RequestOptions
 from google.cloud.spanner_v1._helpers import _retry
 from google.cloud.spanner_v1._helpers import _check_rst_stream_error
@@ -70,6 +73,10 @@ class _BatchBase(_SessionWrapper):
         :param values: Values to be modified.
         """
         self._mutations.append(Mutation(insert=_make_write_pb(table, columns, values)))
+        add_event_on_current_span(
+            "insert mutations added",
+            dict(table=table, columns=columns),
+        )
 
     def update(self, table, columns, values):
         """Update one or more existing table rows.
@@ -84,6 +91,10 @@ class _BatchBase(_SessionWrapper):
         :param values: Values to be modified.
         """
         self._mutations.append(Mutation(update=_make_write_pb(table, columns, values)))
+        add_event_on_current_span(
+            "update mutations added",
+            dict(table=table, columns=columns),
+        )
 
     def insert_or_update(self, table, columns, values):
         """Insert/update one or more table rows.
@@ -100,6 +111,10 @@ class _BatchBase(_SessionWrapper):
         self._mutations.append(
             Mutation(insert_or_update=_make_write_pb(table, columns, values))
         )
+        add_event_on_current_span(
+            "insert_or_update mutations added",
+            dict(table=table, columns=columns),
+        )
 
     def replace(self, table, columns, values):
         """Replace one or more table rows.
@@ -114,6 +129,10 @@ class _BatchBase(_SessionWrapper):
         :param values: Values to be modified.
         """
         self._mutations.append(Mutation(replace=_make_write_pb(table, columns, values)))
+        add_event_on_current_span(
+            "replace mutations added",
+            dict(table=table, columns=columns),
+        )
 
     def delete(self, table, keyset):
         """Delete one or more table rows.
@@ -126,6 +145,10 @@ class _BatchBase(_SessionWrapper):
         """
         delete = Mutation.Delete(table=table, key_set=keyset._to_pb())
         self._mutations.append(Mutation(delete=delete))
+        add_event_on_current_span(
+            "delete mutations added",
+            dict(table=table),
+        )
 
 
 class Batch(_BatchBase):
@@ -207,7 +230,7 @@ class Batch(_BatchBase):
         )
         observability_options = getattr(database, "observability_options", None)
         with trace_call(
-            "CloudSpanner.Commit",
+            f"CloudSpanner.{type(self).__name__}.commit",
             self._session,
             trace_attributes,
             observability_options=observability_options,
--- a/google/cloud/spanner_v1/batch.py
+++ b/google/cloud/spanner_v1/batch.py
@@ -26,10 +26,7 @@ from google.cloud.spanner_v1._helpers import (
     _metadata_with_prefix,
     _metadata_with_leader_aware_routing,
 )
-from google.cloud.spanner_v1._opentelemetry_tracing import (
-    add_event_on_current_span,
-    trace_call,
-)
+from google.cloud.spanner_v1._opentelemetry_tracing import trace_call
 from google.cloud.spanner_v1 import RequestOptions
 from google.cloud.spanner_v1._helpers import _retry
 from google.cloud.spanner_v1._helpers import _check_rst_stream_error
@@ -73,10 +70,8 @@ class _BatchBase(_SessionWrapper):
         :param values: Values to be modified.
         """
         self._mutations.append(Mutation(insert=_make_write_pb(table, columns, values)))
-        add_event_on_current_span(
-            "insert mutations added",
-            dict(table=table, columns=columns),
-        )
+        # TODO: Decide if we should add a span event per mutation:
+        # https://github.com/googleapis/python-spanner/issues/1269
 
     def update(self, table, columns, values):
         """Update one or more existing table rows.
@@ -91,10 +86,8 @@ class _BatchBase(_SessionWrapper):
         :param values: Values to be modified.
         """
         self._mutations.append(Mutation(update=_make_write_pb(table, columns, values)))
-        add_event_on_current_span(
-            "update mutations added",
-            dict(table=table, columns=columns),
-        )
+        # TODO: Decide if we should add a span event per mutation:
+        # https://github.com/googleapis/python-spanner/issues/1269
 
     def insert_or_update(self, table, columns, values):
         """Insert/update one or more table rows.
@@ -111,10 +104,8 @@ class _BatchBase(_SessionWrapper):
         self._mutations.append(
             Mutation(insert_or_update=_make_write_pb(table, columns, values))
         )
-        add_event_on_current_span(
-            "insert_or_update mutations added",
-            dict(table=table, columns=columns),
-        )
+        # TODO: Decide if we should add a span event per mutation:
+        # https://github.com/googleapis/python-spanner/issues/1269
 
     def replace(self, table, columns, values):
         """Replace one or more table rows.
@@ -129,10 +120,8 @@ class _BatchBase(_SessionWrapper):
         :param values: Values to be modified.
         """
         self._mutations.append(Mutation(replace=_make_write_pb(table, columns, values)))
-        add_event_on_current_span(
-            "replace mutations added",
-            dict(table=table, columns=columns),
-        )
+        # TODO: Decide if we should add a span event per mutation:
+        # https://github.com/googleapis/python-spanner/issues/1269
 
     def delete(self, table, keyset):
         """Delete one or more table rows.
@@ -145,10 +134,8 @@ class _BatchBase(_SessionWrapper):
         """
         delete = Mutation.Delete(table=table, key_set=keyset._to_pb())
         self._mutations.append(Mutation(delete=delete))
-        add_event_on_current_span(
-            "delete mutations added",
-            dict(table=table),
-        )
+        # TODO: Decide if we should add a span event per mutation:
+        # https://github.com/googleapis/python-spanner/issues/1269
 
 
 class Batch(_BatchBase):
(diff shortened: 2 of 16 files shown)


odeke-em:
@sakthivelmanii @harshachinta kindly help me run the bots. Thank you

google/cloud/spanner_v1/session.py
```diff
@@ -458,47 +463,100 @@ def run_in_transaction(self, func, *args, **kw):
...
+                            attributes,
+                        )
+
+                    _delay_until_retry(exc, deadline, attempts)
+                    continue
+                except GoogleAPICallError:
+                    del self._transaction
+                    add_span_event(
+                        span,
+                        "Transaction.commit failed due to GoogleAPICallError, not retrying",
```
harshachinta, line 512:
```suggestion
                        "User operation failed due to GoogleAPICallError, not retrying",
```

google/cloud/spanner_v1/session.py
```diff
@@ -458,47 +463,100 @@ def run_in_transaction(self, func, *args, **kw):
...
+                        span_attributes,
+                    )
+                    raise
+                except Exception:
+                    add_span_event(
+                        span,
+                        "Invoking Transaction.rollback(), not retrying",
```
harshachinta, line 519:
```suggestion
                        "User operation failed. Invoking Transaction.rollback(), not retrying",
```

google/cloud/spanner_v1/session.py
```diff
@@ -458,47 +463,100 @@ def run_in_transaction(self, func, *args, **kw):
...
-                txn = self._transaction
+        observability_options = getattr(self._database, "observability_options", None)
+        with trace_call(
+            "CloudSpanner.Session.run_in_transaction",
+            self,
+            observability_options=observability_options,
+        ) as span:
+            while True:
+                if self._transaction is None:
+                    add_span_event(span, "Creating Transaction")
```
harshachinta, line 474: This is an inmemory transaction and not actual transaction call to backend. This might create confusion.
Lets remove this event.

google/cloud/spanner_v1/session.py
```diff
@@ -458,47 +463,100 @@ def run_in_transaction(self, func, *args, **kw):
...
-                _delay_until_retry(exc, deadline, attempts)
-            except GoogleAPICallError:
-                del self._transaction
-                raise
-            else:
-                if self._database.log_commit_stats and txn.commit_stats:
-                    self._database.logger.info(
-                        "CommitStats: {}".format(txn.commit_stats),
-                        extra={"commit_stats": txn.commit_stats},
+                span_attributes = {"attempt": attempts}
```
harshachinta, line 484: I wonder that this span attribute keeps getting overriden during retries?
Instead should we add an event saying `Starting attempt0`, `Starting attempt1` etc?

google/cloud/spanner_v1/session.py
```diff
@@ -458,47 +463,100 @@ def run_in_transaction(self, func, *args, **kw):
...
+                txn_id = getattr(txn, "_transaction_id", "") or ""
+                if txn_id:
+                    span_attributes["transaction.id"] = txn_id
+
+                add_span_event(span, "Using Transaction", span_attributes)
+
+                try:
+                    return_value = func(txn, *args, **kw)
```
harshachinta, line 492: Lets move `attempts += 1` inside this try block. I understand it has not effect but this is to ensure we are not modifying any code that is not linked to observability

google/cloud/spanner_v1/session.py
```diff
@@ -458,47 +463,100 @@ def run_in_transaction(self, func, *args, **kw):
...
-                if self._database.log_commit_stats and txn.commit_stats:
-                    self._database.logger.info(
-                        "CommitStats: {}".format(txn.commit_stats),
-                        extra={"commit_stats": txn.commit_stats},
+                span_attributes = {"attempt": attempts}
+                txn_id = getattr(txn, "_transaction_id", "") or ""
+                if txn_id:
+                    span_attributes["transaction.id"] = txn_id
+
+                add_span_event(span, "Using Transaction", span_attributes)
```
harshachinta, line 489: In this part of code there would never be a transaction id available because of inline begin that happens with statement in user operation or with commit.
So there is not point of tracking this here. Lets remove this entire block of code.

google/cloud/spanner_v1/session.py
```diff
@@ -458,47 +463,100 @@ def run_in_transaction(self, func, *args, **kw):
...
+                except Aborted as exc:
+                    del self._transaction
+                    if span:
+                        delay_seconds = _get_retry_delay(exc.errors[0], attempts)
+                        attributes = dict(delay_seconds=delay_seconds)
+                        attributes.update(span_attributes)
+                        record_span_exception_and_status(span, exc)
+                        add_span_event(
+                            span,
+                            "Transaction was aborted, retrying afresh",
```
harshachinta, line 502:
```suggestion
                            "Transaction was aborted in user operation, retrying afresh",
```

google/cloud/spanner_v1/session.py
```diff
@@ -458,47 +463,100 @@ def run_in_transaction(self, func, *args, **kw):
...
```
harshachinta, line 540:
```suggestion
                            "Transaction got aborted during commit, retrying afresh",
```

google/cloud/spanner_v1/session.py
```diff
@@ -458,47 +463,100 @@ def run_in_transaction(self, func, *args, **kw):
...
+
+                try:
+                    return_value = func(txn, *args, **kw)
+                except Aborted as exc:
+                    del self._transaction
+                    if span:
+                        delay_seconds = _get_retry_delay(exc.errors[0], attempts)
+                        attributes = dict(delay_seconds=delay_seconds)
+                        attributes.update(span_attributes)
+                        record_span_exception_and_status(span, exc)
```
harshachinta, line 499: Thinking out loud.
When there is an aborted exception, then the current span is marked as error wkith exception,
But lets say the next retry attempt succeeds, then the span will still be annotated with this exception right?
Am i missing something?

harshachinta, line 499: I think we don't need these at all?
Here lets say there are x retries then the retry_delay getting captured will be overwritten by the xth retry value and the retry delayas of 1 to x-1 retries will not be captued. I don't see any benefit of this.

tests/unit/test_pool.py
```diff
@@ -241,25 +241,36 @@ def test_spans_bind_get(self):
...
 
-        # Check for the overall spans too.
+        span_list = self.get_finished_spans()
+        got_span_names = [span.name for span in span_list]
+        want_span_names = ["CloudSpanner.FixedPool.BatchCreateSessions", "pool.Get"]
+        print("got", got_span_names)
+        assert got_span_names == want_span_names
+
+        attrs = TestFixedSizePool.BASE_ATTRIBUTES.copy()
+        attrs["db.instance"] = ""
```
harshachinta, line 252: Why is this "" here, though we are passing "name" as the database name above?
Looks like this is because of mock class?

tests/unit/test_pool.py
```diff
@@ -241,25 +241,36 @@ def test_spans_bind_get(self):
...
+
+        # Check for the overall spans.
+        self.assertSpanAttributes(
+            "CloudSpanner.FixedPool.BatchCreateSessions",
+            status=StatusCode.OK,
+            attributes=attrs,
+            span=span_list[0],
```
harshachinta, line 259: since we are getting values in an index from array, this could lead to null pointer exception.
Can you add a assertion befrore this to check whether the span length is what we are looking for?

tests/unit/test_pool.py
```diff
@@ -241,25 +241,36 @@ def test_spans_bind_get(self):
...
+            span=span_list[-1],
         )
-
         wantEventNames = [
             "Acquiring session",
             "Waiting for a session to become available",
             "Acquired session",
         ]
-        self.assertSpanEvents("pool.Get", wantEventNames)
+        self.assertSpanEvents("pool.Get", wantEventNames, span_list[-1])
```
harshachinta, line 273: any reason for doing span_list[-1] instead of span_list[1]?

tests/unit/test_pool.py
```diff
@@ -304,18 +315,22 @@ def test_spans_pool_bind(self):
...
             pass
 
+        span_list = self.get_finished_spans()
+        got_span_names = [span.name for span in span_list]
+        want_span_names = ["testBind", "CloudSpanner.FixedPool.BatchCreateSessions"]
+        assert got_span_names == want_span_names
+
         wantEventNames = [
             "Requesting 1 sessions",
-            "Creating 1 sessions",
```
harshachinta, line 309: Why is this event removed? As per code this event is still there.
Can you check if it is not getting added?

tests/unit/test_pool.py
```diff
@@ -474,16 +490,26 @@ def test_spans_get_empty_pool(self):
             session.create.assert_called()
             self.assertTrue(pool._sessions.empty())
 
+        span_list = self.get_finished_spans()
+        got_span_names = [span.name for span in span_list]
+        want_span_names = [
+            "pool.Get",
+        ]
+        print("got_spans", got_span_names)
```
harshachinta, line 498: remove print

tests/system/test_session_api.py
```diff
@@ -699,6 +711,162 @@ def _transaction_read_then_raise(transaction):
...
+
+    span_list = trace_exporter.get_finished_spans()
+    got_span_names = [span.name for span in span_list]
+    want_span_names = [
+        "CloudSpanner.CreateSession",
+        "CloudSpanner.Transaction.execute_streaming_sql",
+        "CloudSpanner.Transaction.execute_streaming_sql",
+        "CloudSpanner.Transaction.commit",
+        "CloudSpanner.Session.run_in_transaction",
+        "CloudSpanner.Database.run_in_transaction",
```
harshachinta, line 802: There is some mistake in ordering here. Can you check?
Ideally the order should be,
```
        "CloudSpanner.Database.run_in_transaction",
        "CloudSpanner.CreateSession",
        "CloudSpanner.Session.run_in_transaction",
        "CloudSpanner.Transaction.execute_streaming_sql",
        "CloudSpanner.Transaction.execute_streaming_sql",
        "CloudSpanner.Transaction.commit",
        ```

File: tests/system/test_session_api.py [hunk 16]
```diff
@@ -699,6 +711,162 @@ def _transaction_read_then_raise(transaction):
...
+    codes = StatusCode
+    want_statuses = [
+        ("CloudSpanner.CreateSession", codes.OK, None),
+        ("CloudSpanner.Transaction.execute_streaming_sql", codes.OK, None),
+        ("CloudSpanner.Transaction.execute_streaming_sql", codes.OK, None),
+        ("CloudSpanner.Transaction.commit", codes.OK, None),
+        (
+            "CloudSpanner.Session.run_in_transaction",
+            codes.ERROR,
+            "409 Thrown from ClientInterceptor for testing",
```
harshachinta, line 863: Ideally, this span though failed in beginning, after retry got successful. But with the above thing it is likely to confuse customers.
I think this is coming because of this https://github.com/googleapis/python-spanner/pull/1259/files#r1875494695

Can we add this exception as an event in the span instead of recording? This will ensure that customer can see an event that there was failure, then a retry and in case the exception is non-retryable then anyhow it is recorded in opentlemetry during __exit__?

https://github.com/googleapis/python-spanner/blob/a6811afefa6739caa20203048635d94f9b85c4c8/google/cloud/spanner_v1/_opentelemetry_tracing.py#L114

tests/system/test_session_api.py
```diff
@@ -699,6 +711,162 @@ def _transaction_read_then_raise(transaction):
...
 
+@pytest.mark.skipif(
+    not _helpers.USE_EMULATOR,
+    reason="Emulator needed to run this tests",
+)
+@pytest.mark.skipif(
+    not ot_helpers.HAS_OPENTELEMETRY_INSTALLED,
+    reason="Tracing requires OpenTelemetry",
+)
+def test_transaction_abort_then_retry_spans(sessions_database, ot_exporter):
```
harshachinta, line 722: Looks like there is lot of redundant code here that is there in test_observability_options.py.

Can we move all these new tests to test_observability_options.py? It will help to have one single place for new observability tests?

google/cloud/spanner_v1/_opentelemetry_tracing.py
```diff
@@ -131,3 +146,16 @@ def get_current_span():
...
+    if not span:
+        span = get_current_span()
+
+    add_span_event(span, event_name, event_attributes)
+
+
+def record_span_exception_and_status(span, exc):
+    if span:
+        span.set_status(Status(StatusCode.ERROR, str(exc)))
+        span.record_exception(exc)
```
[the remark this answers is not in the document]
odeke-em: Not really, because wherever we invoke this code that's when an exception was created. It would an unnecessary pedantic step, but I can add it if you feel strongly about it.

google/cloud/spanner_v1/batch.py
```diff
@@ -70,6 +73,10 @@ def insert(self, table, columns, values):
         :param values: Values to be modified.
         """
         self._mutations.append(Mutation(insert=_make_write_pb(table, columns, values)))
+        add_event_on_current_span(
```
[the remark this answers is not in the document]
odeke-em: That's a great question but it won't be a problem because:
* spans have  a default event limit of 128 https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#span-limits
* storage costs not really as pricing on telemetry data is by span count rather than span data count unless perhps they have their own storage system but that's unlikely for those who'd be storing tens of thousands of span events

What we can do is perhaps document how to change span event limits and also document that we add events for span mutations.

google/cloud/spanner_v1/snapshot.py
```diff
@@ -335,7 +335,7 @@ def read(
                 iterator = _restart_on_unavailable(
                     restart,
                     request,
-                    "CloudSpanner.ReadOnlyTransaction",
+                    f"CloudSpanner.{type(self).__name__}.read",
```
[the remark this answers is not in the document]
odeke-em: Nope, it would be `CloudSpanner.Snapshot.read`; the final class invoking that method is what `type(self)` returns and we have tests as well for this spans :-)

google/cloud/spanner_v1/pool.py
```diff
@@ -237,29 +238,41 @@ def bind(self, database):
...
+                    dict(count=len(resp.session)),
+                )
+
+                for session_pb in resp.session:
+                    session = self._new_session()
+                    session._session_id = session_pb.name.split("/")[-1]
+                    self._sessions.put(session)
+                    returned_session_count += 1
+
             add_span_event(
```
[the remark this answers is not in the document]
odeke-em: @harshachinta so you mean the events in which invoke "Creating ...sessions" and "Created sessions"? So what happens is that as you know in this loop we create batches of spans; these events are what actually helped me uncover the fact that in our loop we weren't correctly deducting and reducing the number of spans per https://github.com/googleapis/python-spanner/issues/1226.

Did I understand your concern correctly and does my counter make sense to you?

google/cloud/spanner_v1/pool.py
```diff
@@ -550,25 +563,30 @@ def bind(self, database):
...
 
-            created_session_count += len(resp.session)
+                created_session_count += len(resp.session)
 
-        add_span_event(
-            current_span,
-            f"Requested for {requested_session_count} sessions, return {returned_session_count}",
-            span_event_attributes,
-        )
+            add_span_event(
```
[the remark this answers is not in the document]
odeke-em: @harshachinta please see line 560, I do create an event before invoking PingingPool.BatchCreateSessions
Screenshot 2024-12-09 at 10 21 27 PM

google/cloud/spanner_v1/session.py
odeke-em: Gotcha and thanks!

google/cloud/spanner_v1/session.py
odeke-em: I won't get overwritten at all: every time that this loop runs, attempt is incremented and when that you invoke span.add_event that's a new event created it isn't like replacing a value in a map.

google/cloud/spanner_v1/session.py
odeke-em: It won't be overwritten, adding a span event creates a fresh one even if with the exact same event_name and attributes because events are recorded in time and not as a lookup; it is useful to capture the delays and retries and record the events because then customers do not have to guess about what happened where, they'll clearly see the report of a retry attempt and with how much delay.

tests/system/test_session_api.py
odeke-em: It isn't a mistake though but rather the order in which spans are ended @harshachinta.

tests/unit/test_pool.py
odeke-em: Well we are using the mock _Database for which we have to then add an observability_options method which is fallacious because that doesn't even actually test our function, but let me add it.

google/cloud/spanner_v1/session.py
odeke-em: Alright, thank you @harshachinta! I have removed it.

google/cloud/spanner_v1/session.py
odeke-em: Gotcha, done and thank you!

google/cloud/spanner_v1/session.py
odeke-em: Yes it'll be exactly that, so you had a great prior suggestion which would be to create instead "exception" events and then only allow the final raised exception to be added.

google/cloud/spanner_v1/pool.py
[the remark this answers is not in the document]
odeke-em: I think we need to keep them inside BatchCreateSessions because they really are created in chunks.

tests/unit/test_pool.py
odeke-em: It won't as above we already compared against expectations for the span names.

tests/unit/test_pool.py
odeke-em: Yes:
a) pool.bind(database) came first before pool.get() hence produced the span "CloudSpanner.FixedPool.BatchCreateSessions"
b) "pool.Get" was the last span hence the -1 index

tests/unit/test_pool.py
odeke-em: Taht event is no longer on testBind, it'll be on CloudSpanner.FixedPool.BatchCreateSessions @harshachinta.

odeke-em: I added tests to check for all the events so that we can see them.

google/cloud/spanner_v1/_opentelemetry_tracing.py
```diff
@@ -56,36 +56,40 @@ def get_tracer(tracer_provider=None):
...
-        session._last_use_time = datetime.now()
-
-    if not HAS_OPENTELEMETRY_INSTALLED or not session:
+def trace_call(name, session=None, extra_attributes=None, observability_options=None):
+    if not (HAS_OPENTELEMETRY_INSTALLED and name):
         # Empty context manager. Users will have to check if the generated value is None or a span
         yield None
         return
 
+    if session:
```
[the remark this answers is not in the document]
harshachinta: @odeke-em 
this is not yet addresses.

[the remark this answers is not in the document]
odeke-em: I removed that @harshachinta as it is from the bigger PR and I don't think we shall need it anymore once some tests are sent in the other PR. I deleted it.

google/cloud/spanner_v1/_opentelemetry_tracing.py
```diff
@@ -99,6 +103,17 @@ def trace_call(name, session, extra_attributes=None, observability_options=None)
...
+        attributes.pop("sql", False)
+    else:
+        # Otherwise there are places where the annotated sql was inserted
+        # directly from the arguments as "sql", and transform those into "db.statement".
+        db_statement = attributes.get("db.statement", None)
+        if not db_statement:
+            sql = attributes.get("sql", None)
+            if sql:
+                attributes = attributes.copy()
+                attributes.pop("sql", False)
```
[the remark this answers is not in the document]
odeke-em: No longer, it was from the much bigger PR, so this is irrelevant.

google/cloud/spanner_v1/session.py
odeke-em: Done, thanks!

google/cloud/spanner_v1/session.py
odeke-em: Done.

google/cloud/spanner_v1/_opentelemetry_tracing.py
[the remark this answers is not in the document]
harshachinta: It is not from your PR, this was added by us to handle one case. Don't remove that
I am talking about this code. We need this to know when was a session last used to avoid making a backend call to verify session existence.
```
    if session:
        session._last_use_time = datetime.now()
```

google/cloud/spanner_v1/session.py
odeke-em: Done, thank you!

google/cloud/spanner_v1/session.py
odeke-em: Done, thanks!

google/cloud/spanner_v1/session.py
odeke-em: I turned these into events instead of the exceptions.

tests/_helpers.py
```diff
@@ -78,7 +78,8 @@ def tearDown(self):
 
     def assertNoSpans(self):
         if HAS_OPENTELEMETRY_INSTALLED:
-            span_list = self.ot_exporter.get_finished_spans()
+            span_list = self.get_finished_spans()
+            print("got_spans", [span.name for span in span_list])
```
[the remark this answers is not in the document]
odeke-em: Done, thanks!

tests/unit/test_pool.py
```diff
@@ -241,25 +241,36 @@ def test_spans_bind_get(self):
...
-                "Waiting for a session to become available",
-                "Acquired session",
-            ]
-            self.assertSpanEvents("pool.Get", wantEventNames, span)
 
-        # Check for the overall spans too.
+        span_list = self.get_finished_spans()
+        got_span_names = [span.name for span in span_list]
+        want_span_names = ["CloudSpanner.FixedPool.BatchCreateSessions", "pool.Get"]
+        print("got", got_span_names)
```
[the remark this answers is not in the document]
odeke-em: Done, thanks!

tests/unit/test_pool.py
odeke-em: Done, thanks!

tests/unit/test_pool.py
odeke-em: Done, thanks!

tests/system/test_session_api.py
odeke-em: I did make that update, thanks!

google/cloud/spanner_v1/_opentelemetry_tracing.py
```diff
@@ -56,11 +56,8 @@ def get_tracer(tracer_provider=None):
 
 
 @contextmanager
-def trace_call(name, session, extra_attributes=None, observability_options=None):
-    if session:
-        session._last_use_time = datetime.now()
-
-    if not HAS_OPENTELEMETRY_INSTALLED or not session:
+def trace_call(name, session=None, extra_attributes=None, observability_options=None):
```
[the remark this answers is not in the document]
harshachinta: @odeke-em 
This is a followup to this comment
https://github.com/googleapis/python-spanner/pull/1259#discussion_r1875341900

tests/unit/test_pool.py
```diff
@@ -304,20 +326,63 @@ def test_spans_pool_bind(self):
...
+        want_all_events = [
+            ("Requesting 1 sessions", {"kind": "FixedSizePool"}),
+            (
+                "exception",
+                {
+                    "exception.type": "IndexError",
+                    "exception.message": "pop from empty list",
+                    "exception.stacktrace": "EPHEMERAL",
+                    "exception.escaped": "False",
+                },
```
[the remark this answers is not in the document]
odeke-em: So given that we have a direct invocation of `with trace_call("testBind")` per
```go
        try: 
            with trace_call("testBind", fauxSession):
                pool.bind(database)
        except Exception:
            pass 
```

that then invokes code internally where an exception happens. With the exception re-raised and unhandled until the highest parent, OpenTelemetry records the exception even on the parent. If we swallowed the exception internally we wouldn't

google/cloud/spanner_v1/_opentelemetry_tracing.py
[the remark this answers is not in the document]
odeke-em: Oh I see that's an addition that got added with the session usage optimization updates from the prior week, let me revert it.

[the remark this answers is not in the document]
odeke-em: I've addressed it @harshachinta

tests/system/test_observability_options.py
```diff
@@ -128,6 +131,146 @@ def test_propagation(enable_extended_tracing):
...
+
+    span_list = trace_exporter.get_finished_spans()
+    got_span_names = [span.name for span in span_list]
+    want_span_names = [
+        "CloudSpanner.CreateSession",
+        "CloudSpanner.Transaction.execute_streaming_sql",
+        "CloudSpanner.Transaction.execute_streaming_sql",
+        "CloudSpanner.Transaction.commit",
+        "CloudSpanner.Session.run_in_transaction",
+        "CloudSpanner.Database.run_in_transaction",
```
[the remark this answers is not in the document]
odeke-em: So in this code I was pulling spans directly from the exporter as the spans arrived which is based on .end_time but sure I can also sort them by start_time for this one too.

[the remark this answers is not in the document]
odeke-em: Alright, done @harshachinta!

harshachinta (approved):

PR #1262, opened 2024-12-09
feat: support transaction and request tags in dbapi

olavloite:
Adds support for setting transaction tags and request tags in dbapi. This makes these options available to frameworks that depend on dbapi, like SQLAlchemy and Django.

Towards https://github.com/googleapis/python-spanner-sqlalchemy/issues/525


google/cloud/spanner_dbapi/cursor.py
```diff
@@ -97,6 +98,39 @@ def __init__(self, connection):
...
+        executed.
+
+        Returns:
+            str: The request tag that will be applied to the next statement on
+                 this cursor.
+        """
+        return self._request_tag
+
+    @request_tag.setter
+    def request_tag(self, value):
```
aakashanandg, line 116: Should we include some examples for end users on how to use these request tags?

aakashanandg (approved):

tests/mockserver_tests/test_tags.py
```diff
@@ -0,0 +1,180 @@
...
+    def test_select_read_write_transaction_with_request_tag(self):
+        connection = Connection(self.instance, self.database)
+        connection.autocommit = False
+        request = self._execute_and_verify_select_singers(
+            connection, request_tag="my_tag"
+        )
+        self.assertEqual("my_tag", request.request_options.request_tag)
+        self.assertEqual("", request.request_options.transaction_tag)
+
+    def test_select_read_write_transaction_with_transaction_tag(self):
```
aakashanandg, line 87: Suggestion: Could we include some negative test cases where a transaction tag is passed for a read-only transaction and check if the request options correctly reflect that flag?

olavloite: Good point, added to the mock server tests.

aakashanandg (approved):

PR #1264, opened 2024-12-11
feat(x-goog-spanner-request-id): implement Request-ID

### LINKED ISSUE #15905: Implement propagation of x-goog-spanner-request-id gRPC header on every call
This is a feature request to implement propagation of x-goog-spanner-request-id gRPC header on every call.

## Reasoning
x-goog-spanner-request-id allows propagation of an ID that'll be correlated between client and server calls. We can't rely on trace headers because they are sampled, convoluted through a bunch of infrastructure so Google Engineering asked that we craft up a better mechanism. The value will allow tracking the operations of an RPC even through retries

## Requirements
Each x-goog-spanner-request-id requires the following fields
- [x] Version of the specification being implemented: ordinal number: 1 for version 1
- [x] random Uint64 for the process' lifetime
- [x] ID of the SpannerClient being used
- [x] channelID for the gRPC channel being used
- [x] Nth Request Count: nthReq
- [x] RPC attempt number for nthReq

## Specification
Please reference https://orijtech.notion.site/x-goog-spanner-request-id-always-on-gRPC-header-to-aid-in-quick-debugging-of-errors-14aba6bc91348091a58fca7a505c9827?pvs=4

## Reference implementation
I built a Go reference implementation at https://github.com/googleapis/google-cloud-go/pull/11048

/cc @tharoldD

odeke-em:
Implements x-goog-spanner-request-id propagation for end-to-end debugging.

Fixes googleapis/google-cloud-python#15905


google/cloud/spanner_v1/database.py
```diff
@@ -714,17 +775,24 @@ def execute_pdml():
...
+                    return method(*args, **kwargs)
 
                 iterator = _restart_on_unavailable(
-                    method=method,
+                    method=wrapped_method,
                     trace_name="CloudSpanner.ExecuteStreamingSql",
                     request=request,
                     transaction_selector=txn_selector,
                     observability_options=self.observability_options,
+                    attempt=begin_txn_attempt,
```
olavloite, line 795: I think that this should be `partial_attempt`. Retries of this RPC will not retry the `BeginTransaction` RPC, it will just continue to use the same transaction ID as the one that was returned by the original `BeginTransaction` RPC.

google/cloud/spanner_v1/instance.py
```diff
@@ -501,6 +501,7 @@ def database(
                 proto_descriptors=proto_descriptors,
             )
         else:
+            print("enabled interceptors")
```
olavloite, line 504: nit: remove?

google/cloud/spanner_v1/pool.py
```diff
@@ -243,6 +243,7 @@ def bind(self, database):
             "CloudSpanner.FixedPool.BatchCreateSessions",
             observability_options=observability_options,
         ) as span:
+            attempt = 1
```
olavloite, line 246: nit: I don't think we need this. We can just pass in 1 as a constant in the function below just as for the other functions.

google/cloud/spanner_v1/snapshot.py
```diff
@@ -92,6 +95,7 @@ def _restart_on_unavailable(
     ):
         iterator = method(request=request)
     while True:
+        attempt += 1
```
olavloite, line 98: I think that this should be moved into the `except ServiceUnavailable` block. The logic in this function is:
1. Iterate until you encounter a `resume_token`. When we see that, then yield what is in the buffer, as we know that it will be safe to retry.
2. If a `ServiceUnavailable` error is encountered, then do another call (a retry) and replace the underlying iterator with the one that is returned by that new call.

olavloite: It should also be incremented in case of an `InternalServerError` that is retryable.

olavloite: Or can it be removed altogether? It does not seem to be used?

google/cloud/spanner_v1/testing/interceptors.py
```diff
@@ -63,3 +65,75 @@ def reset(self):
...
+                break
+
+        if not x_goog_request_id:
+            raise Exception(
+                f"Missing {X_GOOG_REQUEST_ID} header in {call_details.method}"
+            )
+
+        response_or_iterator = method(request_or_iterator, call_details)
+        streaming = getattr(response_or_iterator, "__iter__", None) is not None
+        print(
```
olavloite, line 94: nit: remove?

tests/unit/test_request_id_header.py
```diff
@@ -0,0 +1,387 @@
...
+            for thread in threads:
+                if thread.is_alive():
+                    thread.join()
+                else:
+                    n_finished += 1
+
+            if n_finished == len(threads):
+                break
+
+            time.sleep(1)
```
olavloite, line 102: nit: can we remove this? The call to `thread.join()` should ensure that the threads have finished.

tests/unit/test_request_id_header.py
```diff
@@ -0,0 +1,387 @@
...
+        channel_id = self.database._channel_id
+        got_stream_segments, got_unary_segments = self.canonicalize_request_id_headers()
+
+        want_unary_segments = [
+            (
+                "/google.spanner.v1.Spanner/BatchCreateSessions",
+                (1, REQ_RAND_PROCESS_ID, client_id, channel_id, 1, 1),
+            ),
+            (
+                "/google.spanner.v1.Spanner/GetSession",
```
olavloite, line 117: Note: [This PR](https://github.com/googleapis/python-spanner/pull/1252) should get rid of all of these `GetSession` calls, so rebasing on main my cause this test to fail.

tests/unit/test_request_id_header.py
```diff
@@ -0,0 +1,387 @@
...
+                (1, REQ_RAND_PROCESS_ID, client_id, channel_id, 18, 1),
+            ),
+            (
+                "/google.spanner.v1.Spanner/ExecuteStreamingSql",
+                (1, REQ_RAND_PROCESS_ID, client_id, channel_id, 20, 1),
+            ),
+        ]
+        assert got_stream_segments == want_stream_segments
+
+    def test_database_run_in_transaction_retries_on_abort(self):
```
olavloite, line 199: This test does not appear to actually verify anything (other than that it does not cause any errors). Can we either remove it, or add verifications for the scenario it should cover?

tests/unit/test_session.py
```diff
@@ -1113,10 +1174,25 @@ def unit_of_work(txn, *args, **kw):
...
+                ),
+                mock.call(
+                    session=self.SESSION_NAME,
+                    options=expected_options,
+                    metadata=[
+                        ("google-cloud-resource-prefix", database.name),
+                        ("x-goog-spanner-route-to-leader", "true"),
+                        (
+                            "x-goog-spanner-request-id",
+                            f"1.{REQ_RAND_PROCESS_ID}.{database.NTH_CLIENT.value}.1.2.2",
```
olavloite, line 1191: Is `attempt` set to 2 here because the transaction is retried? If so, then that is not correct. The `attempt` number should only be increased if the RPC itself is being retried due to a retryable error for that (single) RPC. E.g. if a `ServiceUnavailable` error is returned.

tests/unit/test_session.py
```diff
@@ -1132,10 +1208,24 @@ def unit_of_work(txn, *args, **kw):
...
-            * 2,
+                ),
+                mock.call(
+                    request=request,
+                    metadata=[
+                        ("google-cloud-resource-prefix", database.name),
+                        ("x-goog-spanner-route-to-leader", "true"),
+                        (
+                            "x-goog-spanner-request-id",
+                            f"1.{REQ_RAND_PROCESS_ID}.{database.NTH_CLIENT.value}.1.3.2",
```
olavloite, line 1224: Same here as for the `BeginTransaction` RPC: Transaction retries should not increase the attempt count. Only retries of a single RPC due to `UNAVAILABLE` errors (and other retryable error codes).

google/cloud/spanner_v1/database.py
```diff
@@ -443,6 +448,31 @@ def spanner_api(self):
...
+        This property is paramount to x-goog-spanner-request-id.
+        """
+        with self.__transport_lock:
+            api = self.spanner_api
+            channel_id = self.__transports_to_channel_id.get(api._transport, None)
+            if channel_id is None:
+                channel_id = len(self.__transports_to_channel_id) + 1
+                self.__transports_to_channel_id[api._transport] = channel_id
+
+            return channel_id
```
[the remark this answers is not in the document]
odeke-em: It is indeed cached as it is saved in the Database static __transports_to_channel_id map or would you prefer I name it with capital letters so that it is much easier to spot as __TRANSPORTS_TO_CHANNEL_ID?

google/cloud/spanner_v1/database.py
```diff
@@ -698,10 +728,20 @@ def execute_partitioned_dml(
                 _metadata_with_leader_aware_routing(self._route_to_leader_enabled)
             )
 
+        # Attempt will be incremented inside _restart_on_unavailable.
```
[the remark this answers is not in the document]
odeke-em: It applies to `begin_txn_attempt` and I've updated the comment and moving the comment before the variable itself. Thank you.

google/cloud/spanner_v1/database.py
odeke-em: Oooh, okay, that's very interesting and thanks for letting me know, let me make that update.

odeke-em: I did however already pass in and increment partial_attempt inside wrapped_method. I'll instead remove that and increment them in the appropriate sites.

google/cloud/spanner_v1/snapshot.py
odeke-em: It was being used as a side effect but sure I've removed it and performing the increment appropriately where directly used.

google/cloud/spanner_v1/database.py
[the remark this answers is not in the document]
olavloite: Yeah, what I meant is that I think we can just add an instance field `self.__channel_id` or something like that and store it there. Now it is indeed cached, but the cache look up includes taking a lock and then looking it up in a map. This is a function that is called for __every__ RPC, so keeping it as quick as possible is desirable, and I expect the extra memory usage to be negligible, as I don't think any application is going to create a very large number of database objects.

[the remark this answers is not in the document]
odeke-em: Gotcha and thank you @olavloite, done!

tests/unit/test_request_id_header.py
odeke-em: Roger that and thanks!

odeke-em:
Kindly help me take another pass @olavloite. Thank you and happy New Year to you!

tests/mockserver_tests/test_request_id_header.py
```diff
@@ -0,0 +1,363 @@
...
+        ]
+
+        print("got_unaries", got_unary_segments)
+        assert got_unary_segments == want_unary_segments
+        assert got_stream_segments == want_stream_segments
+
+    def test_streaming_retryable_error(self):
+        add_select1_result()
+        add_error(SpannerServicer.ExecuteStreamingSql.__name__, unavailable_status())
+        add_error(SpannerServicer.ExecuteStreamingSql.__name__, unavailable_status())
```
olavloite, line 319: nit: I think that adding one error is enough

PR #1266, opened 2024-12-12
Updating the README in python spanner to include instructions for running spanner samples

aakashanandg:
- [x] Issue: https://github.com/googleapis/python-spanner/issues/1265
- [x] Ensure the tests and linter pass : Yes
- [x] Code coverage does not decrease : This change will not impact code coverage 
- [x] Appropriate docs were updated: Yes

Fixes #<1265> 🦕

### DIFF
--- a/samples/samples/README.rst
+++ b/samples/samples/README.rst
@@ -41,12 +41,16 @@ Install Dependencies
 
     .. code-block:: bash
 
-        $ git clone https://github.com/GoogleCloudPlatform/python-docs-samples.git
+        $ git clone https://github.com/googleapis/python-spanner.git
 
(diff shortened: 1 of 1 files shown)


olavloite (approved):
The change itself LGTM, but the pull request title needs to be changed. We use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) to generate the release notes. This means that:
1. All commits and pull request titles should start with a commit type (`feat:`, `docs:`, `fix:`, etc.) In this case, the type should be `docs:`
2. The rest of the header automatically goes into the release notes, so it should preferably be worded in a way that it reads naturally for someone reading it after it has been released. So in this case I would suggest `docs: update the README instructions for running Spanner samples`

PR #1270, opened 2024-12-16
fix: update retry strategy for mutation calls to handle aborted transactions

aakashanandg:
### Updating retry strategy for mutation calls to handle aborted transactions

* This PR updates the retry strategy for mutation calls to handle aborted transactions more effectively. Previously, the retry mechanism didn't handle certain edge cases for aborted transactions, leading to failures. The updated strategy ensures retries in these scenarios to improve the robustness of the mutation operations.

### Test Results:
* All unit tests related to the retry logic and mutation operations pass successfully.

Fixes #<1133> 🦕


google/cloud/spanner_v1/batch.py
```diff
@@ -217,10 +217,7 @@ def commit(
                 request=request,
                 metadata=metadata,
             )
-            response = _retry(
-                method,
-                allowed_exceptions={InternalServerError: _check_rst_stream_error},
-            )
+            response = self._session.run_in_transaction(method)
```
olavloite, line 220: Hmmm.... Now that I see this, I realize that this cannot use `run_in_transaction` directly. The reason is that:
1. This method `batch.commit()` in its original form (so without the changes in this pull request) creates and commits a single-use read/write transaction.
2. `run_in_transaction` however creates a new read/write transaction and then executes the given method in the scope of that transaction.

So what happens now is that you create two transactions:
1. `run_in_transaction` creates a transaction that is not being used.
2. `method` that is passed in to `run_in_transaction` executes a `Commit` in a single-use read/write transaction.

tests/unit/test_batch.py
```diff
@@ -263,6 +265,71 @@ def test_commit_ok(self):
...
+            batch.insert(TABLE_NAME, COLUMNS, VALUES)
+            
+            # Try committing the batch, which should call api.commit() and raise an Aborted exception
+            # The retry logic should call commit() again after handling the Aborted exception
+            try:
+                batch.commit()
+            except Aborted:
+                pass 
+            
+            # Verify that commit was called more than once (due to retry)
```
olavloite, line 331: You should verify the exact number. Based on my comment above, my guess is that you are getting 1 more commit than you would expect, as `run_in_transaction` also creates a read/write transaction.

google/cloud/spanner_v1/batch.py
aakashanandg: Thank you for pointing that out. I've updated the PR with a new retry strategy, replicating the same retry behavior as in the run_in_transaction method.

PR #1271, opened 2024-12-16
test: support inline-begin in mock server

olavloite:
Adds support for inline-begin in the mock server.

### DIFF
--- a/google/cloud/spanner_v1/testing/mock_spanner.py
+++ b/google/cloud/spanner_v1/testing/mock_spanner.py
@@ -18,6 +18,13 @@ from concurrent import futures
 
 from google.protobuf import empty_pb2
 from grpc_status.rpc_status import _Status
+
+from google.cloud.spanner_v1 import (
+    TransactionOptions,
+    ResultSetMetadata,
+    ExecuteSqlRequest,
+    ExecuteBatchDmlRequest,
+)
 from google.cloud.spanner_v1.testing.mock_database_admin import DatabaseAdminServicer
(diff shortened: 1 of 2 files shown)


google/cloud/spanner_v1/testing/mock_spanner.py
```diff
@@ -51,23 +58,25 @@ def pop_error(self, context):
...
             for row in result.rows:
                 partial = result_set.PartialResultSet()
                 if first:
-                    partial.metadata = result.metadata
+                    partial.metadata = ResultSetMetadata(result.metadata)
                 partial.values.extend(row)
                 partials.append(partial)
         partials[len(partials) - 1].stats = result.stats
+        if started_transaction:
+            partials[0].metadata.transaction = started_transaction
```
aakashanandg, line 79: Is there a specific reason we're including the transaction object details with the first partial result set?

olavloite: Yes, that is how the Spanner API works:
1. The first `PartialResultSet` in a stream should always contain the metadata for the query result. The remaining `PartialResultSet`s will not have this field set.
1. If the request (e.g. the `ExecuteSqlRequest`) contained a `BeginTransaction` option, then the metadata should also include the transaction ID of the transaction that was started as a side-effect of executing the request.

See https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.TransactionSelector for a bit more background information.

aakashanandg (approved):
LGTM. I just added a question for clarification on why we're including the transaction details with only a single partial result set

PR #1207, opened 2024-09-19
observability: annotate Session+SessionPool events

odeke-em:
This change adds annotations for session and session pool events to aid customers in debugging latency issues with session pool malevolence and also for maintainers to figure out which session pool type is the most appropriate.

Updates #1170

BEGIN_COMMIT_OVERRIDE
feat: add additional opentelemetry span events for session pool
END_COMMIT_OVERRIDE

PR #1274, opened 2024-12-17
observability: PDML + some batch write spans

odeke-em:
This change adds spans for Partitioned DML and making updates for Batch.

Carved out from PR #1241.

### DIFF
--- a/google/cloud/spanner_v1/database.py
+++ b/google/cloud/spanner_v1/database.py
@@ -699,46 +699,43 @@ class Database(object):
             )
 
         def execute_pdml():
-            def do_execute_pdml(session, span):
-                add_span_event(span, "Starting BeginTransaction")
-                txn = api.begin_transaction(
-                    session=session.name, options=txn_options, metadata=metadata
-                )
-
-                txn_selector = TransactionSelector(id=txn.id)
-
-                request = ExecuteSqlRequest(
-                    session=session.name,
-                    sql=dml,
-                    params=params_pb,
-                    param_types=param_types,
-                    query_options=query_options,
-                    request_options=request_options,
-                )
-                method = functools.partial(
-                    api.execute_streaming_sql,
-                    metadata=metadata,
-                )
-
-                iterator = _restart_on_unavailable(
-                    method=method,
-                    trace_name="CloudSpanner.ExecuteStreamingSql",
-                    request=request,
-                    transaction_selector=txn_selector,
-                    observability_options=self.observability_options,
-                )
-
-                result_set = StreamedResultSet(iterator)
-                list(result_set)  # consume all partials
-
-                return result_set.stats.row_count_lower_bound
-
             with trace_call(
                 "CloudSpanner.Database.execute_partitioned_pdml",
                 observability_options=self.observability_options,
             ) as span:
                 with SessionCheckout(self._pool) as session:
-                    return do_execute_pdml(session, span)
+                    add_span_event(span, "Starting BeginTransaction")
+                    txn = api.begin_transaction(
+                        session=session.name, options=txn_options, metadata=metadata
+                    )
+
+                    txn_selector = TransactionSelector(id=txn.id)
+
+                    request = ExecuteSqlRequest(
+                        session=session.name,
+                        sql=dml,
+                        params=params_pb,
+                        param_types=param_types,
+                        query_options=query_options,
+                        request_options=request_options,
+                    )
+                    method = functools.partial(
+                        api.execute_streaming_sql,
+                        metadata=metadata,
+                    )
+
+                    iterator = _restart_on_unavailable(
+                        method=method,
+                        trace_name="CloudSpanner.ExecuteStreamingSql",
+                        request=request,
+                        transaction_selector=txn_selector,
+                        observability_options=self.observability_options,
+                    )
+
+                    result_set = StreamedResultSet(iterator)
+                    list(result_set)  # consume all partials
+
+                    return result_set.stats.row_count_lower_bound
 
         return _retry_on_aborted(execute_pdml, DEFAULT_RETRY_BACKOFF)()
 
@@ -1539,7 +1536,7 @@ class BatchSnapshot(object):
         :rtype: :class:`~google.cloud.spanner_v1.streamed.StreamedResultSet`
         :returns: a result set instance which can be used to consume rows.
         """
-        observability_options = self.observability_options or {}
+        observability_options = self.observability_options
         with trace_call(
             f"CloudSpanner.{type(self).__name__}.process_read_batch",
             observability_options=observability_options,
--- a/google/cloud/spanner_v1/database.py
+++ b/google/cloud/spanner_v1/database.py
@@ -699,7 +699,8 @@ class Database(object):
             )
 
         def execute_pdml():
-            with SessionCheckout(self._pool) as session:
+            def do_execute_pdml(session, span):
+                add_span_event(span, "Starting BeginTransaction")
                 txn = api.begin_transaction(
                     session=session.name, options=txn_options, metadata=metadata
                 )
@@ -731,6 +732,13 @@ class Database(object):
                 list(result_set)  # consume all partials
 
                 return result_set.stats.row_count_lower_bound
+
+            with trace_call(
+                "CloudSpanner.Database.execute_partitioned_pdml",
+                observability_options=self.observability_options,
+            ) as span:
+                with SessionCheckout(self._pool) as session:
+                    return do_execute_pdml(session, span)
 
         return _retry_on_aborted(execute_pdml, DEFAULT_RETRY_BACKOFF)()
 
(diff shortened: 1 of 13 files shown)


odeke-em:
@harshachinta @olavloite kindly help me run the bots for this PR. Thank you.

odeke-em:
@harshachinta @olavloite  kindly help me re-run the bots, I had to account for Python3.7 in the tests and fixed that, all pass locally.

google/cloud/spanner_v1/merged_result_set.py
```diff
@@ -37,6 +39,16 @@ def __init__(self, batch_snapshot, partition_id, merged_result_set):
...
+        observability_options = getattr(
+            self._batch_snapshot, "observability_options", {}
+        )
+        with trace_call(
+            "CloudSpanner.PartitionExecutor.run",
+            observability_options=observability_options,
+        ):
+            return self.__run()
+
+    def __run(self):
```
harshachinta, line 51: I think this will not help since we already have a parent trace `run_partitioned_query` in database.py
You can remove this.

google/cloud/spanner_v1/session.py
```diff
@@ -470,6 +470,7 @@ def run_in_transaction(self, func, *args, **kw):
         ) as span:
             while True:
                 if self._transaction is None:
+                    add_span_event(span, "Creating Transaction")
```
harshachinta, line 473: This can be removed since `self.transaction() ` doesn't make an API call. It is just an inmemory trnasaction object that gets created

google/cloud/spanner_v1/snapshot.py
```diff
@@ -680,10 +680,13 @@ def partition_read(
         )
 
         trace_attributes = {"table_id": table, "columns": columns}
+        can_include_index = (index != "") and (index is not None)
+        if can_include_index:
+            trace_attributes["index"] = index
+
         with trace_call(
             f"CloudSpanner.{type(self).__name__}.partition_read",
-            self._session,
```
harshachinta, line 685: Lets keep this session here since we are updating last use time of session on the trace_call method..

tests/system/test_observability_options.py
```diff
@@ -241,30 +272,70 @@ def select_in_txn(txn):
...
+
+@pytest.mark.skipif(
+    not _helpers.USE_EMULATOR,
+    reason="Emulator needed to run this test",
+)
+@pytest.mark.skipif(
+    not HAS_OTEL_INSTALLED,
+    reason="Tracing requires OpenTelemetry",
+)
+def test_database_partitioned():
```
harshachinta, line 286: Looks like this is a negative case.
Can we test for one successful case for partitioned dml?

tests/system/test_observability_options.py
```diff
@@ -241,30 +272,70 @@ def select_in_txn(txn):
...
+            },
+        ),
+        (
+            "exception",
+            {
+                "exception.type": "google.api_core.exceptions.InvalidArgument",
+                "exception.message": "400 Table not found: NonExistent [at 1:8]\nUPDATE NonExistent SET name = 'foo' WHERE id > 1\n       ^",
+                "exception.stacktrace": "EPHEMERAL",
+                "exception.escaped": "False",
+            },
```
harshachinta, line 320: May be I missed some context,
Can you help me understand why we see 2 exceptions here instead of only one? This was fixed here right?
https://github.com/googleapis/python-spanner/pull/1207/files#r1862942500

tests/unit/test_snapshot.py
```diff
@@ -1225,12 +1230,18 @@ def test_partition_read_other_error(self):
         with self.assertRaises(RuntimeError):
             list(derived.partition_read(TABLE_NAME, COLUMNS, keyset))
 
+        if not HAS_OPENTELEMETRY_INSTALLED:
+            return
```
harshachinta, line 1234: This is not needed right since `assertSpanAttributes ` method already checks for this condition?

tests/unit/test_snapshot.py
```diff
@@ -1225,12 +1230,18 @@ def test_partition_read_other_error(self):
...
+            table_id=TABLE_NAME,
+            columns=tuple(COLUMNS),
+        )
         self.assertSpanAttributes(
             "CloudSpanner._Derived.partition_read",
             status=StatusCode.ERROR,
-            attributes=dict(
-                BASE_ATTRIBUTES, table_id=TABLE_NAME, columns=tuple(COLUMNS)
-            ),
+            attributes=want_span_attributes,
```
harshachinta, line 1244: Nothing has changed here? why not keep it the same?

tests/unit/test_snapshot.py
```diff
@@ -1368,10 +1379,11 @@ def _partition_query_helper(
             timeout=timeout,
         )
 
+        attributes = dict(BASE_ATTRIBUTES, **{"db.statement": SQL_QUERY_WITH_PARAM})
         self.assertSpanAttributes(
-            "CloudSpanner.PartitionReadWriteTransaction",
+            "CloudSpanner._Derived.partition_query",
             status=StatusCode.OK,
-            attributes=dict(BASE_ATTRIBUTES, **{"db.statement": SQL_QUERY_WITH_PARAM}),
```
harshachinta, line 1374: keep this as it is?

tests/unit/test_transaction.py
```diff
@@ -161,6 +162,14 @@ def test_begin_w_other_error(self):
         with self.assertRaises(RuntimeError):
             transaction.begin()
 
+        if not HAS_OPENTELEMETRY_INSTALLED:
+            return
+
+        span_list = self.get_finished_spans()
+        got_span_names = [span.name for span in span_list]
+        want_span_names = ["CloudSpanner.Transaction.begin"]
+        assert got_span_names == want_span_names
```
harshachinta, line 171: We are already asserting on this span below, do we still need this?

tests/unit/test_transaction.py
```diff
@@ -345,10 +354,23 @@ def test_commit_w_other_error(self):
...
+            return
+
+        span_list = sorted(self.get_finished_spans(), key=lambda v: v.start_time)
+        got_span_names = [span.name for span in span_list]
+        want_span_names = [
+            "CloudSpanner.Transaction.commit",
+        ]
+        assert got_span_names == want_span_names
+
+        txn_commit_span = span_list[-1]
```
harshachinta, line 367: same here, we are already asserting on span attributes below which is doing the same thing

google/cloud/spanner_v1/database.py
```diff
@@ -699,7 +699,8 @@ def execute_partitioned_dml(
             )
 
         def execute_pdml():
-            with SessionCheckout(self._pool) as session:
```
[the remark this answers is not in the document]
odeke-em: It was to keep the code much simpler and changes localized but sure, I can just indent it.

google/cloud/spanner_v1/merged_result_set.py
odeke-em: It does help because usually partitions are run in concurrent threads.

google/cloud/spanner_v1/session.py
odeke-em: Got it, thanks!

google/cloud/spanner_v1/snapshot.py
odeke-em: Done, thanks!

tests/unit/test_snapshot.py
odeke-em: Done.

tests/system/test_observability_options.py
odeke-em: An exception was thrown from the lower levels and with the chaining across with calls, it still gets propagated. The case we fixed was entirely different because in https://github.com/googleapis/python-spanner/issues/1246 that status was being overwritten back to OK.

tests/unit/test_snapshot.py
odeke-em: Yeah, before I was manually comparing attributes and span names but with the update that I made I no longer need it, done, thanks!

tests/unit/test_transaction.py
odeke-em: Done, thanks!

odeke-em:
@harshachinta kindly help me run the bots. Thank you

google/cloud/spanner_v1/merged_result_set.py
```diff
@@ -37,6 +39,16 @@ def __init__(self, batch_snapshot, partition_id, merged_result_set):
...
 
     def run(self):
+        observability_options = getattr(
+            self._batch_snapshot, "observability_options", {}
+        )
+        with trace_call(
+            "CloudSpanner.PartitionExecutor.run",
+            observability_options=observability_options,
+        ):
+            return self.__run()
```
harshachinta, line 49: The original self.run() method does not return anything.
Do we need this `return` here?

harshachinta (approved):

google/cloud/spanner_v1/merged_result_set.py
odeke-em: It would return None as a pass through for the original but sure, I can def remove it right now.

odeke-em:
@alkatrivedi @sakthivelmanii  kindly help me run the bots

PR #1275, opened 2024-12-18
feat(x-goog-spanner-request-id): introduce AtomicCounter

odeke-em:
This change introduces AtomicCounter, a concurrency/thread-safe counter do deal with the multi-threaded nature of variables. It permits operations:
* atomic_counter += 1
* value = atomic_counter + 1
* atomic_counter.value

that'll be paramount to bringing in the logic for
x-goog-spanner-request-id in much reduced changelists.

Carved out from PR googleapis/python-spanner#1264

### DIFF
--- a/google/cloud/spanner_v1/request_id_header.py
+++ b/google/cloud/spanner_v1/request_id_header.py
@@ -1,0 +1,42 @@
+# Copyright 2024 Google LLC All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import threading
+
+REQ_ID_VERSION = 1  # The version of the x-goog-spanner-request-id spec.
+REQ_ID_HEADER_KEY = "x-goog-spanner-request-id"
+
+
+def generate_rand_uint64():
+    b = os.urandom(8)
+    return (
+        b[7] & 0xFF
+        | (b[6] & 0xFF) << 8
+        | (b[5] & 0xFF) << 16
+        | (b[4] & 0xFF) << 24
+        | (b[3] & 0xFF) << 32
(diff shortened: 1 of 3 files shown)


tests/unit/test_atomic_counter.py
```diff
@@ -0,0 +1,81 @@
...
+            for i in range(m):
+                ac.increment()
+
+        threads = []
+        for i in range(n):
+            th = threading.Thread(target=do_work)
+            threads.append(th)
+            th.start()
+
+        time.sleep(0.3)
```
olavloite, line 73: Do we really need this?

tests/unit/test_atomic_counter.py
```diff
@@ -0,0 +1,81 @@
...
+
+        random.shuffle(threads)
```
olavloite, line 75: Is there a specific reason for doing this?

google/cloud/spanner_v1/request_id_header.py
```diff
@@ -0,0 +1,42 @@
...
+        | (b[2] & 0xFF) << 36
+        | (b[1] & 0xFF) << 48
+        | (b[0] & 0xFF) << 56
+    )
+
+
+REQ_RAND_PROCESS_ID = generate_rand_uint64()
+
+
+def with_request_id(client_id, channel_id, nth_request, attempt, other_metadata=[]):
```
olavloite, line 39: Could we add a simple test that verifies that the output of this method is in the format we expect?

olavloite (approved):

tests/unit/test_atomic_counter.py
odeke-em: Yes please, just there to give ample time for threads to complete.

tests/unit/test_atomic_counter.py
odeke-em: Yes please, it was to introduce more non-determinism to ensure correct functionality and not luck in threads being joined sequentially.

tests/unit/test_atomic_counter.py
olavloite: But that should be handled by the call to `th.join()`, right? That will block until the thread has finished.

google/cloud/spanner_v1/request_id_header.py
odeke-em: Done!

tests/unit/test_atomic_counter.py
odeke-em: Alright, that works, I really wanted to ensure ample time but indeed, Thread.join would block at least for one. Thanks for the update @olavloite!

PR #1302, opened 2025-01-24
feat: Add Attempt, Operation and GFE Metrics

lszinv:
Implementation of the MetricTracer and SpannerMetricFactory, a specific instance of the MetricFactory that populates the factory with metadata specific to the Python Spanner Client.

The Implemented metrics:

| Metric Name | Description | 
|--|--|
| Attempt Latency | latency of an RPC attempt. | 
| Attempt Count | The count of RPC attempts |
| Operation Latency | The total end-to-end latency across all request attempts, including retries. |
| Operation Count | The count of operations |
| GFE Latency | The latency of GFE |
| GFE Header missing count | The count of RPCs where GFE header is missing |


rahul2393 (commented):
As per offline discussion waiting for unit test to mock RPC failure and validate attempt_count>1

lszinv:
* Rebased onto up-to-date main
* updated the minimum version to address build failure
* Added unit test to verify attempt count metrics increasing on failure

lszinv:
Investigating the 3.7 test failure.

lszinv:
Added an importorskip to prevent the test running in Python3.7:

The test is performed by testing integration from the Client level. The metrics emission is tied with the OpenTelemetry trace logic, which is being skipped in 3.7. 

The logic could be separated and moved to wrap the trace calls instead of the other way around, but would turn every call like:
```
        with trace_call(
            "CloudSpanner.CreateSession",
            self,
            self._labels,
        ):
             ....
```

to
```
       with metrics_capture(): 
            with trace_call(
                "CloudSpanner.CreateSession",
                self,
                self._labels,
            ):
            ......
```
further increasing what is usually an already deep nesting.

My suggestion is that if it's acceptable for users to not expect having tracing in 3.7, then I think it's reasonable to say that 3.7 won't have metrics either.

I could however, make the above change if required.

lszinv:
Ignore the last commit. Will be pushing a new one to fix the test without disabling it.

lszinv:
Apologies, forgot to run the mockserver tests with the last change. I've fixed the test failures.

rahul2393 (approved):

PR #1327, opened 2025-03-25
feat: support transaction isolation level in dbapi

olavloite:
Adds API arguments and functions for setting a default isolation level and an
isolation level per transaction.

Support for specifying the isolation level using SQL commands will be added
in a follow-up PR.

### DIFF
--- a/google/cloud/spanner_dbapi/connection.py
+++ b/google/cloud/spanner_dbapi/connection.py
@@ -29,7 +29,7 @@ from google.cloud.spanner_dbapi.partition_helper import PartitionId
 from google.cloud.spanner_dbapi.parsed_statement import ParsedStatement, Statement
 from google.cloud.spanner_dbapi.transaction_helper import TransactionRetryHelper
 from google.cloud.spanner_dbapi.cursor import Cursor
-from google.cloud.spanner_v1 import RequestOptions
+from google.cloud.spanner_v1 import RequestOptions, TransactionOptions
 from google.cloud.spanner_v1.snapshot import Snapshot
 
 from google.cloud.spanner_dbapi.exceptions import (
@@ -112,6 +112,7 @@ class Connection:
         self._staleness = None
         self.request_priority = None
         self._transaction_begin_marked = False
+        self._transaction_isolation_level = None
         # whether transaction started at Spanner. This means that we had
         # made at least one call to Spanner.
         self._spanner_transaction_started = False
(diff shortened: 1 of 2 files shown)


google/cloud/spanner_dbapi/connection.py
```diff
@@ -283,6 +284,33 @@ def transaction_tag(self, value):
         """
         self._connection_variables["transaction_tag"] = value
 
+    @property
+    def isolation_level(self):
```
surbhigarg92, line 288: Shouldn't we add "default_transaction_options" similar to Python client library.

olavloite: We want a specific `isolation_level` property in the public API to be as compatible with other drivers as possible. Although it is not an official part of the standard, `isolation_level` is an argument and/or connection property that is supported by many dbapi drivers. Examples:
- PostgreSQL: https://www.psycopg.org/psycopg3/docs/api/connections.html#psycopg.Connection.isolation_level
- MySQL: https://dev.mysql.com/doc/connector-python/en/connector-python-api-mysqlconnection-start-transaction.html

We could consider using a `TransactionOptions` object for storage internally, but I'm not sure it would make things much simpler in this case, as this and other options are stored in a generic `_connection_variables` dict. This means that users can use these options directly in the connection string without any changes needed in the driver, e.g. the following works without any changes needed:

```
connection = Connection(self.instance, self.database, isolation_level=level)
```

surbhigarg92: Understood , thanks for explaining.

surbhigarg92 (approved):

google/cloud/spanner_dbapi/connection.py
```diff
@@ -421,6 +455,7 @@ def begin(self):
                 "is already running"
             )
         self._transaction_begin_marked = True
+        self._transaction_isolation_level = isolation_level
```
surbhigarg92, line 458: Shouldn't  `_transaction_isolation_level` fallback to `isolation_level` set at the connection if `isolation_level` is not provided in begin method ?

olavloite: Yes, but we do that at a slightly later moment when the transaction is actually being created here: https://github.com/googleapis/python-spanner/blob/82d15ba4fc1b38f4a604267b93a614bf20dfb031/google/cloud/spanner_dbapi/connection.py#L394

The reason for doing it like this is that this allows us to support a statement like `set transaction isolation level repeatable read` even after calling `begin`. (We don't support this yet in dbapi as it is part of the PostgreSQL dialect, but we do support this in for example JDBC).

The use of the default from the connection is covered by this test case: https://github.com/googleapis/python-spanner/blob/82d15ba4fc1b38f4a604267b93a614bf20dfb031/tests/mockserver_tests/test_dbapi_isolation_level.py#L71

PR #1328, opened 2025-03-25
fix: improve client-side regex statement parser

olavloite:
The client-side regex-based statement parser contained multiple minor errors, like:
- BEGIN would match any string as BEGIN TRANSACTION (including stuff like `BEGIN foo`)
- COMMIT and ROLLBACK had the same problem as BEGIN.
- Mismatches were reported as UPDATE. They are now returned as UNKNOWN.
- DLL missed the ANALYZE keyword

### DIFF
--- a/google/cloud/spanner_dbapi/batch_dml_executor.py
+++ b/google/cloud/spanner_dbapi/batch_dml_executor.py
@@ -54,9 +54,12 @@ class BatchDmlExecutor:
         """
         from google.cloud.spanner_dbapi import ProgrammingError
 
+        # Note: Let the server handle it if the client-side parser did not
+        # recognize the type of statement.
         if (
             parsed_statement.statement_type != StatementType.UPDATE
             and parsed_statement.statement_type != StatementType.INSERT
+            and parsed_statement.statement_type != StatementType.UNKNOWN
         ):
             raise ProgrammingError("Only DML statements are allowed in batch DML mode.")
         self._statements.append(parsed_statement.statement)
--- a/google/cloud/spanner_dbapi/connection.py
+++ b/google/cloud/spanner_dbapi/connection.py
@@ -20,11 +20,7 @@ from google.api_core.gapic_v1.client_info import ClientInfo
 from google.cloud import spanner_v1 as spanner
 from google.cloud.spanner_dbapi import partition_helper
 from google.cloud.spanner_dbapi.batch_dml_executor import BatchMode, BatchDmlExecutor
-from google.cloud.spanner_dbapi.parse_utils import _get_statement_type
-from google.cloud.spanner_dbapi.parsed_statement import (
-    StatementType,
-    AutocommitDmlMode,
-)
+from google.cloud.spanner_dbapi.parsed_statement import AutocommitDmlMode
 from google.cloud.spanner_dbapi.partition_helper import PartitionId
 from google.cloud.spanner_dbapi.parsed_statement import ParsedStatement, Statement
 from google.cloud.spanner_dbapi.transaction_helper import TransactionRetryHelper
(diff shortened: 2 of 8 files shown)


google/cloud/spanner_dbapi/batch_dml_executor.py
```diff
@@ -54,9 +54,12 @@ def execute_statement(self, parsed_statement: ParsedStatement):
         """
         from google.cloud.spanner_dbapi import ProgrammingError
 
+        # Note: Let the server handle it if the client-side parser did not
+        # recognize the type of statement.
         if (
             parsed_statement.statement_type != StatementType.UPDATE
             and parsed_statement.statement_type != StatementType.INSERT
+            and parsed_statement.statement_type != StatementType.UNKNOWN
```
olavloite, line 62: This provides a fallback if the statement is not recognized by the client-side parser. Instead of failing directly, we allow it to be sent to Spanner and let the Spanner parser determine whether it is valid or not. This prevents potential new DML syntax in the future from being rejected here (e.g. if we were to support `merge` statements).

google/cloud/spanner_dbapi/connection.py
```diff
@@ -702,10 +698,6 @@ def set_autocommit_dml_mode(
         self._autocommit_dml_mode = autocommit_dml_mode
 
     def _partitioned_query_validation(self, partitioned_query, statement):
-        if _get_statement_type(Statement(partitioned_query)) is not StatementType.QUERY:
```
olavloite, line 705: Skip validation here and let Spanner return an error if the statement is not valid for `PartitionQuery`.

google/cloud/spanner_dbapi/parse_utils.py
```diff
@@ -155,25 +155,30 @@
...
+# TODO: This and the other regexes do not match statements that start with a hint.
 RE_NON_UPDATE = re.compile(r"^\W*(SELECT|GRAPH|FROM)", re.IGNORECASE)
 
 RE_WITH = re.compile(r"^\s*(WITH)", re.IGNORECASE)
 
 # DDL statements follow
 # https://cloud.google.com/spanner/docs/data-definition-language
 RE_DDL = re.compile(
-    r"^\s*(CREATE|ALTER|DROP|GRANT|REVOKE|RENAME)", re.IGNORECASE | re.DOTALL
+    r"^\s*(CREATE|ALTER|DROP|GRANT|REVOKE|RENAME|ANALYZE)", re.IGNORECASE | re.DOTALL
```
olavloite, line 166: `ANALYZE` is a valid DDL statement in Spanner.

google/cloud/spanner_dbapi/parse_utils.py
```diff
@@ -155,25 +155,30 @@
...
 
 RE_INSERT = re.compile(
     # Only match the `INSERT INTO <table_name> (columns...)
     # otherwise the rest of the statement could be a complex
     # operation.
-    r"^\s*INSERT INTO (?P<table_name>[^\s\(\)]+)\s*\((?P<columns>[^\(\)]+)\)",
+    r"^\s*INSERT(?:\s+INTO)?\s+(?P<table_name>[^\s\(\)]+)\s*\((?P<columns>[^\(\)]+)\)",
     re.IGNORECASE | re.DOTALL,
 )
+"""Deprecated: Use the RE_IS_INSERT, RE_IS_UPDATE, and RE_IS_DELETE regexes"""
```
olavloite, line 181: This is not a new deprecation, these regular expressions were already 'deprecated' in the sense that:
1. They were no longer in use by the library
2. They are not intended for public use

aakashanandg (approved):

PR #1329, opened 2025-03-26
feat: Add support for multiplexed sessions - part 1 (supporting classes and refactoring)

currantw:
### Description

- Adds `SessionOptions` and `DatabaseSessionManager` class support multiplexed session in future work.
- Update `Client`, `Instance`, `Database`, `Session` and associated `Checkout` classes accordingly.
- Add unit tests for new classes and update those for existing classes.

### DIFF
--- a/google/cloud/spanner_v1/database.py
+++ b/google/cloud/spanner_v1/database.py
@@ -40,6 +40,7 @@ from google.cloud.spanner_admin_database_v1 import RestoreDatabaseEncryptionConf
 from google.cloud.spanner_admin_database_v1 import RestoreDatabaseRequest
 from google.cloud.spanner_admin_database_v1 import UpdateDatabaseDdlRequest
 from google.cloud.spanner_admin_database_v1.types import DatabaseDialect
+from google.cloud.spanner_v1.database_sessions_manager import DatabaseSessionsManager
 from google.cloud.spanner_v1.transaction import BatchTransactionId
 from google.cloud.spanner_v1 import ExecuteSqlRequest
 from google.cloud.spanner_v1 import Type
@@ -59,7 +60,6 @@ from google.cloud.spanner_v1.batch import MutationGroups
 from google.cloud.spanner_v1.keyset import KeySet
 from google.cloud.spanner_v1.merged_result_set import MergedResultSet
 from google.cloud.spanner_v1.pool import BurstyPool
-from google.cloud.spanner_v1.pool import SessionCheckout
 from google.cloud.spanner_v1.session import Session
 from google.cloud.spanner_v1.snapshot import _restart_on_unavailable
 from google.cloud.spanner_v1.snapshot import Snapshot
@@ -70,7 +70,6 @@ from google.cloud.spanner_v1.services.spanner.transports.grpc import (
 from google.cloud.spanner_v1.table import Table
 from google.cloud.spanner_v1._opentelemetry_tracing import (
     add_span_event,
-    get_current_span,
     trace_call,
 )
 from google.cloud.spanner_v1.metrics.metrics_capture import MetricsCapture
@@ -191,9 +190,9 @@ class Database(object):
 
         if pool is None:
             pool = BurstyPool(database_role=database_role)
-
-        self._pool = pool
         pool.bind(self)
+
+        self._session_manager = DatabaseSessionsManager(database=self, pool=pool)
 
     @classmethod
     def from_pb(cls, database_pb, instance, pool=None):
@@ -708,7 +707,7 @@ class Database(object):
                 "CloudSpanner.Database.execute_partitioned_pdml",
                 observability_options=self.observability_options,
             ) as span, MetricsCapture():
-                with SessionCheckout(self._pool) as session:
+                with SessionCheckout(self) as session:
                     add_span_event(span, "Starting BeginTransaction")
                     txn = api.begin_transaction(
                         session=session.name, options=txn_options, metadata=metadata
@@ -923,7 +922,7 @@ class Database(object):
             # Check out a session and run the function in a transaction; once
             # done, flip the sanity check bit back.
             try:
-                with SessionCheckout(self._pool) as session:
+                with SessionCheckout(self) as session:
                     return session.run_in_transaction(func, *args, **kw)
             finally:
                 self._local.transaction_running = False
@@ -1160,6 +1159,35 @@ class Database(object):
         return opts
 
 
+class SessionCheckout(object):
+    """Context manager for using a session from a database.
+
+    :type database: :class:`~google.cloud.spanner_v1.database.Database`
+    :param database: database to use the session from
+    """
+
+    _session = None  # Not checked out until '__enter__'.
+
+    def __init__(self, database):
+        if not isinstance(database, Database):
+            raise TypeError(
+                "{class_name} must receive an instance of {expected_class_name}. Received: {actual_class_name}".format(
+                    class_name=self.__class__.__name__,
+                    expected_class_name=Database.__name__,
+                    actual_class_name=database.__class__.__name__,
+                )
+            )
+
+        self._database = database
+
+    def __enter__(self):
+        self._session = self._database._session_manager.get_session_for_read_write()
+        return self._session
+
+    def __exit__(self, *ignored):
+        self._database._session_manager.put_session(self._session)
+
+
 class BatchCheckout(object):
     """Context manager for using a batch from a database.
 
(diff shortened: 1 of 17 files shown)


google/cloud/spanner_v1/database.py
```diff
@@ -1160,6 +1160,29 @@ def observability_options(self):
...
+            raise TypeError(f"database must be an instance of {Database.__name__}")
+
+        self._database = database
+
+    def __enter__(self):
+        self._session = self._database._session_manager.get_session_for_read_write()
+        return self._session
+
+    def __exit__(self, *ignored):
+        self._database._session_manager.put_session(self._session)
```
currantw, line 1183: `SessionCheckout` has been moved from `pool.py` to `database.py`, and updated so that the session is checked out from the _database_ rather than from the _pool_.

Furthermore, `SessionCheckout`'s constructor previous accepted arbitrary keyword arguments, which were passed to the pool's `get` method. While `AbstractSessionPool.get` does not describe any keyword arguments, both `FixedSizePool.get` and `PingingPool.get` accept an optional `timeout` keyword argument, which can be used to override the default timeout.

Because `DatabaseSessionManager` now manages which session is returned, and it not guaranteed to return a session from the pool, we are suggesting removing the ability to provide arbitrary keyword arguments to the pool via `SessionCheckout.__init__`.

No keyword arguments are supported by the default pool implementation (`BurstyPool`), and users are already able to specify their own _default_ timeout for `FixedSizePool` and `PingingPool`. Moreover, it is not clear to us where this functionality is even used: remaining uses of `SessionCheckout` do not support  these keyword arguments.

Please let us know:
- Is it acceptable to remove this functionality?
- Would you also like to remove support for the `timeout` arguments to `FixedSizePool` and `PingingPool`, for consistency?

google/cloud/spanner_v1/database.py
```diff
@@ -1194,6 +1217,9 @@ def __init__(
         isolation_level=TransactionOptions.IsolationLevel.ISOLATION_LEVEL_UNSPECIFIED,
         **kw,
     ):
+        if not isinstance(database, Database):
+            raise TypeError(f"database must be an instance of {Database.__name__}")
```
currantw, line 1221: For better error handling, we added these type checks to `SessionCheckout`, `BatchCheckout`, `MutationGroupsCheckout`, and `SnapshotCheckout`, as well as corresponding unit tests.

google/cloud/spanner_v1/database.py
```diff
@@ -1258,23 +1278,22 @@ class MutationGroupsCheckout(object):
...
         return MutationGroups(session)
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         """End ``with`` block."""
-        if isinstance(exc_val, NotFound):
-            # If NotFound exception occurs inside the with block
-            # then we validate if the session still exists.
-            if not self._session.exists():
-                self._session = self._database._pool._new_session()
-                self._session.create()
```
currantw, line 1276: This check has been removed from `MutationGroupsCheckout` and `SnapshotCheckout` because:
- `exists` only checks whether the session exists on the server (not whether the local session object exists)
- This checkout wasn't for all context managers (not in `BatchCheckout` or `SessionCheckout`)
- This seems to only be done as a slight optimization to potentially reduce overhead of session retrieval by adding it to the tail of a request instead of the head of another.
- In reality, the optimization only plays a part if the session expires while a call was happening, and is unlikely to make a big difference.
- `exists` checks are still performed by the session pool (for non-multiplexed session) and the database session manager (for multiplexed session).

We recommend removing this code, and instead depend on `DatabaseSessionManager` to take care of checking session existence and re-creating them if necessary, rather than (sometimes) handling this in the context manager.

google/cloud/spanner_v1/database.py
```diff
@@ -1160,6 +1159,35 @@ def observability_options(self):
...
+            )
+
+        self._database = database
+
+    def __enter__(self):
+        self._session = self._database._session_manager.get_session_for_read_write()
+        return self._session
+
+    def __exit__(self, *ignored):
+        self._database._session_manager.put_session(self._session)
```
currantw, line 1188: `SessionCheckout` has been moved from `pool.py` to `database.py`, and updated so that the session is checked out from the _database_ rather than from the _pool_.

Furthermore, `SessionCheckout`'s constructor previous accepted arbitrary keyword arguments, which were passed to the pool's `get` method. While `AbstractSessionPool.get` does not describe any keyword arguments, both `FixedSizePool.get` and `PingingPool.get` accept an optional `timeout` keyword argument, which can be used to override the default timeout.

Because `DatabaseSessionManager` now manages which session is returned, and it not guaranteed to return a session from the pool, we are suggesting removing the ability to provide arbitrary keyword arguments to the pool via `SessionCheckout.__init__`.

No keyword arguments are supported by the default pool implementation (`BurstyPool`), and users are already able to specify their own _default_ timeout for `FixedSizePool` and `PingingPool`. Moreover, it is not clear to us where this functionality is even used: remaining uses of `SessionCheckout` do not support  these keyword arguments.

Please let us know:
- Is it acceptable to remove this functionality?
- Would you also like to remove support for the `timeout` arguments to `FixedSizePool` and `PingingPool`, for consistency?

google/cloud/spanner_v1/database.py
```diff
@@ -1194,6 +1222,15 @@ def __init__(
...
         **kw,
     ):
+        if not isinstance(database, Database):
+            raise TypeError(
+                "{class_name} must receive an instance of {expected_class_name}. Received: {actual_class_name}".format(
+                    class_name=self.__class__.__name__,
+                    expected_class_name=Database.__name__,
+                    actual_class_name=database.__class__.__name__,
+                )
+            )
```
currantw, line 1232: For better error handling, added a type check to all `*Checkout` classes, along with corresponding unit tests.

tests/unit/test_database.py
```diff
@@ -1801,7 +1799,11 @@ def _make_spanner_client():
         return mock.create_autospec(SpannerClient)
 
     def test_ctor(self):
-        database = _Database(self.DATABASE_NAME)
+        from google.cloud.spanner_v1.database import Database
+
+        database = Database(
+            database_id=self.DATABASE_ID, instance=_Instance(self.INSTANCE_NAME)
+        )
```
currantw, line 1806: Because of the added type check in the constructors for `*Checkout` classes (see comment [here](https://github.com/googleapis/python-spanner/pull/1329/files#r2015126699)), these tests have been updated to use a `Database` instance instead of `_Database`.

tests/unit/test_database.py
```diff
@@ -1979,129 +2018,90 @@ def _get_target_class(self):
...
-        class Testing(Exception):
-            pass
-
-        self.assertEqual(pool._session, session)
-        with self.assertRaises(Testing):
-            with checkout as _:
-                raise Testing("Unknown error.")
-        # Assert that session-1 was not removed from pool.
-        self.assertEqual(pool._session, session)
-        pool._new_session.assert_not_called()
```
currantw, line 2104: These tests are no longer required because the context managers are no longer responsible for re-creating the session if it fails with a `NotFound` error. See this comment for more details: https://github.com/googleapis/python-spanner/pull/1329/files#r2015092908

tests/unit/test_database.py
```diff
@@ -3020,86 +3032,105 @@ def test_context_mgr_success(self):
...
+        from google.cloud.spanner_v1.database import Database
+
+        database = Database(
+            database_id=self.DATABASE_ID, instance=_Instance(self.INSTANCE_NAME)
+        )
+
         session = _Session(database)
-        pool = database._pool = _Pool()
-        pool._new_session = mock.MagicMock(return_value=[])
-        pool.put(session)
```
currantw, line 3090: These tests are no longer required because the context managers are no longer responsible for re-creating the session if it fails with a `NotFound` error. See this comment for more details: https://github.com/googleapis/python-spanner/pull/1329/files#r2015092908

tests/unit/test_pool.py
```diff
@@ -1073,73 +1053,6 @@ def test_spans_get_and_leave_empty_pool(self):
...
-        self.assertEqual(len(pool._items), 1)
-        self.assertIs(pool._items[0], session)
-
-        with checkout as borrowed:
-            self.assertIs(borrowed, session)
-            self.assertEqual(len(pool._items), 0)
-
-        self.assertEqual(len(pool._items), 1)
-        self.assertIs(pool._items[0], session)
-        self.assertEqual(pool._got, {"foo": "bar"})
```
currantw, line 1131: These tests have mostly been moved to `test_database.py` (since `SessionCheckout` has moved to `database.py`). The "w_kwargs" tests have been removed because they are no longer supported: see this comment here: https://github.com/googleapis/python-spanner/pull/1329/files#r2015125513

tests/unit/test_pool.py
```diff
@@ -1073,73 +1053,6 @@ def test_spans_get_and_leave_empty_pool(self):
...
-
-
-def _make_transaction(*args, **kw):
-    from google.cloud.spanner_v1.transaction import Transaction
-
-    txn = mock.create_autospec(Transaction)(*args, **kw)
-    txn.committed = None
-    txn.rolled_back = False
-    return txn
```
currantw, line 1140: This method was unused so I just removed it.

tests/unit/test_pool.py
```diff
@@ -1183,14 +1096,6 @@ def delete(self):
         if not self._exists:
             raise NotFound("unknown session")
 
-    def transaction(self):
-        txn = self._transaction = _make_transaction(self)
-        return txn
-
-    @property
-    def session_id(self):
-        return self._session_id
```
currantw, line 1192: These were unused, so I just removed them.

PR #1339, opened 2025-04-22
chore: edition sample fix with increased timeout

harshachinta:


samples/samples/snippets.py
```diff
@@ -91,13 +91,13 @@ def update_instance(instance_id):
...
             },
-            edition=spanner_instance_admin.Instance.Edition.STANDARD,  # Optional
+            edition=spanner_instance_admin.Instance.Edition.ENTERPRISE,  # Optional
         ),
         field_mask=field_mask_pb2.FieldMask(paths=["labels", "edition"]),
     )
 
     print("Waiting for operation to complete...")
-    operation.result(OPERATION_TIMEOUT_SECONDS)
+    operation.result(900)
```
harshachinta, line 100: With the current timeout of 240 seconds, the sample is failing. When we increase it to 900 seconds it works.
Raised the issue to the editions team. Meanwhile to unblock PR merges and prevent samples failure this PR will help.

sakthivelmanii (approved):

samples/samples/snippets.py
```diff
@@ -91,13 +91,13 @@ def update_instance(instance_id):
             labels={
                 "sample_name": "snippets-update_instance-explicit",
             },
-            edition=spanner_instance_admin.Instance.Edition.STANDARD,  # Optional
+            edition=spanner_instance_admin.Instance.Edition.ENTERPRISE,  # Optional
```
harshachinta, line 94: In order to merge split API sample I had to change this to STANDARD in that PR so that the tests pass.
Reverting back that change in this PR.

samples/samples/snippets.py
harshachinta: Confirmed with editions team that is expected behaviour.
> Yes, this is expected. Previously we weren't waiting for galaxy and spanfe caches to be populated with the modified edition. Now we are. This takes approximately 10 min.

PR #1340, opened 2025-04-23
feat: add interval type support

rahul2393:

### DIFF
--- a/google/cloud/spanner_v1/_helpers.py
+++ b/google/cloud/spanner_v1/_helpers.py
@@ -20,8 +20,6 @@ import math
 import time
 import base64
 import threading
-import re
-from dataclasses import dataclass
 
 from google.protobuf.struct_pb2 import ListValue
 from google.protobuf.struct_pb2 import Value
@@ -33,7 +31,7 @@ from google.api_core.exceptions import Aborted
 from google.cloud._helpers import _date_from_iso8601_date
 from google.cloud.spanner_v1 import TypeCode
 from google.cloud.spanner_v1 import ExecuteSqlRequest
-from google.cloud.spanner_v1 import JsonObject
+from google.cloud.spanner_v1 import JsonObject, Interval
 from google.cloud.spanner_v1 import TransactionOptions
 from google.cloud.spanner_v1.request_id_header import with_request_id
 from google.rpc.error_details_pb2 import RetryInfo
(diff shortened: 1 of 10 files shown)


google/cloud/spanner_v1/_helpers.py
```diff
@@ -196,6 +198,152 @@ def _datetime_to_rfc3339_nanoseconds(value):
     return "{}.{}Z".format(value.isoformat(sep="T", timespec="seconds"), nanos)
 
 
+@dataclass
+class Interval:
```
surbhigarg92, line 202: In Node , I remember there were a lot of method added to create an Interval type using Months, Days etc.  Eg: 
```
 /**
   * Constructs an `Interval` with specified days.
   */
  static fromDays(days: number): Interval {
    return new Interval(0, days, BigInt(0));
  }
  ```

Comment (surbhigarg92) [inline google/cloud/spanner_v1/_helpers.py:202]: Why this class is added in _helpers file. Shouldn't it be added in `data_types.py` file

File: tests/unit/test__helpers.py [hunk 2]
```diff
@@ -1036,3 +1036,665 @@ def test_default_isolation_and_merge_options_isolation_unspecified(self):
...
+
+
+class Test_interval(unittest.TestCase):
+    def _callFUT(self, *args, **kw):
+        from google.cloud.spanner_v1._helpers import _make_value_pb
+
+        return _make_value_pb(*args, **kw)
+
+    def test_basic_interval(self):
+        from google.protobuf.struct_pb2 import Value
```
surbhigarg92, line 1048: nit: Move the imports to class level as all the tests uses the same imports.

tests/unit/test__helpers.py
```diff
@@ -1036,3 +1036,665 @@ def test_default_isolation_and_merge_options_isolation_unspecified(self):
...
+        from google.cloud.spanner_v1 import Type
+        from google.cloud.spanner_v1 import TypeCode
+        from google.cloud.spanner_v1._helpers import Interval
+
+        interval = Interval(months=10, days=0, nanos=0)
+        field_type = Type(code=TypeCode.INTERVAL)
+        value_pb = self._callFUT(interval)
+        self.assertIsInstance(value_pb, Value)
+        self.assertEqual(value_pb.string_value, "P10M")
```
surbhigarg92, line 1069: nit: Could we have a single test instead where we maintain the actual and expected values array and loop over it.

google/cloud/spanner_v1/_helpers.py
rahul2393: If we want to create that we can do with current code here
`i = Interval(days=number)`
there is no need of exposing extra method here

rahul2393: moved

tests/unit/test__helpers.py
rahul2393: done

tests/unit/test__helpers.py
rahul2393: done

rahul2393:
Build/Test failures are unrelated here, working with release team to get this unblocked.

surbhigarg92 (approved):

google/cloud/spanner_v1/_helpers.py
surbhigarg92: Sounds good. Please if we need documentation for it.

sagarwaal (approved):

PR #1364, opened 2025-04-30
chore(x-goog-spanner-request-id): implement request_id generation and propagation 1/3

odeke-em:
Generates a request_id that is then injected inside metadata that's sent over to the Cloud Spanner backend.

Officially inject the first set of x-goog-spanner-request-id values into header metadata Add request-id interceptor to use in asserting tests Wrap Snapshot methods with x-goog-request-id metadata injector


odeke-em:
No longer needed as I mailed out smaller PRs.

PR #1366, opened 2025-04-30
chore(x-goog-request-id): commit testing scaffold

odeke-em:
This change commits the scaffolding for which testing will be used. This is a carve out of PRs googleapis/python-spanner#1264 and googleapis/python-spanner#1364, meant to make those changes lighter and much easier to review then merge.

Updates googleapis/google-cloud-python#15905

PR #1367, opened 2025-04-30
chore(x-goog-spanner-request-id): plug in functionality after test scaffolding

odeke-em:
This change chops down the load of the large changes for x-goog-spanner-request-id. It depends on PR googleapis/python-spanner#1366 and should only be merged after that PR.

Requires PR googleapis/python-spanner#1366

### DIFF
--- a/google/cloud/spanner_v1/batch.py
+++ b/google/cloud/spanner_v1/batch.py
@@ -26,7 +26,6 @@ from google.cloud.spanner_v1._helpers import (
     _metadata_with_prefix,
     _metadata_with_leader_aware_routing,
     _merge_Transaction_Options,
-    AtomicCounter,
 )
 from google.cloud.spanner_v1._opentelemetry_tracing import trace_call
 from google.cloud.spanner_v1 import RequestOptions
@@ -250,6 +249,7 @@ class Batch(_BatchBase):
             observability_options=observability_options,
             metadata=metadata,
         ), MetricsCapture():
+
             def wrapped_method(*args, **kwargs):
                 method = functools.partial(
                     api.commit,
--- a/google/cloud/spanner_v1/batch.py
+++ b/google/cloud/spanner_v1/batch.py
@@ -250,15 +250,17 @@ class Batch(_BatchBase):
             observability_options=observability_options,
             metadata=metadata,
         ), MetricsCapture():
-            attempt = AtomicCounter()
-            nth_request = getattr(database, "_next_nth_request", 0)
-
             def wrapped_method(*args, **kwargs):
                 method = functools.partial(
                     api.commit,
                     request=request,
                     metadata=database.metadata_with_request_id(
-                        nth_request, attempt.increment(), metadata
+                        # This code is retried due to ABORTED, hence nth_request
+                        # should be increased. attempt can only be increased if
+                        # we encounter UNAVAILABLE or INTERNAL.
+                        getattr(database, "_next_nth_request", 0),
+                        1,
+                        metadata,
                     ),
                 )
                 return method(*args, **kwargs)
--- a/google/cloud/spanner_v1/batch.py
+++ b/google/cloud/spanner_v1/batch.py
@@ -26,6 +26,7 @@ from google.cloud.spanner_v1._helpers import (
     _metadata_with_prefix,
     _metadata_with_leader_aware_routing,
     _merge_Transaction_Options,
+    AtomicCounter,
 )
 from google.cloud.spanner_v1._opentelemetry_tracing import trace_call
 from google.cloud.spanner_v1 import RequestOptions
@@ -249,17 +250,25 @@ class Batch(_BatchBase):
             observability_options=observability_options,
             metadata=metadata,
         ), MetricsCapture():
-            method = functools.partial(
-                api.commit,
-                request=request,
-                metadata=metadata,
-            )
+            attempt = AtomicCounter()
+            nth_request = getattr(database, "_next_nth_request", 0)
+
+            def wrapped_method(*args, **kwargs):
+                method = functools.partial(
+                    api.commit,
+                    request=request,
+                    metadata=database.metadata_with_request_id(
+                        nth_request, attempt.increment(), metadata
+                    ),
+                )
+                return method(*args, **kwargs)
+
             deadline = time.time() + kwargs.get(
                 "timeout_secs", DEFAULT_RETRY_TIMEOUT_SECS
             )
             default_retry_delay = kwargs.get("default_retry_delay", None)
             response = _retry_on_aborted_exception(
-                method,
+                wrapped_method,
                 deadline=deadline,
                 default_retry_delay=default_retry_delay,
             )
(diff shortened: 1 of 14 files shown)


odeke-em:
@rahul2393 @olavloite @harshachinta @sakthivelmanii kindly help run the bots. Thank you.

google/cloud/spanner_v1/database.py
```diff
@@ -745,6 +798,16 @@ def execute_pdml():
...
 
+    @property
+    def _next_nth_request(self):
+        if self._instance and self._instance._client:
+            return self._instance._client._next_nth_request
+        return 1
+
+    @property
+    def _nth_client_id(self):
+        return self._instance._client._nth_client_id
```
olavloite, line 809: Is there a specific reason that we don't need a check like this here?

```python
if self._instance and self._instance._client:
```

google/cloud/spanner_v1/database.py
```diff
@@ -965,7 +1028,8 @@ def restore(self, source):
         )
         future = api.restore_database(
             request=request,
-            metadata=metadata,
+            # TODO: Infer the channel_id being used.
```
olavloite, line 1031: This TODO seems outdated

google/cloud/spanner_v1/session.py
```diff
@@ -255,7 +266,10 @@ def delete(self):
             observability_options=observability_options,
             metadata=metadata,
         ), MetricsCapture():
-            api.delete_session(name=self.name, metadata=metadata)
+            api.delete_session(
+                name=self.name,
+                metadata=metadata,
```
olavloite, line 271: Is this missing a call to `database.metadata_with_request_id`?

google/cloud/spanner_v1/snapshot.py
```diff
@@ -101,7 +108,13 @@ def _restart_on_unavailable(
                     observability_options=observability_options,
                     metadata=metadata,
                 ), MetricsCapture():
-                    iterator = method(request=request, metadata=metadata)
+                    attempt += 1
```
olavloite, line 111: This isn't really needed, as this call will only be used the first time (due to the `if iterator is None` check above)

google/cloud/spanner_v1/snapshot.py
```diff
@@ -129,7 +142,14 @@ def _restart_on_unavailable(
                 if transaction is not None:
                     transaction_selector = transaction._make_txn_selector()
                 request.transaction = transaction_selector
-                iterator = method(request=request)
+                nth_request = next_nth_request()
+                attempt = 1
```
olavloite, line 146: This should not reset the attempt to 1. This should increase the attempt, as this is a retry of the query due to an UNAVAILABLE error.

google/cloud/spanner_v1/snapshot.py
```diff
@@ -129,7 +142,14 @@ def _restart_on_unavailable(
                 if transaction is not None:
                     transaction_selector = transaction._make_txn_selector()
                 request.transaction = transaction_selector
-                iterator = method(request=request)
+                nth_request = next_nth_request()
```
olavloite, line 145: This is a retry, so it should use the same nth_request, but a higher attempt number.

google/cloud/spanner_v1/snapshot.py
```diff
@@ -149,8 +169,15 @@ def _restart_on_unavailable(
                 request.resume_token = resume_token
                 if transaction is not None:
                     transaction_selector = transaction._make_txn_selector()
+                nth_request = next_nth_request()
+                attempt = 1
```
olavloite, line 173: Same here: This is a retry of the query, and should use the same nth request and a higher attempt number.

google/cloud/spanner_v1/snapshot.py
```diff
@@ -969,14 +1023,24 @@ def begin(self):
...
             metadata=metadata,
         ), MetricsCapture():
-            method = functools.partial(
-                api.begin_transaction,
-                session=self._session.name,
-                options=txn_selector.begin,
-                metadata=metadata,
-            )
+            nth_request = getattr(database, "_next_nth_request", 0)
+            counters = dict(attempt=0)
```
olavloite, line 1027: Is there a specific reason that this uses a dictionary with an attempt, instead of an atomic counter?

tests/unit/test_spanner.py
```diff
@@ -911,32 +1000,47 @@ def test_transaction_for_concurrent_statement_should_begin_one_transaction_with_
...
-            timeout=TIMEOUT,
-        )
-
-        api.execute_batch_dml.assert_any_call(
-            request=self._batch_update_expected_request(begin=False),
-            metadata=[
-                ("google-cloud-resource-prefix", database.name),
-                ("x-goog-spanner-route-to-leader", "true"),
+        self.assertEqual(api.execute_batch_dml.call_count, 2)
+        print("\033[34marg_list", api.execute_batch_dml.call_args_list, "\033[00m")
```
olavloite, line 1011: nit: remove

google/cloud/spanner_v1/database.py
odeke-em: More of because we plumbed all the callers to have an _instance and _client: in test and then unlikely that either will be None when correctly setup. I however can make that change, no biggie.

google/cloud/spanner_v1/session.py
odeke-em: Right above on line 255 we created it, to be used inside trace_call but sure I can bring it down and inline it, since anyways trace_call's updates will be added later on independently

google/cloud/spanner_v1/snapshot.py
odeke-em: The very interesting thing is that I have been treating InternalServerError as a non-idempotent error: as it could result in inconsistent state as an operation might have succeeded on the server but it crapped out. However, good points, this makes it consistent with the other implementations and thank you.

google/cloud/spanner_v1/snapshot.py
odeke-em: It is a legacy artifact, fixed and thank you!

google/cloud/spanner_v1/snapshot.py
odeke-em: done, thanks!

google/cloud/spanner_v1/snapshot.py
odeke-em: Gotcha and thanks! Done.

google/cloud/spanner_v1/batch.py
```diff
@@ -249,17 +250,25 @@ def commit(
...
-            )
+            attempt = AtomicCounter()
+            nth_request = getattr(database, "_next_nth_request", 0)
+
+            def wrapped_method(*args, **kwargs):
+                method = functools.partial(
+                    api.commit,
+                    request=request,
+                    metadata=database.metadata_with_request_id(
+                        nth_request, attempt.increment(), metadata
```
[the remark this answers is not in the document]
odeke-em: Done, thanks!

odeke-em:
@olavloite thank you for the first round of review. I have addressed it, please take another look and kindly help me re-run the bots.

olavloite (approved):

sakthivelmanii (approved):

PR #1375, opened 2025-05-18
chore(x-goog-spanner-request-id): more updates for batch_write + mockserver tests

odeke-em:
This change plumbs in some x-goog-spanner-request-id updates for batch_write and some tests too.

### DIFF
--- a/google/cloud/spanner_v1/_helpers.py
+++ b/google/cloud/spanner_v1/_helpers.py
@@ -591,9 +591,7 @@ def _check_unavailable(exc):
     resumable_error = (
         any(
             resumable_message in exc.message
-            for resumable_message in (
-                "Service unavailable",
-            )
+            for resumable_message in ("Service unavailable",)
         ),
     )
     if not resumable_error:
--- a/google/cloud/spanner_v1/_helpers.py
+++ b/google/cloud/spanner_v1/_helpers.py
@@ -587,12 +587,6 @@ def _check_rst_stream_error(exc):
         raise
 
 
-def _check_unavailable(exc):
-    # For UNAVAILABLE, we don't need to check against the
-    # messages as these should just be noop and retryable always.
-    return
-
-
 def _metadata_with_leader_aware_routing(value, **kw):
     """Create RPC metadata containing a leader aware routing header
 
--- a/google/cloud/spanner_v1/_helpers.py
+++ b/google/cloud/spanner_v1/_helpers.py
@@ -588,14 +588,9 @@ def _check_rst_stream_error(exc):
 
 
 def _check_unavailable(exc):
-    resumable_error = (
-        any(
-            resumable_message in exc.message
-            for resumable_message in ("Service unavailable",)
-        ),
-    )
-    if not resumable_error:
-        raise
+    # For UNAVAILABLE, we don't need to check against the
+    # messages as these should just be noop and retryable always.
+    return
 
 
 def _metadata_with_leader_aware_routing(value, **kw):
--- a/google/cloud/spanner_v1/_helpers.py
+++ b/google/cloud/spanner_v1/_helpers.py
@@ -587,6 +587,19 @@ def _check_rst_stream_error(exc):
         raise
 
 
+def _check_unavailable(exc):
+    resumable_error = (
+        any(
+            resumable_message in exc.message
+            for resumable_message in (
+                "Service unavailable",
+            )
+        ),
+    )
+    if not resumable_error:
+        raise
+
+
 def _metadata_with_leader_aware_routing(value, **kw):
     """Create RPC metadata containing a leader aware routing header
 
--- a/google/cloud/spanner_v1/pool.py
+++ b/google/cloud/spanner_v1/pool.py
@@ -15,15 +15,22 @@
 """Pools managing shared Session objects."""
 
 import datetime
+import functools
 import queue
 import time
 
+from google.api_core.exceptions import InternalServerError
+from google.api_core.exceptions import ServiceUnavailable
 from google.cloud.exceptions import NotFound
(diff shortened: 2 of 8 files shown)


odeke-em:
Kindly help me run the bots on this @olavloite @surbhigarg92 @harshachinta @sakthivelmanii

odeke-em:
Kindly help me re-run those bots @olavloite

tests/mockserver_tests/mock_server_test_base.py
```diff
@@ -61,6 +61,27 @@ def aborted_status() -> _Status:
 def unavailable_status() -> _Status:
     error = status_pb2.Status(
         code=code_pb2.UNAVAILABLE,
+        message="Received unexpected EOS on DATA frame from server",
```
olavloite, line 64: This error message would only be returned in combination with an INTERNAL error code from Spanner.

tests/mockserver_tests/mock_server_test_base.py
```diff
@@ -61,6 +61,27 @@ def aborted_status() -> _Status:
...
+                "google.rpc.retryinfo-bin",
+                retry_info.SerializeToString(),
+            ),
+        ),
+    )
+    return status
+
+
+# Creates an INTERNAL status with the smallest possible retry delay.
+def internal_status() -> _Status:
```
olavloite, line 82: INTERNAL errors normally do not include any retry info.

google/cloud/spanner_v1/_helpers.py
```diff
@@ -587,6 +587,17 @@ def _check_rst_stream_error(exc):
         raise
 
 
+def _check_unavailable(exc):
```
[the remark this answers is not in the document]
odeke-em: Got it, thank you @olavloite!

tests/mockserver_tests/mock_server_test_base.py
odeke-em: True, it was a mistype. Fixed and thank you.

tests/mockserver_tests/mock_server_test_base.py
odeke-em: Roger that and thanks. Done!

odeke-em:
@olavloite @surbhigarg92 @sakthivelmanii @harshachinta @rahul2393 kindly help me re-run the bots. Thank you.

google/cloud/spanner_v1/pool.py
```diff
@@ -254,11 +261,25 @@ def bind(self, database):
...
+                        ),
+                    )
+                    return method(*args, **kwargs)
+
+                resp = _retry(
+                    wrapped_method,
+                    allowed_exceptions={
+                        InternalServerError: _check_rst_stream_error,
+                        ServiceUnavailable: _check_unavailable,
+                    },
```
olavloite, line 282: We should not add new retry logic here (and in general also not elsewhere). The general rules for retries are:

1. For unary RPCs: We rely on Gax handling all retries. This means that any RPC that has been configured as retryable at that level are automatically retried by Gax. We don't need to do anything in the handwritten client library. Currently, the retry [configuration specifies](https://github.com/googleapis/googleapis/blob/7d43527b927f65a80de9f0071d0a2e1c172de849/google/spanner/v1/spanner_grpc_service_config.json#L55) that `UNAVAILABLE` and `RESOUCE_EXHAUSTED` are retryable error codes. But the client does not need to know, as this is handled by Gax.
2. For streaming RPCs: The have disabled retries by Gax by [not adding any retryable error codes to the configuration](https://github.com/googleapis/googleapis/blob/7d43527b927f65a80de9f0071d0a2e1c172de849/google/spanner/v1/spanner_grpc_service_config.json#L18). Retries of streaming RPCs therefore need to be handled manually in the handwritten client. Streaming RPCs should be retried in case of an UNAVAILABLE error, OR if it is one of the specific INTERNAL errors. (Note however that it is fair to consider the current retry implementation as complete, even if for example specific INTERNAL errors are currently not retried. Adding support for request-id is not about fixing potentially missing retry logic.)
3. For transactions, we retry the __entire transaction__ if the transaction is aborted by Spanner. This type of retry is 'irrelevant' for the request-id. With 'irrelevant' I mean that these retries do not need any special handling, and should not lead to the attempt number being increased.

tests/mockserver_tests/test_request_id_header.py
```diff
@@ -0,0 +1,317 @@
...
+                (1, REQ_RAND_PROCESS_ID, NTH_CLIENT, CHANNEL_ID, 3, 1),
+            )
+        ]
+
+        assert got_unary_segments == want_unary_segments
+        assert got_stream_segments == want_stream_segments
+
+    def test_unary_retryable_error(self):
+        add_select1_result()
+        add_error(SpannerServicer.BatchCreateSessions.__name__, unavailable_status())
```
olavloite, line 229: I changed the error code here from `INTERNAL` to `UNAVAILABLE`, because:
1. `INTERNAL` was not retried before for unary RPCs, and this change should not add retries for that. The specific `INTERNAL` errors that are retried are only relevant for streaming RPCs.
2. This PR seems to try to add additional retry handling for `UNAVAILABLE` errors. That handling is however never used, because the RPC itself is already retried by Gax, and this happens at a lower level. This again causes this test to seem to work, however the attempt number is not being increased during these retries. Because this test initially used `INTERNAL` to test, the additional retry handler was being invoked, and that did increase the attempt number.

odeke-em:
@olavloite I've added a TODO so that we can firstly merge the code in and then can fully focus on figuring out request-id with unary retries. Kindly please take a look at the existing code

olavloite (approved):

PR #1381, opened 2025-05-23
feat: Add support for multiplexed sessions

rahul2393:


tests/system/test_session_api.py
```diff
@@ -3169,3 +3328,148 @@ def test_interval_array_cast(transaction):
...
+    snapshot1_is_multiplexed = None
+    snapshot2_is_multiplexed = None
+
+    snapshot1 = sessions_database.snapshot()
+    snapshot2 = sessions_database.snapshot()
+
+    try:
+        with snapshot1 as snap1, snapshot2 as snap2:
+            rows1 = list(snap1.read(sd.TABLE, sd.COLUMNS, sd.ALL))
+            rows2 = list(snap2.read(sd.TABLE, sd.COLUMNS, sd.ALL))
```
currantw, line 3357: Tests seem to be failing here because the snapshot's session has does not exist. This means that `Session.create()` has not been called. Not sure exactly why that wouldn't be the case, but tough to find out without being able to run locally.

rahul2393: Issue was some tests were calling session.delete which prematurely deletes the multiplex session and hence "Session Not Found" error, Using Emulator calling session.delete() does not actually delete it from Emulator so issue does not occur using Emualtor

sakthivelmanii (dismissed):

sakthivelmanii (approved):

PR #1383, opened 2025-05-29
feat: Add support for multiplexed sessions

currantw:
feat: Add support for multiplexed sessions:
- Enable the use of multiplexed session for partitioned transactions.
- Support disabling multiplexed for all client database instances in case of failure.
- Support precommit tokens ingest and output in commit requests.
- Add mutation key to begin transaction requests for mutations-only transactions.
- Retry commit if commit response contains a precommit token.
- Deprecate methods that do not support multiplexed sessions.
- Cleanup: add type hints; change class attributes to instance attributes.


tests/unit/test_database_session_manager.py
```diff
@@ -0,0 +1,243 @@
...
+        """Sets environment variables to disable multiplexed sessions for all transactions types."""
+        environ[SessionOptions.ENV_VAR_FORCE_DISABLE_MULTIPLEXED] = "true"
+
+    @staticmethod
+    def _enable_multiplexed_sessions() -> None:
+        """Sets environment variables to enable multiplexed sessions for all transaction types."""
+        environ[SessionOptions.ENV_VAR_ENABLE_MULTIPLEXED] = "true"
+        environ[SessionOptions.ENV_VAR_ENABLE_MULTIPLEXED_FOR_PARTITIONED] = "true"
+        environ[SessionOptions.ENV_VAR_ENABLE_MULTIPLEXED_FOR_READ_WRITE] = "true"
+        environ[SessionOptions.ENV_VAR_FORCE_DISABLE_MULTIPLEXED] = "false"
```
rahul2393, line 243: We don't need to force disable multiplexed

currantw: We don't need it _at all_? Is in, you are removing it from the scope of the project?

google/cloud/spanner_v1/request_id_header.py
```diff
@@ -49,6 +49,10 @@ def with_request_id(
     return all_metadata
 
 
+def build_request_id(client_id, channel_id, nth_request, attempt):
```
currantw, line 52: Added helper method to reduce duplication in tests.

rahul2393 (approved):

PR #1385, opened 2025-06-06
perf: Skip gRPC trailers for StreamingRead & ExecuteStreamingSql

sakthivelmanii:
**Description:**

gRPC trailers are headers which is usually sent after all the messages are sent by the server in the stream. Trailers usually causes additional latency of 250-300 μs after last message is sent in the stream. From the server, we will receiving last field which tells us about last PartialResultSet and we no longer will be waiting for trailers to assume that all the messages are received.


tests/mockserver_tests/test_basics.py
```diff
@@ -17,26 +17,32 @@
...
     unavailable_status,
-    add_single_result,
+    add_execute_streaming_sql_results,
 )
 
 
 class TestBasics(MockServerTestBase):
+    def setUp(self):
+        super().setUp()
+        super().setup_class()
```
olavloite, line 44: Why do we need to call `setup_class()` for each test method?

olavloite (dismissed):

tests/mockserver_tests/test_basics.py
sakthivelmanii: Let me remove this. This is needed only when I am running individual tests in PyCharm

olavloite (approved):

PR #1389, opened 2025-06-10
feat: Add support for multiplexed sessions - read/write

currantw:
feat: Add support for multiplexed sessions:
- Support multiplexed sessions for read/write transactions.
- Remove `Session._transaction` attribute.
- Refactor logic for creating transaction selector to base class.
- Add previous commit ID for transaction using `run_in_transaction`.

### DIFF
--- a/google/cloud/spanner_v1/session.py
+++ b/google/cloud/spanner_v1/session.py
@@ -74,9 +74,6 @@ class Session(object):
         self._database = database
         self._session_id: Optional[str] = None
 
-        # TODO multiplexed - remove
-        self._transaction: Optional[Transaction] = None
-
         if labels is None:
             labels = {}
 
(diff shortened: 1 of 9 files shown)


google/cloud/spanner_v1/session.py
```diff
@@ -528,42 +520,43 @@ def run_in_transaction(self, func, *args, **kw):
...
-                if self._transaction is None:
-                    txn = self.transaction()
-                    txn.transaction_tag = transaction_tag
-                    txn.exclude_txn_from_change_streams = (
-                        exclude_txn_from_change_streams
-                    )
-                    txn.isolation_level = isolation_level
-                else:
-                    txn = self._transaction
+                # [A] Build transaction
```
lszinv, line 540: If you're labeling the branching as A/B/C, it would be good to talk about the general flow before it happens, so that when the user reading the code knows what to expect as they read through. 

Or else maybe don't include them. The separator comments are also not consistent with the way the rest of the code base is commented.

currantw: Just removed them 🤷

google/cloud/spanner_v1/snapshot.py
```diff
@@ -983,6 +972,34 @@ def before_next_retry(nth_retry, delay_in_seconds):
...
+
+        :rtype: :class:`transaction_pb2.TransactionSelector`
+        :returns: a transaction selector for this snapshot.
+        """
+
+        # Select a previously begun transaction.
+        if self._transaction_id is not None:
+            return TransactionSelector(id=self._transaction_id)
+
+        options = self._build_transaction_options_pb()
```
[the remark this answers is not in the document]
currantw: It will if run from `_SnapshotBase`. But both `Snapshot` and `Transaction` override this method.

rahul2393 (approved):

Issue #1392, opened 2025-07-02
Remove Python 3.7/3.8 or re-enable tests for these runtimes


parthea:
Python 3.7/3.8 tests have been disabled in https://github.com/googleapis/python-spanner/pull/1391. We should either remove Python 3.7/3.8 support in the main branch, or restore the tests if/when there is a fix for `CVE-2025-47273/CVE-2024-6345`.

PR #1393, opened 2025-07-04
tests: update constraints for python 3.9 testing

parthea:
Since Python 3.7 is no longer being tested following https://github.com/googleapis/python-spanner/pull/1391 (see follow up in https://github.com/googleapis/python-spanner/issues/1392), we should update the constraints files for python 3.9 testing to ensure that the minimum versions of dependencies stated in setup.py are compatible.

https://github.com/googleapis/python-spanner/blob/f81fbd5ba135194d06987f98e1b031c0fd846979/testing/constraints-3.7.txt#L1-L6


chalmerlowe:

PR #1394, opened 2025-07-16
feat: default enable multiplex session for all operations unless explicitly set to false

rahul2393:
Read-Only, Read-Write and Partition Ops transactions now use a single multiplexed session, instead of a session from the session pool. Multiplexed sessions can handle any number of queries and read-only, read-write and partition ops transactions transactions at the same time.

The use of multiplexed sessions can be disabled for read-only, read-write and partition-ops transactions by setting the environment variables.

GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS=false
GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS_FOR_RW=false.
GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS_PARTITIONED_OPS=false.

See [https://cloud.google.com/spanner/docs/sessions#multiplexed_sessions](https://cloud.google.com/spanner/docs/sessions#multiplexed_sessions) for more background information about multiplexed sessions.


.github/workflows/integration-tests-against-emulator-with-regular-session.yaml
```diff
@@ -21,7 +21,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v5
         with:
-          python-version: 3.8
+          python-version: 3.12
```
rahul2393, line 24: this is because in followup PR we will be removing 3.8 runtime support
https://github.com/googleapis/python-spanner/pull/1395

sakthivelmanii (dismissed):

tests/mockserver_tests/mock_server_test_base.py
```diff
@@ -228,3 +229,109 @@ def database(self) -> Database:
...
+            CreateSessionRequest,
+            ExecuteSqlRequest,
+            BeginTransactionRequest,
+        )
+
+        # Count session creation requests that come before the first non-session request
+        session_requests_before = 0
+        for req in requests:
+            if isinstance(req, (BatchCreateSessionsRequest, CreateSessionRequest)):
+                session_requests_before += 1
```
surbhigarg92, line 319: Will BatchCreateSessionsRequest still happen, if Mux sessions are enabled. If No, then we should only check for CreateSessionRequest here?

surbhigarg92 (dismissed):

tests/mockserver_tests/mock_server_test_base.py
rahul2393: It will still happen, because on can disable mux for individual operation like read-only, read-write or partition ops, at that time client will fallback to use regular pools which are initialized via these batch create calls

google/cloud/spanner_v1/snapshot.py
```diff
@@ -1031,7 +1030,7 @@ def _update_for_transaction_pb(self, transaction_pb: Transaction) -> None:
             self._transaction_id = transaction_pb.id
 
         if transaction_pb.precommit_token:
-            self._update_for_precommit_token_pb(transaction_pb.precommit_token)
+            self._update_for_precommit_token_pb_unsafe(transaction_pb.precommit_token)
```
sakthivelmanii, line 1033: any reason why we need two methods? 

`_update_for_precommit_token_pb`,  `_update_for_precommit_token_pb_unsafe`

rahul2393: Read this comment `The caller is responsible for locking until the transaction ID is updated.`

https://github.com/googleapis/python-spanner/blob/main/google/cloud/spanner_v1/snapshot.py#L1026-L1029

And then when we were calling _update_for_precommit_token_pb we were taking the lock again
https://github.com/googleapis/python-spanner/blob/main/google/cloud/spanner_v1/snapshot.py#L1046

sakthivelmanii (approved):

PR #1400, opened 2025-08-12
feat: support configuring logger in dbapi kwargs

olavloite:
Allow the kwargs for dbapi connections to inlcude a logger, and use this as the logger for the database that is used.

Also set a default logger that only logs at WARN level for the mock server tests to stop them from spamming the test log with a bunch of "Created multiplexed session." messages that are logged at INFO level.

Also removes some additional log spamming from the request-id tests.


rahul2393 (dismissed):

rahul2393:
Let me fix github checks in other PR

rahul2393 (approved):

PR #1446, opened 2025-10-17
fix(spanner): resolve TypeError in metrics resource detection

sinhasubham:
**Note**: fix(spanner): resolve TypeError in metrics resource detection

**Root cause:** 
The metrics tracer factory was attempting to access the `cloud.region` attribute directly from an OpenTelemetry `Resource` object using dictionary-style subscription (`resource['key']`).

However, the `Resource` object is not subscriptable. The correct way to access its attributes is through the `.attributes` property (`resource.attributes['key']`).

This change corrects the access pattern, resolving a `TypeError` that would crash applications upon `spanner.Client` initialization when built-in metrics were enabled in a GCP environment.

**Testing**

Tested by replicating the error locally using [this script](https://paste.googleplex.com/5874183288520704).

### DIFF
--- a/google/cloud/spanner_v1/client.py
+++ b/google/cloud/spanner_v1/client.py
@@ -25,6 +25,7 @@ In the hierarchy of API concepts
 """
 import grpc
 import os
+import logging
 import warnings
 
 from google.api_core.gapic_v1 import client_info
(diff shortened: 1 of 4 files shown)


google/cloud/spanner_v1/client.py
```diff
@@ -240,19 +244,24 @@ def __init__(
...
-                            export_interval_millis=METRIC_EXPORT_INTERVAL_MS,
-                        )
-                    ]
+                        ]
+                    )
+                metrics.set_meter_provider(meter_provider)
+                SpannerMetricsTracerFactory()
+            except Exception as e:
+                log.warning(
+                    "Failed to initialize Spanner built-in metrics. Error: %s", e
```
surbhigarg92:
```suggestion
                    "Failed to initialize Spanner client side metrics. Error: %s", e
```

surbhigarg92 (approved):

sakthivelmanii (approved):

PR #1463, opened 2025-12-11
fix: transaction_tag should be set on BeginTransactionRequest

surbhigarg92:
When using multiplexed sessions, the transaction_tag should also be set on the BeginTransactionRequest.


sakthivelmanii (dismissed):

tests/unit/test_transaction.py
```diff
@@ -463,10 +463,14 @@ def _commit_helper(
...
             self.assertEqual(transaction._transaction_id, TRANSACTION_ID)
 
+            request_options = RequestOptions()
+            request_options.transaction_tag = TRANSACTION_TAG
+
             expected_begin_transaction_request = BeginTransactionRequest(
                 session=session.name,
                 options=TransactionOptions(read_write=TransactionOptions.ReadWrite()),
                 mutation_key=expected_begin_mutation,
+                request_options=request_options,
```
sakthivelmanii: `request_options=RequestOptions(transaction_tag=TRANSACTION_TAG)`

We can just pass it like this.

sakthivelmanii: We don't need an additional variable here. You can just set Request options(transaction-tag = TRANSACTION_TAG)

rahul2393 (approved):

PR #1480, opened 2026-01-16
chore: Add gcp resource name span attribute

surbhigarg92:
Adding a new span attribute called gcp.resource.name which contains an identifier to a particular spanner instance and database in the following format:

//spanner.googleapis.com/projects/{project}/instances/{instance_id}/databases/{database_id}
Example:

//spanner.googleapis.com/projects/my_project/instances/my_instance/databases/my_database


bhatt4982 (approved):
LGTM

PR #1495, opened 2026-02-11
feat(spanner): add ClientContext support to options

aseering:
This change adds support for ClientContext in Options and ensures it is propagated to ExecuteSql, Read, Commit, and BeginTransaction requests. It aligns with go/spanner-client-scoped-session-state design.

ClientContext allows passing opaque, RPC-scoped side-channel information (like application-level user context) to Spanner. This implementation supports setting ClientContext at the Client, Database, and Request levels, with request-level options taking precedence.

Key changes:
- Added ClientContext to types/spanner.py and exposed it.
- Updated Client.__init__ to accept a default client_context.
- Added helpers for merging ClientContext with correct precedence.
- Updated Snapshot, Transaction, Batch, and Database wrappers to propagate the context.
- Added comprehensive unit tests in tests/unit/test_client_context.py.


google/cloud/spanner_v1/transaction.py
```diff
@@ -651,6 +673,9 @@ def batch_update(
         if self._transaction_id is None:
             is_inline_begin = True
             self._lock.acquire()
+            if self._transaction_id is not None:
+                is_inline_begin = False
+                self._lock.release()
```
olavloite: Here also

